PR #2, opened 2023-11-09
Add example demonstrating graceful shutdown

brandur:
As I was writing the docs for graceful shutdown, I realized that it
wasn't a half bad idea to include a full code sample for a realistic
program that when exiting, would (1) try a soft stop to start, and (2)
do a hard stop if the soft stop didn't work in time.

This is somewhat non-trivial code though, and I'd be afraid to just
include it it docs without testing it. I started writing the code to be
actually runnable, and realized that as long as it was going to be
runnable anyway, we may as well also include an example for it and
commit to the repository.

So that's what we do here: add an example test that shows a realistic
shutdown loop demonstrating both a soft and hard stop.

### DIFF
--- a/example_graceful_shutdown_test.go
+++ b/example_graceful_shutdown_test.go
@@ -1,0 +1,149 @@
+package river_test
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+
+	"github.com/riverqueue/river"
+	"github.com/riverqueue/river/internal/riverinternaltest"
+	"github.com/riverqueue/river/internal/util/slogutil"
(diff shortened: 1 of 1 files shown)


example_graceful_shutdown_test.go
```diff
@@ -0,0 +1,149 @@
...
+	// The soft stop will never work in this example because our job only
+	// respects context cancellation, but wait a short amount of time to give it
+	// a chance. After it elapses, send another SIGTERM to initiate a hard stop.
+	select {
+	case <-riverClientStopped:
+		// Will never be reached in this example because our job will only ever
+		// finish on context cancellation.
+		fmt.Printf("Soft stop succeeded\n")
+
+	case <-time.After(100 * time.Millisecond):
```
brandur, line 139: Not super happy with this wait. It's not strictly necessary to include it, but if we don't, then we don't show any opportunity for the soft stop to work at all, so I put it in.

bgentry (approved):

brandur:
TY.

PR #4, opened 2023-11-10
A few fixes/changes for the graceful shutdown example

brandur:
Follows up #2 with a few fixes/changes. As I was writing documentation
for graceful shutdown I realized there were a few things that weren't
quite ideal:

* The job should return an error in the event of cancellation so that it
  can be persisted as errored and be worked again.

* The example now respects either `SIGINT` or `SIGTERM`. `SIGTERM` is
  what's used on Heroku, but `SIGINT` is the standard signal from
  `Ctrl+C` in a terminal, so by respecting both we can have a program
  that works well in either development or a common hosted environment.

* Add a third phase in which the program initiates an unclean stop by
  not waiting on stop any longer. This is probably something that most
  programs should have because it's going to be reasonably easy to write
  workers that accidentally don't respect context cancellation and get
  stuck.

* Add a 10 second timeout to each phase. This is for Heroku's benefit.
  It'll send one `SIGTERM` and wait 30 seconds before issuing `SIGKILL`.
  So the program here waits 10 seconds for a soft stop, another 10
  seconds for a hard stop, and then exits uncleanly on its own volition
  before getting `SIGKILL`ed.

### DIFF
--- a/example_graceful_shutdown_test.go
+++ b/example_graceful_shutdown_test.go
@@ -32,16 +32,22 @@ type WaitsForCancelOnlyWorker struct {
 func (w *WaitsForCancelOnlyWorker) Work(ctx context.Context, job *river.Job[WaitsForCancelOnlyArgs]) error {
 	fmt.Printf("Working job that doesn't finish until cancelled\n")
 	close(w.jobStarted)
+
 	<-ctx.Done()
 	fmt.Printf("Job cancelled\n")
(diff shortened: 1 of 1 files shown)


bgentry (approved):

brandur:
Thanks! I also updated to use `ctx.Err()` as you pointed out in the docs review.

PR #6, opened 2023-11-11
Add instructions on using `NewClient` in its Godoc docblock

brandur:
There's an example at the top of the River package demonstrating the use
of `NewClient`, but as I was reading the docs today, I realized that
since `NewClient` is somewhat non-trivial to invoke because it requires
a generic parameter, it'd be a good idea to also have an example of its
use right in `NewClient`'s docblock. That way, if a user is viewing this
function they don't need to go hunt around for examples to try and
figure out how to use it.

### DIFF
--- a/client.go
+++ b/client.go
@@ -321,8 +321,30 @@ var (
 // NewClient creates a new Client with the given database driver and
 // configuration.
 //
-// Currently only one driver is supported, which is Pgx V5. See package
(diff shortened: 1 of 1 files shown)


bgentry (approved):
Nice little improvement :clap: :shipit:

brandur:
Thx!

PR #7, opened 2023-11-11
Error on batch insert if `InsertOpts.UniqueOpts` is specified

brandur:
As I was writing docs for batch insertion today [1], I remembered that
we don't support job uniqueness on batch inserts because the mechanism
uses PG advisory locks, and holding many locks at once could easily lead
to contention and deadlocks.

We may remove this limitation in the future, but I figure that in the
meantime, it might not be the worst idea to error if they're specified
on batch insert so as not to mislead the user that what they expected to
happen happened. That's what this change does.

[1] https://riverqueue.com/docs/batch-job-insertion

### DIFF
--- a/client.go
+++ b/client.go
@@ -1059,21 +1059,9 @@ type InsertManyParams struct {
 //		// handle error
 //	}
 func (c *Client[TTx]) InsertMany(ctx context.Context, params []InsertManyParams) (int64, error) {
-	if len(params) < 1 {
-		return 0, errors.New("no jobs to insert")
-	}
-
-	insertParams := make([]*dbadapter.JobInsertParams, len(params))
-	for i, param := range params {
-		if err := c.validateJobArgs(param.Args); err != nil {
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -1087,6 +1091,10 @@ func (c *Client[TTx]) InsertManyTx(ctx context.Context, tx TTx, params []InsertM
 			return 0, err
 		}
 
+		if err := c.validateManyInsertOpts(param.InsertOpts); err != nil {
+			return 0, err
+		}
+
```
bgentry, line 1097: These method bodies are getting to be fairly long for being entirely duplicated, thoughts on extracting everything except the last line of each into a shared unexported method or helper?

bgentry (approved):

client.go
brandur: WFM. Update to include `Client.insertManyParams`.

PR #8, opened 2023-11-11
Rename `RunImmediately` to `RunOnStart`

brandur:
This is another bikeshed issue, but I was thinking about the naming of
`RunImmediately` today and I think it might be too ambiguous. It sounds
a little too much like it could mean "run immediately when scheduled".

I propose we rename it to `RunOnStart` to make it clear that it's
expected to run on the scheduler's start (this is also said in the
property's docblock). `RunOnStart` is also a little more succinct which
is always nice.

The repository is technically public now and this is technically a
breaking change, but I think it's fair for us to make a few API tweaks
in the early days before we've done anything to send the project around
(and maybe even a little after).

This property in mentioned in our public docs already [1]. If this gets
merged I'll follow up by updating it there as well.

[1] https://riverqueue.com/docs/periodic-jobs

### DIFF
--- a/client.go
+++ b/client.go
@@ -487,8 +487,8 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 					ConstructorFunc: func() (*dbadapter.JobInsertParams, error) {
 						return insertParamsFromArgsAndOptions(periodicJob.constructorFunc())
 					},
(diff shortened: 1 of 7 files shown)


brandur:
@bgentry Thoughts on this one?

bgentry (approved):

PR #10, opened 2023-11-11
Rename all `*RetentionTime` vars/constants to `*RetentionPeriod`

brandur:
Our various maintenance services take a series of durations that are
currently name `*RetentionTime` like `CompletedJobRetentionTime`. Given
that these are durations and not times, more apt naming for them is a
suffix of `*RetentionPeriod`, which is the change made here.


bgentry (approved):

brandur:
Thanks!

PR #11, opened 2023-11-11
Update unique job example to be a little more realistic

brandur:
As I was writing the docs for unique jobs, I found that I wanted to have
an example that was a little more elaborate than what we already had
which demonstrated the use of uniqueness along multiple properties
instead of just duration.

Here, update the unique opts example so that the job becomes reconciling
a single account that should run every 24 hours, so jobs for account 1
and account 2 are inserted separately, and we show that a duplicate for
account 1 isn't allowed. I suspect that this change makes the example a
little more realistic along the lines of what users will want to do so
it has that benefit in addition to being more conducive for the docs.

### DIFF
--- a/example_unique_job_test.go
+++ b/example_unique_job_test.go
@@ -14,44 +14,48 @@ import (
 	"github.com/riverqueue/river/riverdriver/riverpgxv5"
 )
 
-// Account represents a minimal account containing a unique identifier, recent
-// expenditures, and a remaining total.
+// Account represents a minimal account including recent expenditures and a
+// remaining total.
 type Account struct {
-	ID                 int
 	RecentExpenditures int
 	AccountTotal       int
 }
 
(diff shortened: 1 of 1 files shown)


example_unique_job_test.go
```diff
@@ -14,44 +14,47 @@ import (
 	"github.com/riverqueue/river/riverdriver/riverpgxv5"
 )
 
-// Account represents a minimal account containing a unique identifier, recent
-// expenditures, and a remaining total.
+// Account represents a minimal account recent expenditures and remaining total.
```
bgentry, line 17: typo or missing words in here I think? "account recent expenditures"

bgentry (approved):

example_unique_job_test.go
brandur: Ah dammit. Last second change and of course I messed it up. Thanks, fixed.

PR #13, opened 2023-11-12
Add context to error handler's `HandleError` and `HandlePanic`

brandur:
As I was writing the documentation for `ErrorHandler` I realized that
it'd probably be desirable for its `HandleError` and `HandlePanic`
functions to take a context. The user might, for example, have a logger
embedded in context which they'd extract and use the log the error or
panic. This could also be done with a member field on implementing
struct of course, but it feels like making context available doesn't
really have a downside and its presence would likely be expected by some
users.

Another benefit is that in case someone is doing some heavy lifting in
one of these (they probably shouldn't be, but just in case), their
handlers could respond to the context cancellation caused by a client
`StopAndCancel` shutdown. Currently, the handlers are immune to
cancellation and could conceivably cause a shutdown to be stuck.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -871,7 +871,7 @@ func Test_Client_ErrorHandler(t *testing.T) {
 
 		var errorHandlerCalled bool
 		config.ErrorHandler = &testErrorHandler{
-			HandleErrorFunc: func(job *JobRow, err error) *ErrorHandlerResult {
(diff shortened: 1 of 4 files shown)


bgentry (approved):
Wow, great catch, I'm not sure why we didn't think of this when we implemented this. Thank you!

brandur:
Thanks! And yeah, also surprised I didn't catch this at the time.

PR #16, opened 2023-11-12
Add example test for `ErrorHandler`

brandur:
I was writing docs on how to handle errors and panics in Go and realized
that we didn't have an example demonstrating the use of `ErrorHandler`
that I could copy/link to. Here, add one in.

### DIFF
--- a/example_error_handler_test.go
+++ b/example_error_handler_test.go
@@ -1,0 +1,117 @@
+package river_test
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+
+	"github.com/riverqueue/river"
+	"github.com/riverqueue/river/internal/riverinternaltest"
+	"github.com/riverqueue/river/internal/util/slogutil"
+	"github.com/riverqueue/river/riverdriver/riverpgxv5"
+)
+
+type CustomErrorHandler struct{}
+
+func (*CustomErrorHandler) HandleError(ctx context.Context, job *river.JobRow, err error) *river.ErrorHandlerResult {
+	fmt.Printf("Job errored with: %s\n", err)
+	return nil
+}
+
+func (*CustomErrorHandler) HandlePanic(ctx context.Context, job *river.JobRow, panicVal any) *river.ErrorHandlerResult {
+	fmt.Printf("Job panicked with: %v\n", panicVal)
+
+	// Either function can also set the job to be immediately cancelled.
+	return &river.ErrorHandlerResult{SetCancelled: true}
+}
+
+type ErroringArgs struct {
+	ShouldError bool
+	ShouldPanic bool
+}
+
+func (ErroringArgs) Kind() string { return "erroring" }
+
+// Here to make sure our jobs are never accidentally retried which would add
+// additional output and fail the example.
+func (ErroringArgs) InsertOpts() river.InsertOpts {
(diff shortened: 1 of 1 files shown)


example_error_handler_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+type CustomErrorHandler struct{}
+
+func (*CustomErrorHandler) HandleError(ctx context.Context, job *river.JobRow, err error) *river.ErrorHandlerResult {
+	fmt.Printf("Job errored with: %s\n", err)
+
+	// Either function can also set the job to be immediately cancelled, which
+	// we take advantage of here to make sure it's not retried in the example.
+	// Can also be `return nil`.
+	return &river.ErrorHandlerResult{SetCancelled: true}
```
brandur, line 24: See note here, but sort of had to set cancelled because otherwise there's a chance the job could retry before the example finishes up.

I guess I could also set max attempts to 1. Not sure which one is better.

example_error_handler_test.go
```diff
@@ -0,0 +1,113 @@
...
+	if err := riverinternaltest.TruncateRiverTables(ctx, dbPool); err != nil {
+		panic(err)
+	}
+
+	workers := river.NewWorkers()
+	river.AddWorker(workers, &ErroringWorker{})
+
+	riverClient, err := river.NewClient(riverpgxv5.New(dbPool), &river.Config{
+		ErrorHandler: &CustomErrorHandler{},
+		Logger:       slog.New(&slogutil.SlogMessageOnlyHandler{Level: 9}), // Suppress logging so example output is cleaner (9 > slog.LevelError).
```
brandur, line 74: This is a bit unconventional, but there's a fair bit of error output on error/panic from `jobExecutor`. Setting level to > `slog.LevelError` suppresses it and makes the test output much cleaner, but in case of something bad happening might also suppress other things I suppose. Not sure which is the lesser of two evils, but this is the one I went with.

bgentry: The verbose error/panic output might be something we need to take a look at. I don't think there was much intentional design there, more that it just evolved that way as we pieced together functionality.

example_error_handler_test.go
bgentry: I'm torn. It's useful for this example to also demonstrate `SetCancelled` because there's nowhere else we'd do that. However I worry a bit about people copy-pasta'ing this without realizing what they're doing, particularly in `HandlePanic` where there's no comment about it.

Maybe the answer is to do both? If we set MaxAttempts to 1 for the entire worker, then we could easily show off both the `SetCancelled: true` usage as well as the more common usage where you probably don't want to cancel any job that panics or errors.

If we go that route, I'd say it makes the most sense to put the `SetCancelled` in `HandlePanic` so that `HandleError` can demonstrate the most common usage.

bgentry (approved):

example_error_handler_test.go
brandur: True. I don't think we're _too_ far off though. If I was running this thing I'd expect to that there was reasonable error logging by default so problems aren't black holed without me having to chase down more configuration.

example_error_handler_test.go
brandur: +1 all that. Changed to max attempts of 1, and made `HandleError` return nil for copy/pasta convenience.

brandur:
Thanks!

PR #17, opened 2023-11-13
Add example for using subscriptions

brandur:
This one's a little odd because basically all the other examples already
use a subscription to wait for jobs to finish, but I figure that
subscriptions should have their own distinct example showing their
various idiosyncrasies


example_subscription_test.go
```diff
@@ -0,0 +1,120 @@
...
+	_, err = riverClient.Insert(ctx, SubscriptionArgs{Fail: true}, nil)
+	if err != nil {
+		panic(err)
+	}
+	_, err = riverClient.Insert(ctx, SubscriptionArgs{Cancel: true}, nil)
+	if err != nil {
+		panic(err)
+	}
+
+	waitForJob := func(subscribeChan <-chan *river.Event) {
```
brandur, line 99: I didn't use the common `waitForNJobs` helper here because I figure it's better to expose the exact details of what's happening here for this particular example.

event.go
```diff
@@ -18,8 +18,8 @@ const (
 
 	// EventKindJobFailed occurs when a job fails. Occurs both when a job fails
 	// and will be retried and when a job fails for the last time and will be
-	// discarded. Callers can use job-specific fields like `Attempt` and
-	// `Status` to differentiate each type of occurrence.
+	// discarded. Callers can use job fields like `Attempt` and `State` to
```
brandur, line 21: Gonna try to sneak this tiny fix in — `Status` should be `State`.

example_subscription_test.go
bgentry: Yep, that's the right call, keep everything important inline for an example like this 👍

bgentry (approved):

brandur:
Thanks!

PR #18, opened 2023-11-13
Drop client `Schema` configuration

### LINKED ISSUE #15: Client's `Schema` config does nothing
Unless I'm mistaken, the `Schema` option that can be sent to a client doesn't do anything.

I think all we need to do is to make sure to run `SET SCHEMA 'xxx'` on every new connection?

Also, even if this feature did work, the migration tool doesn't currently support an alternate schema, so that needs to be corrected too.

cc @bgentry

brandur:
As discussed in #15, the `Schema` configuration parameter doesn't
actually do anything, and we realized that it's possible to support this
feature by using an alternate configuration for pgxpool in `search_path`.

Here, drop `Schema` and add a test that verifies that this technique
works. So as to not have to create and migrate an alternate schema, we
cheat a little bit by repointing the schema and then just verifying that
we can't find a `river_job` table there.

I'm finding the test hierarchy in `client_test.go` to be pretty
confusing in determining on what best practice is for what should go
where, so I'm starting out with a new `Test_Client` which I'm hoping we
can start using as a base for most of the general tests in here in a
more reusable way. I leverage `JobArgsReflectKind` and `WorkFunc` in its
subtests so that it's not dependent on our various global job args that
are defined all over the place. Instead, args and the work routine are
colocated right inside the test.

Configuring an alternate schema is definitely going to need its own
documentation page somewhere because it's not obvious, but that'll come
separately.

Fixes #15.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/riverqueue/river/internal/maintenance"
 	"github.com/riverqueue/river/internal/rivercommon"
 	"github.com/riverqueue/river/internal/riverinternaltest"
+	"github.com/riverqueue/river/internal/rivertest"
 	"github.com/riverqueue/river/internal/util/ptrutil"
 	"github.com/riverqueue/river/internal/util/sliceutil"
 	"github.com/riverqueue/river/internal/util/valutil"
@@ -141,7 +142,7 @@ func newTestClient(ctx context.Context, t *testing.T, config *Config) *Client[pg
 	return client
 }
 
-func runClient(ctx context.Context, t *testing.T, client *Client[pgx.Tx]) {
+func startClient(ctx context.Context, t *testing.T, client *Client[pgx.Tx]) {
 	t.Helper()
 
 	if err := client.Start(ctx); err != nil {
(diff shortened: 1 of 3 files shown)


client_test.go
```diff
@@ -141,7 +142,7 @@ func newTestClient(ctx context.Context, t *testing.T, config *Config) *Client[pg
 	return client
 }
 
-func runClient(ctx context.Context, t *testing.T, client *Client[pgx.Tx]) {
+func startClient(ctx context.Context, t *testing.T, client *Client[pgx.Tx]) {
```
brandur, line 145: Sorry for the spam, but renamed this since we renamed `Run` -> `Start`.

client_test.go
```diff
@@ -1896,7 +1967,7 @@ func Test_Client_UnknownJobKindErrorsTheJob(t *testing.T) {
 	require.NoError(client.Stop(ctx))
 }
 
-func Test_Client_Run_Error(t *testing.T) {
+func Test_Client_Start_Error(t *testing.T) {
```
brandur, line 1970: Same idea here. `Run` -> `Start`.

client_test.go
```diff
@@ -158,10 +159,80 @@ func runClient(ctx context.Context, t *testing.T, client *Client[pgx.Tx]) {
 func runNewTestClient(ctx context.Context, t *testing.T, config *Config) *Client[pgx.Tx] {
 	t.Helper()
 	client := newTestClient(ctx, t, config)
-	runClient(ctx, t, client)
+	startClient(ctx, t, client)
 	return client
 }
 
+func Test_Client(t *testing.T) {
```
brandur, line 166: Started a new branch of client tests because it's not clear what should go where in here anymore, and I'm trying to put in something where new tests can be added without starting a fresh test every time. We'll try to take advantage of `JobArgsReflectKind` and `WorkFunc` in subtests so that args and work definition can be colocated right inside the test rather than being floating around in global state somewhere.

`StartInsertAndWork` below is a basic client test (not related to the `Schema` change) added to show roughly what this look like.

bgentry (approved):
Nicely done, I like the new test pattern. If nothing else, inline worker definitions are great for concise & easy to follow tests.

brandur:
TY TY.

PR #19, opened 2023-11-13
Demonstrate use of goreadme to generate README

brandur:
Here, we demonstrate the use of goreadme [1] to generate a README for
the project. This is not mergable, but it's good to see what it looks
like.

Top problems:

* goreadme hasn't been updated to allow for internal Godoc links, so
  links like `[Client]` that Godoc will point to a local reference,
  don't get a link target in the README.

* Any line that beings with `-` is interpreted as a diff block, which
  unfortunately includes our bullet point lists. I actually prefer the
  use of `*` for lists, but something in VSCode (gofmt I think?) doesn't
  let me use `*` and always converts to `-`, so those not being handle
  is a non-starter.

Command used to generate this:

    goreadme -badge-godoc -credit=false -skip-examples -skip-sub-packages -import-path=github.com/riverqueue/river > README.md

[1] https://github.com/posener/goreadme

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -1,0 +1,28 @@
+# River
+
+River is an experimental Postgres queue for Go.
+
+## Development
+
+### Run tests
+
+Raise test databases:
+
+    go run ./internal/cmd/testdbman create
+
+Run tests:
+
+    go test ./...
+
+### Run lint
+
+Run the linter and try to autofix:
+
+    golangci-lint run --fix
+
+### Generate sqlc
+
+The project uses sqlc (`brew install sqlc`) to generate Go targets for Postgres
+queries. After changing an sqlc `.sql` file, generate Go with:
+
+    make generate
--- a/README.md
+++ b/README.md
@@ -1,28 +1,182 @@
-# River
+# river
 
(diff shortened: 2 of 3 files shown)


brandur:
Rendered:
https://github.com/riverqueue/river/blob/brandur-demonstrate-goreadme/README.md

docs/development.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Run tests
+
+Raise test databases:
+
+    go run ./internal/cmd/testdbman create
+
+Run tests:
+
+    go test ./...
```
bgentry, line 15: Unfortunately our current `-p 1` requirement makes this not quite right 😞

bgentry (commented):
Darn. It's reasonably close to a good starting point, but maybe this approach isn't really maintainable long-term anyway. Do you think we should keep the readme as a short summary and punt to godoc + website for more details?

docs/development.md
brandur: Heh, true. Existing README problem unfortunately.

brandur:
Closing in favor of #21.

PR #22, opened 2023-11-13
pin to riverpgxv5@v0.0.1

bgentry:
To make this work, also set up a replace directive to use the local riverpgxv5 from within the river package's go.mod.

I tested this locally from our riverdemo test app and confirmed that I was able to import riverpgxv5@v0.0.1, and then even tested with a locally tagged river@v0.0.2 and was able to use that at the same time.

The [gomoddirectives](https://github.com/ldez/gomoddirectives/blob/master/gomoddirectives_test.go) linter does not allow any `replace` directives by default, but we can customize that to allow this one replace if you think this seems sane.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -64,6 +64,9 @@ linters-settings:
       - Default
       - Prefix(github.com/riverqueue)
 
+  gomoddirectives:
+    replace-local: true
+
   gosec:
     excludes:
       - G404 # use of non-crypto random; overly broad for our use case
--- a/go.mod
+++ b/go.mod
@@ -2,11 +2,14 @@ module github.com/riverqueue/river
 
 go 1.21.0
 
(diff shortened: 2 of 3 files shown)


bgentry:
You would think I would be able to use the [`replace-allow-list`](https://golangci-lint.run/usage/linters/#gomoddirectives) option, but no, that still fails due to the local replace path. So I just have to allow all local replaces afaict.

brandur (approved):
Think it looks sane?

I was trying to read what the effect of a `replace` was on other projects that might be importing River. The Go Modules reference says this:

> replace directives only apply in the main module’s go.mod file and are ignored in other modules. See [Minimal version selection](https://go.dev/ref/mod#minimal-version-selection) for details.

I think that means it won't affect anything else? But as usual with Go documentation, the wording is too vague/ambiguous to say anything for sure.

PR #28, opened 2023-11-15
Refactor: Reduce rescuer LOCs / number of queries + optimize

brandur:
I've been looking at ways we could start consolidating queries in
`river_job.sql` a bit because the file's currently quite a bit of a
mess. One pretty easy one that came out is that we multiple queries for
the rescuer that can be just one without much trouble.

Here, make that change, which also allows to simplify the rescuer's core
run loop by quite a bit and reduce the lines of code in there. Also
expect a slight performance improvement as we need only one round trip
to the database and the streamlined code lets us preallocate a number of
Go slices and reduce the number of `append`s happening.

### DIFF
--- a/rescuer.go
+++ b/rescuer.go
@@ -25,15 +25,13 @@ const (
 
 // Test-only properties.
 type rescuerTestSignals struct {
-	DiscardedJobs rivercommon.TestSignal[struct{}] // notifies when runOnce has discarded jobs from the batch
-	FetchedBatch  rivercommon.TestSignal[struct{}] // notifies when runOnce has fetched a batch of jobs
-	RetriedJobs   rivercommon.TestSignal[struct{}] // notifies when runOnce has retried jobs from the batch
+	FetchedBatch rivercommon.TestSignal[struct{}] // notifies when runOnce has fetched a batch of jobs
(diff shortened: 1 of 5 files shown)


rescuer.go
```diff
@@ -25,15 +25,13 @@ const (
...
-	FetchedBatch  rivercommon.TestSignal[struct{}] // notifies when runOnce has fetched a batch of jobs
-	RetriedJobs   rivercommon.TestSignal[struct{}] // notifies when runOnce has retried jobs from the batch
+	FetchedBatch rivercommon.TestSignal[struct{}] // notifies when runOnce has fetched a batch of jobs
+	UpdatedBatch rivercommon.TestSignal[struct{}] // notifies when runOnce has updated rescued jobs from a batch
 }
 
 func (ts *rescuerTestSignals) Init() {
-	ts.DiscardedJobs.Init()
 	ts.FetchedBatch.Init()
-	ts.RetriedJobs.Init()
```
brandur, line 36: Dropped down to only one signal because these are both part of the same operation now.

bgentry (approved):
Solid refactor, and basically no changes to tests required. Nice ✨

brandur:
Thanks!

PR #31, opened 2023-11-15
Add missing integration-level job snooze test

brandur:
I noticed while trying to refactor something the other day that I had a
problem around snoozing, but only the example snooze test was failing.
Looking into it further, it turned out that it was because the snooze
example was the only snooze test we had.

Example tests aren't great because they do obnoxious things like swallow
all output (making it harder to debug), but they also don't support some
feature like `-count`, making it harder to debug intermittent problems.

Here, add another client test for snooze. We use the new client test
line that makes use of `JobArgsReflectKind` and `WorkFunc`, keeping the
implementation clean and all local to the specific test case that needs
it.

And since snooze and cancel are quite similar, I added a test case for
cancel as well. Cancel did have another test elsewhere, but I figure I
might as well add one in this new line too to line up with the snooze
test case. It should be a pretty fast test, so doesn't hurt to have a
duplicate.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -168,7 +168,10 @@ func Test_Client(t *testing.T) {
 
 	ctx := context.Background()
 
-	type testBundle struct{}
+	type testBundle struct {
+		queries       *dbsqlc.Queries
+		subscribeChan <-chan *Event
+	}
 
 	setup := func(t *testing.T) (*Client[pgx.Tx], *testBundle) {
 		t.Helper()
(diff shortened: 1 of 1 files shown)


client_test.go
```diff
@@ -203,6 +214,54 @@ func Test_Client(t *testing.T) {
...
+
+		client, bundle := setup(t)
+
+		type JobArgs struct {
+			JobArgsReflectKind[JobArgs]
+		}
+
+		AddWorker(client.config.Workers, WorkFunc(func(ctx context.Context, job *Job[JobArgs]) error {
+			return JobCancel(fmt.Errorf("a persisted internal error"))
+		}))
```
bgentry, line 228: Should we check the job in the database after this is all done for completeness? i.e. make sure it is cancelled and the error is saved

client_test.go
```diff
@@ -203,6 +214,54 @@ func Test_Client(t *testing.T) {
...
+
+		client, bundle := setup(t)
+
+		type JobArgs struct {
+			JobArgsReflectKind[JobArgs]
+		}
+
+		AddWorker(client.config.Workers, WorkFunc(func(ctx context.Context, job *Job[JobArgs]) error {
+			return JobSnooze(15 * time.Minute)
+		}))
```
bgentry, line 252: Likewise, should we check anything about the job in the DB after this executes? Since this is an integration test it might be a good safety check.

bgentry (approved):
Nice, sorry I missed this coverage but thank you for adding it.

client_test.go
brandur: Sure! Added.

client_test.go
brandur: Added.

PR #32, opened 2023-11-17
Update changelog to prepare for 0.0.4 release

brandur:
Add the changes in #29 and #30 to the changelog so that we can cut a
0.0.4 release. For brevity and to keep signal high, I didn't include
changes that came in which aren't user facing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,18 +7,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.0.4] - 2023-11-17
+
+### Changed
+
+- Allow River clients to be created with a driver with `nil` database pool for use in testing.
+- Update River test helpers API to use River drivers like `riverdriver/riverpgxv5` to make them agnostic to the third party database package in use.
+- Document `Config.JobTimeout`'s default value.
+- Functionally disable the `Reindexer` queue maintenance service. It'd previously only operated on currently unused indexes anyway, indexes probably do _not_ need to be rebuilt except under fairly rare circumstances, and it needs more work to make sure it's shored up against edge cases like indexes that fail to rebuild before a clien restart.
+
 ## [0.0.3] - 2023-11-13
 
 ### Changed
 
-- Fix license detection issues with riverdriver/riverpgxv5 submodule.
-- Ensure that river requires the riverpgxv5 module with the same version.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,18 +7,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Allow River clients to be created with a driver with `nil` database pool for use in testing.
+- Update River test helpers API to use River drivers like `riverdriver/riverpgxv5` to make them agnostic to the third party database package in use.
+
 ## [0.0.3] - 2023-11-13
 
 ### Changed
 
-- Fix license detection issues with riverdriver/riverpgxv5 submodule.
-- Ensure that river requires the riverpgxv5 module with the same version.
+- Fix license detection issues with `riverdriver/riverpgxv5` submodule.
```
brandur, line 21: Just added backticks to these things.

bgentry (approved):

bgentry:
Should we get in the habit of bumping the riverpgxv5 version [here](https://github.com/riverqueue/river/blob/3440549ecdec6ed83bbd10121b98919a652a1abb/go.mod#L12) for each release? I think it affects which version gets pulled down on `go get github.com/riverqueue/river`.

brandur:
> here

Sure. What's the order of operations there exactly? Do you have to bump it before you cut the release tag? (Presumably leaving the repo in a temporarily invalid state.)

bgentry:
Yes, I think that’s correct. We should probably get in the habit of a “release prep” PR like this one which includes changelog entries and the version bump, then tag both modules immediately after merge.

brandur:
> Yes, I think that’s correct. We should probably get in the habit of a “release prep” PR like this one which includes changelog entries and the version bump, then tag both modules immediately after merge.

Alright, I'll give it a shot.

Also added changelog entries for the minor docs change and reindexer disable.

brandur:
Build passed even with the manually tweaked `go.mod` pointing to a non-existent version ... surprisingly.

bgentry:
> Build passed even with the manually tweaked `go.mod` pointing to a non-existent version ... surprisingly.

Yeah, builds within this module use the local `replace` directive in the `go.mod`, whereas that's ignored outside this module/repo. So you can point to an invalid version number in here and it just gets ignored because it's importing directly from the filesystem anyway.

PR #35, opened 2023-11-18
Comment out `replace` directive for `riverpgxv5`

brandur:
I was trying to get CI working for the River demo project [1] and found
that it's not currently possible to install the River CLI to another Go
project:

    Run go install github.com/riverqueue/river/cmd/river@latest

    go: downloading github.com/riverqueue/river v0.0.4
    go: github.com/riverqueue/river/cmd/river@latest (in github.com/riverqueue/river@v0.0.4):
            The go.mod file for the module providing named packages contains one or
            more replace directives. It must not contain directives that would cause
            it to be interpreted differently than if it were the main module.
    Error: Process completed with exit code 1.

This is roughly consistent with what I'd previously understood for
`replace`. It's useful for debugging problems during development, but I
don't think it's meant to be a permanent decoration on a project. A
decent compromise is to keep it commented out, but available in case
it's needed to debug something, which is what I've done here.

Also, probably goes to show why it's important to keep a demo app
maintained so we can notice problems of this sort.

[1] https://github.com/bgentry/riverdemo/pull/7

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -23,6 +23,8 @@ github.com/oklog/ulid/v2 v2.1.0/go.mod h1:rcEKHmBBKfef9DhnvX7y1HZBYxjXb0cP5ExxNs
 github.com/pborman/getopt v0.0.0-20170112200414-7148bc3a4c30/go.mod h1:85jBQOZwpVEaDAr341tbn15RS4fCAsIst0qp7i8ex1o=
 github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
+github.com/riverqueue/river/riverdriver/riverpgxv5 v0.0.4 h1:336o1gbjpsSe1I9Es0eoJkBagXTmkXdzZ1M8j9kwLUU=
+github.com/riverqueue/river/riverdriver/riverpgxv5 v0.0.4/go.mod h1:k6hsPkW9Fl3qURzyLHbvxUCqWDpit0WrZ3oEaKezD3E=
 github.com/robfig/cron/v3 v3.0.1 h1:WdRxkvbJztn8LMz/QEvLN5sBU+xKpSqwwUO1Pjr4qDs=
 github.com/robfig/cron/v3 v3.0.1/go.mod h1:eQICP3HwyT7UooqI/z+Ov+PtYAWygg1TEWWzGIFLtro=
 github.com/rogpeppe/go-internal v1.11.0 h1:cWPaGQEPrBb5/AsnsZesgZZ9yb1OQ+GOISoDNXVBh4M=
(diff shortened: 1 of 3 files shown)


brandur:
@bgentry I suppose the other thing we could do here is to also make `./cmd/river` its own Go module.

I guess I don't really care that much as long as we have something that works. Do you want to play around with this a bit and see if you can find something that works? Feel free to push to https://github.com/bgentry/riverdemo/pull/7 to see if you can get that working.

brandur:
Doh, the `./cmd/river` submodule doesn't work for now, I think because it's still using internal River packages. We can likely fix this by exposing a Go API for migrations (which would largely just involve moving the migrator to a public package since it should be mostly good), but it'll take a little more effort to get there.

```
$ go mod tidy
go: github.com/riverqueue/river/cmd/river: ambiguous import: found package github.com/riverqueue/river/cmd/river in multiple modules:
        github.com/riverqueue/river v0.0.4 (/Users/brandur/Documents/go/pkg/mod/github.com/riverqueue/river@v0.0.4/cmd/river)
```

go.sum
```diff
@@ -23,6 +23,8 @@ github.com/oklog/ulid/v2 v2.1.0/go.mod h1:rcEKHmBBKfef9DhnvX7y1HZBYxjXb0cP5ExxNs
 github.com/pborman/getopt v0.0.0-20170112200414-7148bc3a4c30/go.mod h1:85jBQOZwpVEaDAr341tbn15RS4fCAsIst0qp7i8ex1o=
 github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
+github.com/riverqueue/river/riverdriver/riverpgxv5 v0.0.4 h1:336o1gbjpsSe1I9Es0eoJkBagXTmkXdzZ1M8j9kwLUU=
+github.com/riverqueue/river/riverdriver/riverpgxv5 v0.0.4/go.mod h1:k6hsPkW9Fl3qURzyLHbvxUCqWDpit0WrZ3oEaKezD3E=
```
bgentry, line 27: 1. What's the process for cutting a new release if we go this route?
2. As part of cutting a new release, will the `go.mod` and `go.sum` disagree with each other around the required version of `riverpgxv5`?
3. After shipping this change, if a user installs `river@latest`, will they also get the latest `riverpgxv5` no matter what this `go.sum` says?

bgentry (approved):

go.sum
brandur: > 1. What's the process for cutting a new release if we go this route?

I think either:

1. Use the same release process as before and accept that the build is temporarily broken.
2. Two phase release where river is tagged and after that's in, riverpgxv5 is tagged.

> 2. As part of cutting a new release, will the `go.mod` and `go.sum` disagree with each other around the required version of `riverpgxv5`?

I don't know.

> 3. After shipping this change, if a user installs `river@latest`, will they also get the latest `riverpgxv5` no matter what this `go.sum` says?

I think the `go.sum` is only locking for the current module, but not 100% sure about this.

brandur:
Doh, this is blocking me from getting another project online, so going to merge for now. I'll play around with the release process and update docs according to what I find.

PR #40, opened 2023-11-19
Add CI check to make sure River CLI is installable

brandur:
Pertaining to the problem described in #35, seems like a good idea to
have a CI check make sure that nothing we've done with Go modules has
made the CLI not installable.

We have two variants:

* Install within a Go Module. This is theoretically nice because it'd
  get the same version of the River CLI as the River you're using, but
  has a very large downside currently in that it requires you to add
  the CLI's dependencies to your stack before it works. e.g. Cobra.

* Install outside a Go Module with the equivalent of `@latest` (uses the
  ref of the branch being built in CI). This is currently broken until a
  fix like #35 can come in.

Long term, this job may become too much of a maintenance headache
because Go has some really annoying caching behavior around modules, but
I didn't experience that while writing it, so it might be worthwhile
seeing whether this turns out to be valuable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,79 @@ jobs:
       - name: river migrate-down
         run: ./river migrate-down --database-url $DATABASE_URL --max-steps 100
 
+  cli_installable:
+    runs-on: ubuntu-latest
+    timeout-minutes: 3
+
+    steps:
+      - uses: actions/setup-go@v4
+        with:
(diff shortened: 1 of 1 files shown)


brandur:
Moved this to draft while we consider what to do.

skydiver1980 (approved):

brandur:
I'll bring this back if it looks like it's going to be a problem again. Closing for now.

PR #41, opened 2023-11-19
Tee up the changelog for v0.0.5

brandur:
Changelog for v0.0.5 in which we mention removing the `replace`
directive as in #35.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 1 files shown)


brandur:
@bgentry I didn't bump `go.mod`'s `riverpgxv5` version here — thought about it, but I don't think having a failing build is a good idea and it may cause caching issues in the Go proxy. It'll have to be bumped in a later change, although since nothing in the driver changed, it probably doesn't matter if it is right now. Can experiment with this a bit later out-of-project to see if we can find an improved workflow.

bgentry (approved):

brandur:
Thanks!

PR #43, opened 2023-11-20
Document `AttemptError` and rename `AttemptError.Num` to `Attempt`

brandur:
Two error-related changes:

* Document `AttemptError` and all its fields. This one's been on my list
  for a while and just brings it up to par with other exported structs
  like `JobRow`.

* Rename `AttemptError.Num` to `Attempt`. I hadn't gone in attending to
  change this one, but as I was looking into what `Num` was, I realized
  that it'd more obvious if it matched up with the field that it comes
  from on `JobRow`. By having the same name as `JobRow.Attempt`, it's
  very obvious that these two things are related.

### DIFF
--- a/rivertype/job_row.go
+++ b/rivertype/job_row.go
@@ -96,6 +96,8 @@ type JobRow struct {
 	// metadata []byte
 }
 
+// JobState is the state of a job. Jobs start as `available` or `scheduled`, and
+// if all goes well eventually transition to `completed` as they're worked.
 type JobState string
 
 const (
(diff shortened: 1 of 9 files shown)


rivertype/job_row.go
```diff
@@ -108,9 +110,21 @@ const (
...
-	At    time.Time `json:"at"`
-	Error string    `json:"error"`
-	Num   int       `json:"num"`
-	Trace string    `json:"trace"`
+	// At is the time at which the error occurred.
+	At time.Time `json:"at"`
+
+	// Attempt is the attempt number on which the error occurred (maps to
+	// Attempt on a job row).
+	Attempt int `json:"attempt"`
```
brandur, line 121: @bgentry Wanted to float this rename of `Num` to `Attempt`. I had to look up what was being set to `Num` and realized that it'd be more obvious if this had the same name as `JobRow.Attempt`, which is what produces this number. I feel like it's just one of these things that we had since the beginning and never reevaluated. What do you think?

bgentry: I think it’s a nice thinking improvement at the small cost of a few more bytes for each error in the db. Given we’re serializing stack traces that’s an insignificant difference.

Related, should we be limiting the buffer size for stack traces somehow? These could result in very large payloads.

bgentry (approved):

rivertype/job_row.go
brandur: +1.

And yeah ... we should take a look at just what the average size of one of those panic traces looks like again. I guess the good news is that errors should be the rarer case, panics should be the rarer case within errors, and long strings like this all get shoved in TOAST anyway, so it shouldn't affect the actual table itself every much.

But there might be something relatively smart/easy, low-hanging that we can do to improve things. I'll add a TODO note.

PR #45, opened 2023-11-20
Rename constants to make adjectives like "Default" or "Min" suffix instead of prefix

brandur:
One last last-second API change I wanted to float: rename constants so
that adjectives like "Default" and "Min" become name suffixes instead of
prefixes. So e.g. `DefaultFetchCooldown` becomes `FetchCooldownDefault`.

The primary benefit in my mind is that constants that are related to
each other in referencing the same value, now sort together
alphabetically. Take for example:

    const (
            FetchCooldownDefault = 100 * time.Millisecond
            FetchCooldownMin     = 1 * time.Millisecond

Previously, these were organized together, but didn't actually line up
if the list were to be sorted.

Another benefit is better autocomplete: when trying to autocomplete a
default value, previously to disambiguate from all the other ones you'd
have to type the whole word "Default" which is common to all of them.
But with the prefix switched to a suffix, if looking for say the default
for max attempts, you'd get a match within just a couple characters.

Probably the most contentious part of this is that `QueueDefault` flips,
which is the one constant that's actually referenced a fair bit from
user code. e.g.

            river.QueueDefault: {MaxWorkers: 100},
    },

This may be a tiny bit worse in terms of how it rolls off the tongue,
but I've been staring at it a while now, and I think it's something
you'd get used to. Also, it's kind of nice how it fits better with names
like `QueueConfig` which are present in the same invocation.


brandur:
@bgentry One last color of the shed change I wanted to try and get in before we go API stable ... thoughts?

bgentry (approved):
I give you approximately 2 days until you find another constant you’d like to rename, but it will be too late 😉

There is a small downside to this change, which is that if I’m trying to reference a default I have to correctly guess what it’s called instead of starting with `Default` and getting a list of all defaults. But I think it’s minor and this is a small win on net, particularly for grouping related constants.

Any corresponding doc fixes for this one?

brandur:
> I give you approximately 2 days until you find another constant you’d like to rename, but it will be too late 😉

lol. God, that's probably true.

> There is a small downside to this change, which is that if I’m trying to reference a default I have to correctly guess what it’s called instead of starting with `Default` and getting a list of all defaults. But I think it’s minor and this is a small win on net, particularly for grouping related constants.

Yeah, true, it is a bit of a trade of there. I'm sort of coming at it from the angle that I'm just looking for a particular "thing" (e.g. job timeout) compared to looking for some kind of default.

> Any corresponding doc fixes for this one?

Yep. I updated all the constants in comments I could find, and will take updating docs in the homepage repo.

Thanks!

PR #51, opened 2023-11-21
add thank you note to readme

bgentry:


brandur (approved):
Only possible concern is this does introduce some content drift between the README and Godoc file. Probably okay, but something to keep an eye on in case the project tries to generate one from the other.

PR #53, opened 2023-11-21
Default tags to an empty slice, make not null

### LINKED ISSUE #46: Default tags to [] in the database
From https://github.com/riverqueue/river/pull/44#issuecomment-1819170445:

> The other strong contender here is to make tags NOT NULL and DEFAULT [] in the database, so it becomes always not-nil and an empty array on any job row that comes back.

bgentry:
Fixes #46.

To make this work, I had to reintroduce an allocation in `JobInsertMany`. Let me know if you have other ideas.


brandur (approved):

internal/dbadapter/db_adapter.go
```diff
@@ -298,6 +298,10 @@ func (a *StandardAdapter) JobInsertManyTx(ctx context.Context, tx pgx.Tx, params
 			metadata = []byte("{}")
 		}
 
+		tags := params.Tags
+		if tags == nil {
+			tags = []string{}
+		}
```
brandur, line 304: Doh. COPY/FROM.

internal/dbmigrate/003_river_job_tags_non_null.down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE river_job ALTER COLUMN tags DROP NOT NULL,
```
brandur, line 1: Doh, this is still kinda bad and could be changed, but you still need to add these to a list of migrations here along with `go:embed` statements:

https://github.com/riverqueue/river/blob/master/internal/dbmigrate/db_migrate.go#L47-L48

We should fix that, but it was the fastest/easiest way to introduce migrations at the time.

brandur (commented):
Nice. Okay only caveat to point out is that if you upgrade your River dependency, it's not necessarily obvious that you have to re-migrate.

Other projects have migration guides and such. Probably overkill given the v0 nature, but some people may still get broken accidentally.

brandur (approved):
lgtm.

PR #64, opened 2023-11-23
fix v003 migration: set default + update rows before set not null

### LINKED ISSUE #63: v0.0.8 migration fails when there are existing jobs with `NULL` tags
I've got a database with a bunch of jobs still around in `completed` state, which causes the migration to fail:

```
failed: error applying version 003 [UP]: ERROR: column "tags" of relation "river_job" contains null values (SQLSTATE 23502)
```

Seems like the migration could update all existing rows to have `[]`, assuming there's a way to do that without race conditions. I guess you'd have to [lock the entire table](https://www.postgresql.org/docs/current/sql-lock.html)?

In my case I think I'll just `TRUNCATE` or `DELETE FROM` the full `river_job` table prior to running the migration.

bgentry:
The migration introduced in #53 / c5bd949 was flawed because it attempted to set a default and also set not null on an existing field all in one statement. This of course won't work if there are any existing rows in the table, because those need to be updated to be non-null before adding the not null constraint.

This change is made in place on the existing migration (rather than a new one) because it fixes the fact that the migration simply can't succeed on existing installs.

Note that there is still a chance of this failing on an existing system with a high write throughput. In the future @brandur we may need a different strategy to make this safe, such as splitting the change into multiple migrations. Hopefully we don't need to add any additional not null constraints though. 🤞

Fixes #63.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.0.9] - 2023-11-23
+
+### Fixed
+
+- **DATABASE MIGRATION**: Database schema v3 was introduced in v0.0.8 and contained an obvious flaw preventing it from running against existing tables. This migration was altered to execute the migration in multiple steps.
+
 ## [0.0.8] - 2023-11-21
 
 ### Changed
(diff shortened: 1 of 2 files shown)


brandur (approved):
Oops, shoulda caught that one.

You might want to add a changelog here too since it'll need a new release cut so people can access the change.

bgentry:
Yep, we both should have caught it. Turns out it was a wild week 😅 I was in the midst of adding the changelog but also wanted to test this on an existing riverdemo installation to be certain it worked. That all went smoothly, launched and migrated it up to v2, then added some jobs, watched it fail on `master` and succeed with this commit :v:

PR #67, opened 2023-11-26
New `rivermigrate` Go API for running migrations from code + target version support

brandur:
Here, as requested by several users, add a Go API that's able to run
migrations as an alternative to migrating via the CLI. We add a new
package `rivermigrate` to mirror the conventions of `riverdriver` and
`rivertest`.

`rivermigrate` is largely just a refactor of the existing internal
package `internal/dbmigrate` with a couple important tweaks:

* We give it the ability to take a driver the same way as we do for
  `river.Client` and `rivertest`.

* To mirror client and `rivertest` we add a transaction-specific variant
  that takes a `TTx` (`MigrateTx`).

* Because we now have the non-transaction and transaction variants, I
  take away the `Down`/`Up` function distinctions, and make up/down a
  parameter to `Migrate`/`MigrateTx` instead (so we only need two
  functions instead of four).

This still can't support Goose because it takes an `sql.DB`, but I think
we can get there by adding a minimal driver for the `sql` package that
can run migrations and nothing else. More of this to be explored later.


rivermigrate/river_migrate.go
```diff
@@ -0,0 +1,502 @@
...
+
+// Migrate migrates the database in the given direction (up or down). The opts
+// parameter may be omitted for convenience.
+//
+// By default, applies all outstanding migrations possible in either direction.
+// When migrating up all outstanding migrations are applied, and when migrating
+// down all existing migrations are unapplied.
+//
+// When migrating down, use with caution. MigrateOpts.MaxSteps should be set to
+// 1 to only migrate down one step.
```
bgentry, line 168: This seems like a bit of a footgun. I would have been inclined to default to running a single migration at a time just to avoid this. Do you think it's too risky?

At the very least we might want the examples to show the single-migration usage.

rivermigrate/river_migrate.go
```diff
@@ -0,0 +1,502 @@
...
+//
+//	res, err := migrator.Migrate(ctx, rivermigrate.DirectionUp, nil)
+//	if err != nil {
+//		// handle error
+//	}
+func (m *Migrator[TTx]) Migrate(ctx context.Context, direction Direction, opts *MigrateOpts) (*MigrateResult, error) {
```
bgentry, line 174: Related to the other comment, I'm not quite sure about the API here. I'm thinking in terms of trying to make this foolproof and as difficult as possible to accidentally mess up your prod database, while also being sure exactly what's being run at a given step.

Rather than stating _how many_ migrations to run, do you think it might be safer to specify _which_ migration you want to run? This would require a bit of a different API, because it'd be something like `RunUp(ctx, 1)` or `Run(ctx, rivermigrate.DirectionUp, 3)`. That forces the use case of running multiple migrations into a different API like `RunN()`, but I expect that use case to be more unusual. Also if you went that route, the `N` arg in `RunN()` could be an explicit arg and not something with a default zero value.

My thought process here is that if I'm running an app with River and I see that there's a new version out, I'll check the changelog or website for notes about new migrations in that version. If I see one, I'm going to make a new migration file in my goose/golang-migrate/etc migration setup where I have it call out to `rivermigrate`. I would want to be extra sure that I'm running exactly the migration I expect to be running so that I'm sure I'm prepared and have followed any related instructions if necessary.

rivermigrate/river_migrate.go
brandur: I could see running one at a time to be somewhat defensible, but my fear is that it'll be the wrong thing for what might end up being the most common use of the Go migration API, which is to raise tables for test and development purposes.

In these cases what you most want out of the migrate API is for it to give you a fully functional River schema. Running one migration at a time doesn't do that, so you'd either have to hammer it into place by setting `MaxSteps: 1000` or by setting a new `TargetVersion` every time a new River DB version comes out (not convenient).

I think I'd lean toward documenting this and making sure to have `MaxSteps: 1` in most examples as you suggest.

rivermigrate/river_migrate.go
brandur: So both use cases are supported, with what you describe being the `TargetVersion` option. I originally had this even closer to what you described as just `Version` that'd run a single version one off, but I changed it after thinking about the various footguns and ergonomic issues that fell out of being able to target versions that are multiple steps beyond your current version, and would therefore have to error or be applied out of order. `TargetVersion` applies a target version and any versions up to it that are necessary based on your current database state.

I think this would be our general recommendation for usage when writing upgrade guides for specific versions, so the up migration would look like:

``` go
rivermigrate.New(...).Migrate(ctx, rivermigrate.DirectionUp, &riverigrateOpts{
     TargetVersion: 4,
}
```

And then the down migration something like:

``` go
rivermigrate.New(...).Migrate(ctx, rivermigrate.DirectionUp, &riverigrateOpts{
     TargetVersion: 3,
}

// or (neither of these is strictly right when upgrading multiple versions at once — the user would have to make sure to set their original version)

rivermigrate.New(...).Migrate(ctx, rivermigrate.DirectionUp, &riverigrateOpts{
     MaxSteps: 1,
}
```

See also the note above, but I didn't want `TargetVersion` to be the only API because I think there's going to be demand to use this in test suites and one-off small projects (which don't have a full migration line) where you just want to have River raise a usable schema with no hassle and without needing to update code when new River versions come out. Those can just use the really easy:

``` go
rivermigrate.New(...).Migrate(ctx, rivermigrate.DirectionUp, nil)
```

bgentry: Ah yeah I missed this on the first pass, my bad! I think you've highlighted that there are indeed two distinct use cases here: get a working DB as fast as possible, vs give me control and certainty for integrating with my existing migration setup. Iff folks already have an existing DB they likely already have an existing migration setup, which for dev and tests could either consist of stepping through migrations one-at-a-time or doing a bulk schema load.

Since this API supports both, that's great! I want to make sure we at least document these conventions clearly and ensure there aren't any major footguns in them (per above thread):

1. I'm just getting started with River in an existing migration-driven project. What code do I add to a new migration file to go from 0 to latest?
2. I've already adopted river per the above with an older migration version (say, v2). What do I put in a new migration to ensure I go precisely from v2 to v3, no more and no less?
3. I've already adopted river per the above with an older migration version (say, v2) and multiple migrations have come out since my last upgrade. What do I put in a new migration to ensure I go precisely from v2 to i.e. v4, no more and no less?
4. I don't want to integrate my River tables into an existing migration system and just want to use a CLI to quickly bring River to the latest version from wherever it is.
5. I don't want to integrate my River tables into an existing migration system, but I still want the tight control and guarantees from (2) and (3) above.

For (2) and (3) above, would the recommendation be to use _both_ `TargetVersion` and `MaxSteps` together? Or how would I guarantee I'm running only the code from migration 3's up migration, no matter what else I've done?

rivermigrate/river_migrate.go
```diff
@@ -0,0 +1,502 @@
...
+type MigrateOpts struct {
+	// MaxSteps is the maximum number of migrations to apply either up or down.
+	// Leave zero for an unlimited number. Set to -1 to apply no migrations (for
+	// testing/checking purposes).
+	MaxSteps int
+
+	// TargetVersion is a specific migration version to apply migrations to. The
+	// version must exist and it must be in the possible list of migrations to
+	// apply. e.g. If requesting an up migration with version 3, version 3 not
+	// already be applied.
```
bgentry, line 126:
```suggestion
	// apply. e.g. If requesting an up migration with version 3, version 3 must
	// not already be applied.
```

rivermigrate/river_migrate.go
brandur: > For (2) and (3) above, would the recommendation be to use _both_ `TargetVersion` and `MaxSteps` together? Or how would I guarantee I'm running only the code from migration 3's up migration, no matter what else I've done?

No, not recommended — by including a `MaxSteps` with everything, you're making things strictly more dangerous.

Say you were trying to get to version 3, and you thought you were on version 2, so you had `MaxSteps: 1`. But actually, you were on version 1! So `TargetVersion: 3, MaxSteps: 1` actually takes you to version 2 and then River might not start up properly.

You might combine these two options while testing something, but you wouldn't want to do so as part of any upgrade guides or the like. `TargetVersion` knows what it's trying to do and will do the right thing based on the current state of the local database.

> For down migrations, I want to be 100% sure that I'm undoing exactly the same steps as I specified above, no more and no less. Does the API let me guarantee that?

Nope, and I don't think there's any way to do so without farming the work out to the user themselves. We can tell people that if they know they're incrementing exactly one version, they can use a `MaxSteps: 1` , but otherwise they'll have to look at their current version and include that as `TargetVersion: <version>`.

One thing I should point out: you generally should never get to the point where you have to run down migrations for any app in production, because if you do, the battle's already lost. River may be an exception here to some degree, but chances are that down migrations are poorly tested and may be unexpectedly destructive/lossy. It's probably less risky to fix forward rather than run them in most cases.

bgentry: > No, not recommended — by including a `MaxSteps` with everything, you're making things strictly more dangerous.
> 
> Say you were trying to get to version 3, and you thought you were on version 2, so you had `MaxSteps: 1`. But actually, you were on version 1! So `TargetVersion: 3, MaxSteps: 1` actually takes you to version 2 and then River might not start up properly.
> 
> You might combine these two options while testing something, but you wouldn't want to do so as part of any upgrade guides or the like. `TargetVersion` knows what it's trying to do and will do the right thing based on the current state of the local database.

Alright, this is a fair point. If you're trying to get to v4 (the current version) and you know this, it should be safe to let the migration engine do what it needs to in order to get to that version. On the down migrations, you could get bit by not migrating down far enough (if you were actually further behind than you thought you were and set a `MaxSteps`) but that could be uncovered through actually testing your up + down migrations together.

The overall design here should be workable as long as we document all these common use cases clearly.

bgentry (approved):
I'm approving the PR broadly, nice work on this. However I do want to reiterate that I think defaulting down migrations to unlimited steps is super dangerous and we should think about adding a failsafe for that scenario. I can't think of many (any?) reasons why someone would really _ever_ want to do that, let alone have it be the default. Thoughts on that remaining bit?

rivermigrate/river_migrate.go
brandur: Fixed, thanks!

bgentry (approved):
I agree the different behavior on `down` is asymmetric and feels a bit odd, but if it saves just one person from accidentally nuking their jobs table by mistake it seems worthwhile IMO. Most people shouldn't be migrating down in a non-dev environment to begin with.

Nice work on this :shipit:

brandur:
Cut release 0.0.12. I'll write up some additional docs for the site too.

PR #70, opened 2023-11-26
Add example for a job scheduled in the future

brandur:
It came up on a GitHub thread that someone was confused about how to
schedule a job to be worked in the future. You can figure this out from
the godoc, but it doesn't hurt to add an explicit example for how to
schedule a job for the future.


brandur:
Actually, turned out that we already have a docs page for scheduled jobs, so no need to add that.

PR #71, opened 2023-11-26
Link up scheduled job docs from README and Godoc + changelog entry

brandur:
Small follow up to #70 in which we add a link to scheduled jobs from the
features list in the README and Godoc (making it really easy to find by
scanning in either place), along with a changelog entry for the new
example.


bgentry (approved):

brandur:
Cut 0.0.10.

PR #79, opened 2023-11-27
respect provided context in Stop + StopAndCancel

### LINKED ISSUE #78: Support abandoning jobs that take too long to cancel
### Context

Hiya, thanks for making River! Really digging it so far.

I have an app that queues jobs that I don't really need to wait to complete before exiting, even after cancelling. I just want to guarantee fast deploys and let jobs just get tossed around like a hot potato during a rolling deploy.

Before I continue: this isn't blocking me, and it's a nuanced topic on an edge case within an edge case, so no rush.

`StopAndCancel` is almost what I want. Cancelled jobs end up in a `retryable` state and they either get picked back up elsewhere or recovered when the instance comes back. The one case it doesn't handle is when jobs take too long to _stop_ (typically by not respecting `ctx.Done()`, likely a bug).

I tried to pass a timeout `ctx` to `StopAndCancel`, but it doesn't currently respect it:

```go
stopCtx, stop := context.WithTimeout(context.Background(), 10*time.Second)
defer stop()
if err := db.workerClient.StopAndCancel(stopCtx); err != nil {
	slog.Error("failed to stop and cancel workers", "error", err)
}
if stopCtx.Err() != nil {
	slog.Warn("timed out cleaning up jobs")
}
```

With a job that does `time.Sleep(time.Minute)` the `StopAndCancel` will just wait the full 1 minute instead of being interrupted after 10 seconds. This seems like a simple bug, happy to submit a PR.

Even if it did respect the timeout, there's a secondary problem: if we bail out at that point, the worker's jobs will be left in a `running` state. At this point the job is stuck, and if you have uniqueness rules like I do that'll prevent future attempts too. As far as I can tell the only recourse at this point is to delete the job from the database or set its state to `retryable`.

### Proposal

This might be problematic for other use cases, but at least for mine, this would be a big help:

1. `StopAndCancel` respects ctx cancellation
2. When `StopAndCancel` times out it marks all `running` jobs on the local worker as `retryable`

Alternatively, having some way to do (2) independently would be fine. For example I could do a `db.workerClient.Abandon(stopCtx)` in the `if stopCtx.Err() != nil {` branch above.

This would allow graceful shutdown in most cases, prevent hanging in the "truly stuck" case, and allow jobs to be retried in all cases except `kill -9`.

### Alternative

The root of the problem here is really the stuck `running` jobs. I would also be OK with leaving everything as-is and instead just enforcing a stop timeout at the infra layer, letting the app get `kill -9`ed, and having River automatically detect when a `running` job is abandoned so things aren't stuck forever.

bgentry:
An earlier refactor caused these graceful shutdown routines to stop respecting the provided context. No matter what happened with the provided context, they would continue blocking, though they would ultimately return the context's error upon return.

With this change, both stop methods will return early if the provided context is cancelled or timed out _prior_ to the clean shutdown completing. This makes it straightforward to implement a shutdown flow which first calls `Stop` with a context that expires after 5 seconds, and after that call `StopAndCancel`. If the shutdown _still_ isn't complete, the user can choose to exit anyway.

Partially fixes #78. cc @vito

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `Stop` and `StopAndCancel` have been changed to respect the provided context argument. When that context is cancelled or times out, those methods will now immediately return with the context's error, even if the Client's shutdown has not yet completed. Apps may need to adjust their graceful shutdown logic to account for this. PR #79.
+
 ## [0.0.10] - 2023-11-26
 
 ### Added
--- a/example_graceful_shutdown_test.go
+++ b/example_graceful_shutdown_test.go
@@ -105,52 +105,40 @@ func Example_gracefulShutdown() {
 		<-sigintOrTerm
 		fmt.Printf("Received SIGINT/SIGTERM; initiating soft stop (try to wait for jobs to finish)\n")
 
-		softStopSucceeded := make(chan struct{})
+		softStopCtx, softStopCtxCancel := context.WithTimeout(ctx, 10*time.Second)
+		defer softStopCtxCancel()
+
 		go func() {
-			if err := riverClient.Stop(ctx); err != nil {
-				if !errors.Is(err, context.Canceled) {
-					panic(err)
-				}
+			select {
+			case <-sigintOrTerm:
+				fmt.Printf("Received SIGINT/SIGTERM again; initiating hard stop (cancel everything)\n")
+				softStopCtxCancel()
+			case <-softStopCtx.Done():
+				fmt.Printf("Soft stop timeout; initiating hard stop (cancel everything)\n")
 			}
-			close(softStopSucceeded)
 		}()
 
-		// Wait for soft stop to succeed, or another SIGINT/SIGTERM.
-		select {
-		case <-sigintOrTerm:
-			fmt.Printf("Received SIGINT/SIGTERM again; initiating hard stop (cancel everything)\n")
-
-		case <-time.After(10 * time.Second):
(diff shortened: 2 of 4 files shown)


example_graceful_shutdown_test.go
bgentry, line 1: I took some liberty in trying to restructure this example based on what I think the intent was. With these methods now respecting the context, you can use a context timeout to make them return early and don't need a separate timer for this purpose. However it was a bit tricky deciding how to also listen for an additional signal to immediately escalate to a hard stop.

It does at least end up being a fair bit shorter after these changes, but please feel free to suggest any improvements based on how you might want this to work.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
```
bgentry, line 10: Do you think this should be noted as "breaking" given the potential impact on callers?

brandur: IMO, it might be better to describe how an app might be broken by this change and why we did it.

Might not be the best habit to get into to start including a regular "Breaking" section in the semver doc heh.

example_graceful_shutdown_test.go
```diff
@@ -105,52 +105,40 @@ func Example_gracefulShutdown() {
 		<-sigintOrTerm
 		fmt.Printf("Received SIGINT/SIGTERM; initiating soft stop (try to wait for jobs to finish)\n")
 
-		softStopSucceeded := make(chan struct{})
+		softStopCtx, softStopCtxCancel := context.WithTimeout(ctx, 10*time.Second)
```
brandur, line 108: This example's fully reproduced in the docs, so worth updating it over there too.

brandur (approved):

PR #81, opened 2023-11-28
Fix grammar that's wrong-by-template in examples

brandur:
Small fix to grammar on the comment around the subscribe channel in many
of the example tests.


brandur:
Thx.

PR #82, opened 2023-11-29
ci: Use setup-sqlc@v4 action

andrewmbenton:
Just noticed that your sqlc workflow would be simpler with the [setup-sqlc GitHub action](https://github.com/sqlc-dev/setup-sqlc).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -191,28 +191,17 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 2
 
-    env:
-      SQLC_VERSION: 1.22.0
-
     steps:
-      - name: Create BIN_PATH and add to PATH
-        run: |
-          mkdir -p "$BIN_PATH"
-          echo "$BIN_PATH" >> $GITHUB_PATH
-
-      - name: Install sqlc
-        run: |
-          curl -L https://github.com/kyleconroy/sqlc/releases/download/v${{ env.SQLC_VERSION }}/sqlc_${{ env.SQLC_VERSION }}_linux_amd64.tar.gz | tar -xz -C $BIN_PATH
-          chmod +x $BIN_PATH/sqlc
-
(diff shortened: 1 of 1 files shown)


brandur:
@andrewmbenton Thanks!

Would you mind doing a no-op `git commit --amend` and force pushing again? Something odd seems to have happened with the GitHub Actions checks here — they're still in limbo an hour later and GitHub is not currently reporting any status events.

bgentry:
@brandur I think we might need to approve this first for a first-time contributor before the actions will run
Screenshot 2023-11-28 at 8 18 23 PM

bgentry (approved):

brandur:
@bgentry So I have this feature on for other repos too, but usually there's an "approve CI run" button you can press. I'm not sure if I'm blind or there's a bug somewhere, but I'm not seeing it anywhere in my UI. Do you see it?

bgentry:
I've been looking everywhere to find a setting that will make that button show up, afaict all the settings look correct but I don't see any way to approve the workflow per [the docs](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks#approving-workflow-runs-on-a-pull-request-from-a-public-fork) 😖

brandur:
WTH. I think it might be a bug ... I've seen the button hundreds of times before. It's just not here right now.

bgentry:
@brandur I think we might need the `pull_request` trigger on the CI workflow: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull-request-events-for-forked-repositories

brandur:
@bgentry Doh. Okay NM, I guess that's it. Good idea to actually RTFM.

bgentry:
Thank you @andrewmbenton 🙏

andrewmbenton:
> Thank you @andrewmbenton 🙏

Happy to help work out the configuration kinks early on.

PR #88, opened 2023-12-02
Move check on zero workers from client constructor to `Start`

### LINKED ISSUE #87: Passing a `riverClient` to a worker?
I am trying to trigger a job from within another job (think: send a notification when a long-running job is complete). This requires that I have access to the `riverClient` in the worker context.

I could pass the client to the worker, like

```go
river.AddWorker[action.SomeArgs](workers, &action.SomeWorker{DbPool: dbPool, RiverClient: riverClient})
```

but that creates a chicken-egg problem because I need an initialized `Workers` to create the `riverClient`.

Adding the workers after initialization does not work either because `NewClient` does not accept an empty `Workers`:

```go
workers := river.NewWorkers()

riverClient, err := river.NewClient[pgx.Tx](riverpgxv5.New(dbPool.Pool), &river.Config{
    Workers: workers,
})
if err != nil {
    panic(err)
}

river.AddWorker[action.SomeArgs](workers, &action.SomeWorker{DbPool: dbPool, RiverClient: riverClient})

// panic: at least one Worker must be added to the Workers bundle
```

Adding some worker that does not depend on the client does actually work, but is kind of awkward:

```go
workers := river.NewWorkers()
river.AddWorker[action.DummyArgs](workers, &action.DummyWorker{})

riverClient, err := river.NewClient[pgx.Tx](riverpgxv5.New(dbPool.Pool), &river.Config{
    Workers: workers,
})
if err != nil {
    panic(err)
}

river.AddWorker[action.SomeArgs](workers, &action.SomeWorker{DbPool: dbPool, RiverClient: riverClient})

// -> both workers work as expected
```

Is there any other "pattern" to achieve this? If not, I think the status quo should at least be documented somewhere.

brandur:
This one's in pursuit of trying to solve #87, where it's difficult to
inject a River client into a worker because trying to initialize a
client with the workers bundle empty is an error, creating a chicken and
egg problem.

There's no real reason to disallow a zero worker bundle from the
constructor, and in fact a lot of our tests already add additional
workers after the client was originally initialized (although
`newTestClient` injects a default worker, which is why there's no
error). If it's a useful pattern for us, it's probably useful for other
users too.

Here, move the zero workers check from the constructor over to the
`Start` function instead. While it seems okay to initialize a client
without workers, starting it without any does seem like a potential
problem that we'd want to keep an eye out for.

Fixes #87.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `NewClient` no longer errors if it was provided a workers bundle with zero workers. Instead, that check's been moved to `Client.Start` instead. This allows adding workers to a bundle that'd like to reference a River client by letting `AddWorker` be invoked after a client reference is available from `NewClient`. PR #87.
 - `Stop` and `StopAndCancel` have been changed to respect the provided context argument. When that context is cancelled or times out, those methods will now immediately return with the context's error, even if the Client's shutdown has not yet completed. Apps may need to adjust their graceful shutdown logic to account for this. PR #79.
 
 ## [0.0.10] - 2023-11-26
--- a/client.go
+++ b/client.go
@@ -210,9 +210,6 @@ func (c *Config) validate() error {
 	if c.Workers == nil && c.Queues != nil {
 		return errors.New("Workers must be set if Queues is set")
 	}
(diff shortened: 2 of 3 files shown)


bgentry (approved):

brandur:
Thanks! Also added a changelog.

dhermes:
Awesome change, thanks!

PR #90, opened 2023-12-02
expose Client.Stopped() to simplify waiting on graceful shutdown

bgentry:
This follows [some discussion](https://github.com/riverqueue/river/pull/80#issuecomment-1837261256) in #80 about how this would be a small improvement because it eliminates the need for users to make their own channel for the same purpose.

### DIFF
--- a/client.go
+++ b/client.go
@@ -705,6 +705,14 @@ func (c *Client[TTx]) StopAndCancel(ctx context.Context) error {
 	return c.awaitStop(ctx)
 }
 
+// Stopped returns a channel that will be closed when the Client has stopped.
+// It can be used to wait for a graceful shutdown to complete.
+//
+// It is not affected by any contexts passed to Stop or StopAndCancel.
+func (c *Client[TTx]) Stopped() <-chan struct{} {
+	return c.stopComplete
+}
+
 // Subscribe subscribes to the provided kinds of events that occur within the
 // client, like EventKindJobCompleted for when a job completes.
 //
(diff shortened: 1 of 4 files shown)


client_test.go
```diff
@@ -492,6 +492,46 @@ func Test_Client_Stop(t *testing.T) {
...
+		stopCtx, stopCancel := context.WithTimeout(ctx, 5*time.Second)
+		t.Cleanup(stopCancel)
+
+		if err := client.StopAndCancel(stopCtx); err != nil {
+			t.Fatal(err)
+		}
+
+		riverinternaltest.WaitOrTimeout(t, client.Stopped())
+	})
+}
```
brandur, line 533: Probably not worth rewriting for this, but it might not be the worst idea to put this sort of thing in the general `Test_Client` suite for other changes of this sort. Normalizes code a little more, and needs less bespoke set up code for every client function.

client.go
```diff
@@ -705,6 +705,12 @@ func (c *Client[TTx]) StopAndCancel(ctx context.Context) error {
 	return c.awaitStop(ctx)
 }
 
+// Stopped returns a channel that will be closed when the Client has stopped.
+// It can be used to wait for a graceful shutdown to complete.
```
brandur, line 709: Maybe add a sentence on what happens with it when a stop context is cancelled?

brandur (approved):

client.go
bgentry: Done ✅

PR #93, opened 2023-12-03
golangci-lint version upgrade + fixes

bgentry:
Upgrades golangci-lint and fixes the many issues that popped up in the process. Most of them seem unequivocally good, though a few of the constant ones are questionable. I decided to just take the advice rather than adding a bunch of `nolint` comments though.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -1146,7 +1146,7 @@ func Test_Client_ErrorHandler(t *testing.T) {
 			HandleErrorFunc: func(ctx context.Context, job *rivertype.JobRow, err error) *ErrorHandlerResult {
 				var unknownJobKindErr *UnknownJobKindError
 				require.ErrorAs(t, err, &unknownJobKindErr)
-				require.Equal(t, *unknownJobKindErr, UnknownJobKindError{Kind: "RandomWorkerNameThatIsNeverRegistered"})
+				require.Equal(t, UnknownJobKindError{Kind: "RandomWorkerNameThatIsNeverRegistered"}, *unknownJobKindErr)
 				errorHandlerCalled = true
 				return &ErrorHandlerResult{}
 			},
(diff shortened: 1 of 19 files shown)


internal/maintenance/job_cleaner.go
```diff
@@ -111,8 +111,8 @@ func (s *JobCleaner) Start(ctx context.Context) error { //nolint:dupl
 		// races.
 		defer close(stopped)
 
-		s.Logger.InfoContext(ctx, s.Name+": Run loop started")
-		defer s.Logger.InfoContext(ctx, s.Name+": Run loop stopped")
+		s.Logger.InfoContext(ctx, s.Name+logPrefixRunLoopStarted)
```
brandur, line 114: I could take or leave this one for sure. Let's try it for a while and see how it feels.

internal/riverinternaltest/riverinternaltest_test.go
```diff
@@ -16,7 +16,7 @@ import (
 // `riverinternaltest_test`, except that the test below uses internal variables
 // like `dbPool`.
 type Executor interface {
-	Exec(context.Context, string, ...interface{}) (pgconn.CommandTag, error)
+	Exec(ctx context.Context, query string, args ...interface{}) (pgconn.CommandTag, error)
```
brandur, line 19: Like this one too. The nice part about having default param names is that when you tell your IDE to autocomplete an implementation scaffold, it can give you nice-looking names in it instead of something generically generated.

internal/util/sliceutil/slice_util_test.go
```diff
@@ -70,8 +70,8 @@ func TestMap(t *testing.T) {
 		return strconv.FormatInt(x, 10)
 	})
 
-	require.Equal(t, len(result1), 4)
-	require.Equal(t, len(result2), 4)
-	require.Equal(t, result1, []string{"Hello", "Hello", "Hello", "Hello"})
-	require.Equal(t, result2, []string{"1", "2", "3", "4"})
+	require.Len(t, result1, 4)
```
brandur, line 73: Another good one IMO.

client_test.go
```diff
@@ -1146,7 +1146,7 @@ func Test_Client_ErrorHandler(t *testing.T) {
 			HandleErrorFunc: func(ctx context.Context, job *rivertype.JobRow, err error) *ErrorHandlerResult {
 				var unknownJobKindErr *UnknownJobKindError
 				require.ErrorAs(t, err, &unknownJobKindErr)
-				require.Equal(t, *unknownJobKindErr, UnknownJobKindError{Kind: "RandomWorkerNameThatIsNeverRegistered"})
```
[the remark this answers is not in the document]
bgentry: Yep, likewise, it's all too easy to swap these. Impressive that this can be linted easily enough.

PR #94, opened 2023-12-03
More fixes for golangci-lint upgrade

brandur:
So after #93 came in I ran my own `brew upgrade golangci-lint`, and I
guess I got a new point release, because I got a whole bunch of new lint
problems. I wish it was clearer which version I even got, but apparently
golangci-lint is releasing HEAD versions to Homebrew:

    $ golangci-lint --version
    golangci-lint has version HEAD-185b2c7 built with go1.21.4 from 185b2c7 on 2023-11-29T17:26:42Z

I fixed most of the problems, but I ended up disabling one of the
features of `testifylint` that disallows uses of testify functions in
goroutines, which would've required some very significant changes, and
which in our project produces a false positive rate (of actual problems)
of 100.0% as far as I can tell. See the description of this feature
here:

https://github.com/Antonboom/testifylint#go-require

    $ golangci-lint run --fix
    client_test.go:378:6: go-require: require must only be used in the goroutine running the test function (testifylint)
                                            require.NoError(t, err)
                                            ^
    internal/baseservice/base_service_test.go:40:4: go-require: require must only be used in the goroutine running the test function (testifylint)
                            require.FailNow(t, "Test case took too long to run (sleep statements should return immediately)")
                            ^
    internal/dbadapter/db_adapter_test.go:409:6: go-require: require must only be used in the goroutine running the test function (testifylint)
                                            require.NoError(t, err)
                                            ^
    internal/riverinternaltest/riverinternaltest_test.go:78:8: go-require: TestTx contains assertions that must only be used in the goroutine running the test function (testifylint)
                            _ = TestTx(ctx, t)

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -226,7 +226,7 @@ func Test_Client(t *testing.T) {
 		}
 
 		AddWorker(client.config.Workers, WorkFunc(func(ctx context.Context, job *Job[JobArgs]) error {
-			return JobCancel(fmt.Errorf("a persisted internal error"))
+			return JobCancel(errors.New("a persisted internal error"))
 		}))
 
 		startClient(ctx, t, client)
@@ -897,7 +897,7 @@ func Test_Client_InsertMany(t *testing.T) {
 
 		jobs, err := bundle.queries.JobGetByKind(ctx, client.driver.GetDBPool(), (noOpArgs{}).Kind())
 		require.NoError(t, err)
-		require.Len(t, jobs, 2, fmt.Sprintf("Expected to find exactly two jobs of kind: %s", (noOpArgs{}).Kind()))
+		require.Len(t, jobs, 2, "Expected to find exactly two jobs of kind: "+(noOpArgs{}).Kind()) //nolint:goconst
 	})
 
 	t.Run("ErrorsOnDriverWithoutPool", func(t *testing.T) {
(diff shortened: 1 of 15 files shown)


client_test.go
```diff
@@ -897,7 +897,7 @@ func Test_Client_InsertMany(t *testing.T) {
 
 		jobs, err := bundle.queries.JobGetByKind(ctx, client.driver.GetDBPool(), (noOpArgs{}).Kind())
 		require.NoError(t, err)
-		require.Len(t, jobs, 2, fmt.Sprintf("Expected to find exactly two jobs of kind: %s", (noOpArgs{}).Kind()))
+		require.Len(t, jobs, 2, "Expected to find exactly two jobs of kind: "+(noOpArgs{}).Kind()) //nolint:goconst
```
brandur, line 900: Added nolint on these lines ... I dunno about adding constants for assertion messages.

brandur:
Ah, actually you're probably right. I likely installed `--head` at some point in the past, and the option was persisted through until now.

```
$ brew info golangci-lint
==> golangci-lint: stable 1.55.2 (bottled), HEAD
Fast linters runner for Go
https://golangci-lint.run/
/opt/homebrew/Cellar/golangci-lint/HEAD-185b2c7 (9 files, 30.5MB) *
  Built from source on 2023-12-02 at 18:19:30
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/g/golangci-lint.rb
License: GPL-3.0-only
==> Dependencies
Required: go ✔
==> Options
--HEAD
        Install HEAD version
==> Caveats
zsh completions have been installed to:
  /opt/homebrew/share/zsh/site-functions
==> Analytics
install: 10,193 (30 days), 31,449 (90 days), 84,287 (365 days)
install-on-request: 10,189 (30 days), 31,438 (90 days), 84,255 (365 days)
build-error: 0 (30 days)
```

brandur:
Thx. I may have jumped the gun, but going to merge this since the work's done and these will be coming down the pipeline soon-ish anyway.

PR #95, opened 2023-12-03
Readme use const QueueDefault

timwmillard:
Incorrect const used in the Readme.

Should be `QueueDefault` not `DefaultQueue`


brandur (approved):
Thanks Tim!

Issue #57, opened 2023-11-21
Make SQL Migrations accessible as a method


mradile:
Thank you for publishing such a great library. The library fits exactly my needs for private project I would to like to open source once its reaches stability.

Currently I am doing database migrations with https://github.com/pressly/goose on the startup of the app. This makes it easy to deploy on hosters like heroku or fly.io. 

Would it be possible to make the functionality of 'river migrate-up' accessible as method? I know this could be a conflict for the leader election if every instance tries to do the migrations on startup but it should be no problem on a single instance.

If you are open to it, I could try to create a pull request for this changes.

bgentry:
I had a feeling this would come up quickly 😅 Is your thought that we could export the migrations as SQL string constants? The issue I'm seeing with that is that [Goose Go migrations](https://github.com/pressly/goose#go-migrations) appear to only work with individual SQL statements, whereas River's [migrations](https://github.com/riverqueue/river/blob/633968c85d03dbbb09448ea086a2fb83de6c6bff/internal/dbmigrate/002_initial_schema.up.sql) are comprised of multiple statements.

Or maybe I'm misreading that?

mradile:
You can use sql files with use goose and separate the up/down migrations.
Example:
SQL file:

```sql
-- +goose Up
-- +goose StatementBegin

STATEMENT_1;
STATEMENT_2;
STATEMENT_N;

-- +goose StatementEnd

-- +goose Down
-- +goose StatementBegin

DOWN_STATEMENT_1;
DOWN_STATEMENT_2;
DOWN_STATEMENT_N;

-- +goose StatementEnd
```

Embed the whole directory with all sql migration files:
```go
//go:embed sql/migrations/*.sql
var EmbedMigrations embed.FS
```

Run the migrations:
```go
var sqlDB *sql.DB
goose.SetBaseFS(EmbedMigrations)
goose.SetDialect("postgres")
goose.Up(db, "sql/migrations")
```

mradile:
> Another option is to expose this as a Go API which takes a pgx pool or driver wrapper and which you can then call something like `migrate.Up(1)` and `migrate.Up(2)` to step through the migrations in sequence. This would allow them to still manipulate multiple tables as needed and cleanly encapsulated several statements.

This is a great idea. Maybe is would also be possible to just run a method like `migrate.All()` do apply all needed migrations.

mfridman:
[...]

PR #98, opened 2023-12-03
Support for `database/sql` in migrations + framework for multi-driver River

brandur:
Here, add a new minimal driver called `riverdriver/riversql` that
supports Go's built-in `database/sql` package, but only for purposes of
migrations. The idea here is to fully complete #57 by providing a way of
making `rivermigrate` interoperable with Go migration frameworks that
support Go-based migrations like Goose, which provides hooks for
`*sql.Tx` [1] rather than pgx.

`riverdriver/riversql` is not a full driver and is only meant to be used
with `rivermigrate`. We document this clearly in a number of places.

To make a multi-driver world possible with River, we have to start the
work of building a platform that does more than `riverpgxv5`'s "cheat"
workaround. This works by having each driver implement specific database
operations like `MigrationGetAll`, which target their wrapped database
package of choice.

This is accomplished by having each driver bundle in its own sqlc that
targets its package. So `riverpgxv5` has an `sqlc.yaml` that targets
`pgx/v5`, while `riversql` has one that targets `database/sql`. There's
some `sqlc.yaml` duplication involved here, but luckily both drivers can
share a `river_migration.sql` file that contains all queries involved,
so you only need to change one place. `river_migration.sql` also migrates
entirely out of the main `./internal/dbsqlc`.

The idea here is that eventually `./internal/dbsqlc` will disappear
completely, usurped entirely by driver-specific versions. As this is
done, all references to `pgx` will disappear from the top-level package.
There are some complications here to figure out like `LISTEN`/`NOTIFY`
though, and I'm not clear whether `database/sql` could ever become a
fully functional driver as it might be missing some needed functionality
(e.g. subtransactions are still not supported after talking about them
for ten f*ing years [2]. However, even if it's not, the system would let
us support other fully functional packages or future major versions of
pgx (or even past ones like `pgx/v4` if there's demand).

`river/riverdriver` becomes a package as it now has types in it that
need to be referenced by driver implementations, and this would
otherwise not be possible without introducing a circular dependency.

~~Notably, this development branch has to use some `go.mod` `replace`
directives to demonstrate that it works correctly. If we go this
direction, we'll need to break it into chunks to release it without
them:~~

1. ~~Break out changes to `river/riverdriver`. Tag and release it.~~

2. ~~Break out changes to `riverdriver/river*` drivers. Have them target
   the release in (1), comment out `replace`s, then tag and release them.~~

3. ~~Target the remaining River changes to the releases in (1) and (2),
   comment out `replace`s, then tag and release the top-level driver.~~

~~Unfortunately future deep incisions to drivers will require similar
gymnastics, but I don't think there's a way around it (we already have
this process except it's currently two steps instead of three). The hope
is that these will change relatively rarely, so it won't be too painful.~~

[1] https://github.com/pressly/goose#go-migrations
[2] https://github.com/golang/go/issues/7898


riverdriver/river_driver_interface.go
```diff
@@ -32,10 +42,95 @@ import (
...
 	// internal infrastructure. This doesn't make sense in a world where
 	// multiple drivers are supported and is subject to change.
+	//
+	// API is not stable. DO NOT USE.
 	UnwrapTx(tx TTx) pgx.Tx
 }
+
+// Executor provides River operations against a database. It may be a database
+// pool or transaction.
+type Executor interface {
```
brandur, line 69: The executor will eventually contain all possible operations against a River schema. Notably, it looks quite similar to an adapter, and long term, will probably replace it.

go.mod
```diff
@@ -1,15 +1,20 @@
...
 
-go 1.21.0
+go 1.21.4
 
-// replace github.com/riverqueue/river/riverdriver/riverpgxv5 => ./riverdriver/riverpgxv5
+replace github.com/riverqueue/river/riverdriver => ./riverdriver
+
+replace github.com/riverqueue/river/riverdriver/riverpgxv5 => ./riverdriver/riverpgxv5
+
+replace github.com/riverqueue/river/riverdriver/riversql => ./riverdriver/riversql
```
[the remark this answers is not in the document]
brandur: Also note: lint's going to fail for this PR because of these, but the other checks are all working, so safe to ignore for now.

brandur:
@bgentry Arg, sorry this one's kind of big, but I figured it was better to build a bit more of a solid platform for the drivers rather than introduce some new backdoor to make them usable with `database/sql`. The tl;dr is this will give us a way to support Goose through `rivermigrate` and _hopefully_ provide a driver platform that'll allow us to move operations into one by one until we converge on full non-hack support over time.

riverdriver/riversql/go.mod
```diff
@@ -0,0 +1,25 @@
+module github.com/riverqueue/river/riverdriver/riversql
```
bgentry, line 1: I have some qualms with this package name. Primarily worried that it could be confusing or take away a name that we might want to use for other purposes. However the best alternative I can come up with is `riverdbsql` which I don't love either 😕

riverdriver/river_driver_interface.go
```diff
@@ -32,10 +42,95 @@ import (
...
+
+	// TableExists checks whether a table exists for the schema in the current
+	// search schema.
+	//
+	// API is not stable. DO NOT USE.
+	TableExists(ctx context.Context, tableName string) (bool, error)
+}
+
+// ExecutorTx is an executor which is a transaction. In addition to standard
+// Executor operations, it may be commited or rolled back.
```
bgentry, line 105:
```suggestion
// Executor operations, it may be committed or rolled back.
```

riverdriver/riversql/internal/dbsqlc/sqlc.yaml
```diff
@@ -0,0 +1,24 @@
+version: "2"
+sql:
+  - engine: "postgresql"
+    queries:
+      - ../../../riverpgxv5/internal/dbsqlc/river_migration.sql
+    schema:
+      - ../../../riverpgxv5/internal/dbsqlc/river_migration.sql
```
bgentry, line 7: Is the thought that in general we should be able to share queries between drivers and let sqlc abstract away the differences? More generated code to execute or compile, but less code to actually maintain?

After reading your comment about how each driver would need to implement the full set of query methods I had some concerns about the amount of duplication we'd end up with, but if we can share the underlying schema and query definitions then that should help considerably.

riverdriver/riverpgxv5/river_pgx_v5_driver.go
```diff
@@ -29,8 +34,96 @@ type Driver struct {
...
+// mapSlice manipulates a slice and transforms it to a slice of another type.
+func mapSlice[T any, R any](collection []T, mapFunc func(T) R) []R {
+	result := make([]R, len(collection))
+
+	for i, item := range collection {
+		result[i] = mapFunc(item)
+	}
+
+	return result
+}
```
bgentry, line 129: I assume this is included in the pgxv5 driver to avoid a circular dependency on `internal/util/sliceutil`?

bgentry (approved):
Fantastic work on this, it's a nice step toward broader driver support 🎉 🚀 

> There are some complications here to figure out like `LISTEN`/`NOTIFY`
> though, and I'm not clear whether `database/sql` could ever become a
> fully functional driver as it might be missing some needed functionality
> (e.g. subtransactions are still not supported after talking about them
> for ten f*ing years [2]. However, even if it's not, the system would let
> us support other fully functional packages or future major versions of
> pgx (or even past ones like `pgx/v4` if there's demand).

Yeah, the listener implementation is deeply tied to pgxv5 and might either be very difficult or impossible to implement in other drivers. My current thinking is that all the drivers will need to have a dependency on pgxv5 and provide a way to retrieve a `*pgx.ConnConfig` that the `notifier` can use to connect to the DB with pgxv5. That or we pull the notifier into the driver interface, and they'll all still be implemented with pgxv5 under the covers even if another driver is also being used for other queries.

riverdriver/river_driver_interface.go
brandur: Thx. Fixed.

riverdriver/riverpgxv5/river_pgx_v5_driver.go
brandur: Yeah :/ We could add another package that both top-level `river` and the drivers could both use, but doesn't feel worth it. We could put it in `riverdriver` to share, but that feels pretty weird. I suspect this is the main function that'll be useful so there won't be too much more duplication.

riverdriver/riversql/go.mod
brandur: @bgentry 

I'm -1 on `riverdbsql` just because it doesn't really reflect the actual naming of `database/sql` very well (i.e. abbreviation in one but not the other).

I just pushed a change that renames it to `riverdatabasesql` — it's a little bit of a mouthful (one reason I originally avoided it), but I figure that this package will be used so rarely that it probably doesn't matter. What do you think?

brandur:
@bgentry Alright, it seems like #102 is also bad, so back to the original plan. What I'd do for release is:

1. Merge this. Tag `riverdriver` submodule and release it.
2. Remove `replace` in `riverdatabasesql` and `riverpgxv5` and point them to release from (1) instead. Commit, tag and release both of them.
3. Remove `replace`s in top-level River package, pointing them to (2) and (3) instead. Commit, tag and release it.

I also had to bring this back in `.golanci.yml` so as not to break the master build as this comes in:

``` yaml
  gomoddirectives:
    replace-local: true
```

We'd remove that again as part of step (3).

This seems pretty painful for sure, but I don't think there's another way? We could automate most of this, but we'd have to unprotect the master branch because it'd require creating multiple commits in succession.

Does this sound right you?

PR #101, opened 2023-12-06
Consolidate on one set state query + adapter/completer function [boilerplate reduction]

brandur:
This one's a clean up aimed at doing a massive boilerplate reduction in
our sqlc queries, the adapter, and the completer(s). We have a few
current problems:

* Every "set state" query like `JobCompleteIfRunning`, `JobErrorIfRunning`,
  etc. are all separate sqlc queries, and worse yet, all very long ones
  that are all practically identical with very minor but important
  differences that are hard to spot without examining them character by
  character.

* Every one of these queries get exposed through the adapter as a
  separate function that's implemented in the interface, `StandardAdapter`,
  and the mock `TestAdapter`, requiring lots of LOC.

* They're all exposed yet again in the `Completer` interface, along with
  both implementations of `Completer`. Which again, is a lot of LOC.

Here, we show that with some fairly minor tweaks, we can consolidate all
of this state updating into a single query that's no longer than any one
of the originals, and then simplify both the adapter and completer to
model access to it as a general `JobSetState` function with a number of
params constructors that function to produce the necessary parameters
for each type of state of update (complete, error, snooze, etc.).

Total LOCs are reduced by *a lot*. `river_job.sql` just by itself has
40% fewer lines, and with many more removed in the adapters and
completers. This has the effect of a lot less noise, but also makes
refactoring easier because there's less that has to change. Furthermore,
I don't believe there's any significant reduction in API ergonomics or
safety.

It's still not perfect -- there's still way too much scaffolding in the
completers in particular for example, but more improvements to come.

### DIFF
--- a/internal/dbadapter/db_adapter.go
+++ b/internal/dbadapter/db_adapter.go
@@ -81,9 +81,6 @@ type JobInsertResult struct {
 // expedience, but this should be converted to a more stable API if Adapter
 // would be exported.
 type Adapter interface {
-	JobCompleteMany(ctx context.Context, jobs ...JobToComplete) error
-	// TODO: should all dbsqlc need to implement this? Or is it a pro feature?
-	JobCompleteManyTx(ctx context.Context, tx pgx.Tx, jobs ...JobToComplete) error
 	JobInsert(ctx context.Context, params *JobInsertParams) (*JobInsertResult, error)
 	JobInsertTx(ctx context.Context, tx pgx.Tx, params *JobInsertParams) (*JobInsertResult, error)
 
@@ -94,12 +91,10 @@ type Adapter interface {
 	JobGetAvailable(ctx context.Context, queueName string, limit int32) ([]*dbsqlc.RiverJob, error)
 	JobGetAvailableTx(ctx context.Context, tx pgx.Tx, queueName string, limit int32) ([]*dbsqlc.RiverJob, error)
 
-	JobSetCancelledIfRunning(ctx context.Context, id int64, cancelledAt time.Time, err []byte) (*dbsqlc.RiverJob, error)
-	JobSetCompletedIfRunning(ctx context.Context, job JobToComplete) (*dbsqlc.RiverJob, error)
-	JobSetCompletedTx(ctx context.Context, tx pgx.Tx, id int64, completedAt time.Time) (*dbsqlc.RiverJob, error)
-	JobSetDiscardedIfRunning(ctx context.Context, id int64, discardedAt time.Time, err []byte) (*dbsqlc.RiverJob, error)
-	JobSetErroredIfRunning(ctx context.Context, id int64, scheduledAt time.Time, err []byte) (*dbsqlc.RiverJob, error)
-	JobSetSnoozedIfRunning(ctx context.Context, id int64, scheduledAt time.Time) (*dbsqlc.RiverJob, error)
+	// JobSetStateIfRunning sets the state of a currently running job. Jobs which are not
+	// running (i.e. which have already have had their state set to something
+	// new through an explicit snooze or cancellation), are ignored.
+	JobSetStateIfRunning(ctx context.Context, params *JobSetStateIfRunningParams) (*dbsqlc.RiverJob, error)
 
 	// LeadershipAttemptElect attempts to elect a leader for the given name. The
 	// bool alreadyElected indicates whether this is a potential reelection of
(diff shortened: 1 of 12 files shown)


internal/dbadapter/db_adapter.go
```diff
@@ -377,66 +352,53 @@ func (a *StandardAdapter) JobGetAvailableTx(ctx context.Context, tx pgx.Tx, queu
...
-func (a *StandardAdapter) JobSetCompletedIfRunning(ctx context.Context, job JobToComplete) (*dbsqlc.RiverJob, error) {
-	ctx, cancel := context.WithTimeout(ctx, a.deadlineTimeout)
-	defer cancel()
-
-	return a.queries.JobSetCompletedIfRunning(ctx, a.executor, dbsqlc.JobSetCompletedIfRunningParams{
-		ID:          job.ID,
-		FinalizedAt: job.FinalizedAt.UTC(),
-	})
+func JobSetStateCancelled(id int64, finalizedAt time.Time, errData []byte) *JobSetStateParams {
+	return &JobSetStateParams{ID: id, errData: errData, finalizedAt: &finalizedAt, state: dbsqlc.JobStateCancelled}
```
brandur, line 368: See these new params constructors, but the general idea is that they help ensure that only valid parameter combinations are generated for the update query.

internal/util/ptrutil/ptr_util.go
```diff
@@ -4,3 +4,23 @@ package ptrutil
 func Ptr[T any](v T) *T {
 	return &v
 }
+
+// ValOrDefault returns a pointer's value if it's non-nil, and the given default
+// value otherwise.
+func ValOrDefault[T any](val *T, defaultVal T) T {
```
brandur, line 10: Similar to `valutil.ValOrDefault`. I find these pretty useful and needed them for an adapter change, so brought 'em in.

job_executor.go
```diff
@@ -245,7 +245,7 @@ func (e *jobExecutor) reportResult(ctx context.Context, res *jobExecutorResult)
 		return
 	}
 
-	if err := e.Completer.JobSetCompleted(e.JobRow.ID, e.stats, e.TimeNowUTC()); err != nil {
+	if err := e.Completer.JobSetState(e.stats, dbadapter.JobSetStateCompleted(e.JobRow.ID, e.TimeNowUTC())); err != nil {
```
brandur, line 248: The invocations from the executor get a tiny bit more verbose, but not my that much.

internal/dbsqlc/river_job.sql
```diff
@@ -238,226 +225,42 @@ FROM (
...
-  SELECT
-    updated_job.*
-  FROM
-    updated_job
-) ORDER BY scheduled_at DESC NULLS LAST LIMIT 1;
-
+SELECT *
+FROM river_job
+WHERE id = @id::bigint
+    AND id NOT IN (SELECT id FROM updated_job)
```
brandur, line 260: (Repost since I tweaked this line and the comment was hidden.)

This is a key change here — by excluding an updated job from the lookup on `river_job`, we can drop the `ORDER BY` + `LIMIT` clauses, which means that when combining all the queries we don't need any `CASE WHEN` shenanigans inside the `ORDER BY` clause.

brandur:
@bgentry This one's a non-user facing refactor, but should help us get to a drivers system more easily if we go that way.

internal/dbadapter/db_adapter.go
```diff
@@ -377,66 +352,53 @@ func (a *StandardAdapter) JobGetAvailableTx(ctx context.Context, tx pgx.Tx, queu
...
-		ID:          id,
-		ScheduledAt: scheduledAt,
+	return a.queries.JobSetStateIfRunning(ctx, a.executor, dbsqlc.JobSetStateIfRunningParams{
+		ID:                  params.ID,
+		ErrorDoUpdate:       params.errData != nil,
+		Error:               params.errData,
+		FinalizedAtDoUpdate: params.finalizedAt != nil,
+		FinalizedAt:         params.finalizedAt,
+		MaxAttemptsUpdate:   params.maxAttempts != nil,
+		MaxAttempts:         int16(min(ptrutil.ValOrDefault(params.maxAttempts, 0), math.MaxInt16)), // default never used
```
bgentry, line 398: This line is confusing me a bit. What is the `min` achieving?

bgentry (approved):
Looks fantastic, thank you for refactoring away all that duplication I produced in the run up to release :)

I have one minor question on naming within the adapter but otherwise this LGTM :shipit:

internal/dbadapter/db_adapter.go
```diff
@@ -377,66 +352,53 @@ func (a *StandardAdapter) JobGetAvailableTx(ctx context.Context, tx pgx.Tx, queu
...
-		ID:          id,
-		Error:       err,
-		ScheduledAt: scheduledAt,
-	})
+func JobSetStateSnoozed(id int64, scheduledAt time.Time, maxAttempts int) *JobSetStateParams {
+	return &JobSetStateParams{ID: id, maxAttempts: &maxAttempts, scheduledAt: &scheduledAt, state: dbsqlc.JobStateScheduled}
 }
 
-func (a *StandardAdapter) JobSetSnoozedIfRunning(ctx context.Context, id int64, scheduledAt time.Time) (*dbsqlc.RiverJob, error) {
+func (a *StandardAdapter) JobSetState(ctx context.Context, params *JobSetStateParams) (*dbsqlc.RiverJob, error) {
```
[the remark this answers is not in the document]
brandur: Hah, I actually had it as `*IfRunning` almost the entire time I was developing this branch, but removed it at the last second because it felt like it was adding a lot of verbosity everywhere. I don't feel strongly about it though, so I added it back in. The completer function also becomes `*IfRunning` so we're consistent everywhere.

internal/dbadapter/db_adapter.go
brandur: I actually put this in because I thought you'd want it :) Basically, it's protecting against a possible overflow as `int` becomes `int16`.

That said, this will never actually be a problem in practice, so I don't care if it's there or not. Took it out.

Issue #65, opened 2023-11-23
flaky StopAndCancel test in CI


bgentry:
GitHub Actions continues to surprise us with [timing issues](https://github.com/riverqueue/river/actions/runs/6973035679/job/18976333126) @brandur:

```
client_test.go:546: Waiting on job to be done
        client_test.go:559: 
            	Error Trace:	/home/runner/work/river/river/client_test.go:559
            	Error:      	Max difference between 2023-11-23 18:02:54.632264664 +0000 UTC m=+0.706377072 and 2023-11-23 18:02:54.531149022 +0000 UTC m=+0.6052614[30](https://github.com/riverqueue/river/actions/runs/6973035679/job/18976333126#step:8:31) allowed is 100ms, but difference was 101.115642ms
            	Test:       	Test_Client_StopAndCancel/jobs_in_progress,_only_completing_when_context_is_canceled
```

brandur:
1 ms over budget. Seems like the tolerance on that might be a little too tight.

PR #102, opened 2023-12-09
Put River CLI into its own submodule

brandur:
Background: While trying to figure out how to safely release [1]
(which includes a new `riverdriver` submodule), I threw myself through a
loop and realized that it'd be practically impossible to do without a
whole bunch of manual finnagling on the release tags/lint settings in
the repo, or breaking the master build, or most likely, both.

This got me thinking: maybe we should, maybe we _should_ be using
`replace` directives to make this less painful/brittle, which were
originally removed in #35 because using `replace` makes it impossible to
use `go install ...@latest`.

A change that we made recently was the addition of a new Go migration
API in #65 (`rivermigrate`), and a side effect of that change is that
the API being used by the River CLI became entirely public (previously,
it'd depended on packages in River's `internal`). By extension, that
means it's now possible to make the River CLI its own Go module,
something that was infeasible before because of the use of `internal`.

So my thinking currently: maybe we should go back to trying to see if we
can make `replace` in use by most of River's core packages, and keep the
River CLI as its own module without `replace` so that it can still be
installed with `go install ...@latest`. I'm not entirely sure we won't
run into another problem, but it might be the easiest thing in the
meantime. As the River CLI expands, we'll need to make sure to only use
public APIs, but that might not be the worst thing anyway -- we could
adopt the philosophy that any function the CLI accesses should also be
accessible by the Go API.

Here, notably we still use a `replace`, which I'm finding that I need to
have a passing build for now, and which I think will have to temporarily
stay until we cut a new release. Trying to build the new submodule
without results in this error that I was unable to otherwise find a way
around:

    $ go test .
    ambiguous import: found package github.com/riverqueue/river/cmd/river in multiple modules:
            github.com/riverqueue/river v0.0.12 (/Users/brandur/Documents/go/pkg/mod/github.com/riverqueue/river@v0.0.12/cmd/river)

[1] https://github.com/riverqueue/river/pull/98


cmd/river/go.mod
```diff
@@ -0,0 +1,26 @@
+module github.com/riverqueue/river/cmd/river
+
+go 1.21.4
+
+replace github.com/riverqueue/river => ../..
+
+replace github.com/riverqueue/river/riverdriver/riverpgxv5 => ../../riverdriver/riverpgxv5
```
brandur, line 7: So the whole point of this is that `cmd/river` will not use a `replace` in the end, but I found that I had to have one for now because you can't build this module without one given the current release contains `cmd/river` as a subpackage:

```
$ go test .
ambiguous import: found package github.com/riverqueue/river/cmd/river in multiple modules:
        github.com/riverqueue/river v0.0.12 (/Users/brandur/Documents/go/pkg/mod/github.com/riverqueue/river@v0.0.12/cmd/river)
```

So I believe what we could do is (1) merge this, (2) cut a new release, and (3) update this package to use that new release and remove this `replace`, then (4) uncomment the other `replace`s in the project.

brandur:
Damn it. This does work, but makes the install experience more painful for new projects. See: https://github.com/brandur/go-module-koan-1.

I checked at how Bun does it (as an example), and it's a little better there because their main module doesn't reference any of the submodules, so you don't get the attempted resolution of the invalid v000000 version, which is a little better, but you still have to add each submodule separately. It seems they're able to get away without referencing submodules because there's no actual common driver interface except the built-in one from `database/sql/driver`.

bgentry:
@brandur ok, so [this](https://github.com/bgentry/bgmodtest) works as of the last commit: 

```
➜  bgtest2 git:(master)   go install github.com/bgentry/bgmodtest/cmd/go-module-koan-1@latest
go: downloading github.com/bgentry/bgmodtest/bgmoddriver/bgmodpgxv5 v0.0.6
go: downloading github.com/bgentry/bgmodtest/bgmoddriver v0.0.6
```

In this setup you can even install the top-level module directly and have it pull in the required submodules:

```
➜  bgtest2 git:(master)  go get github.com/bgentry/bgmodtest@v0.0.6
go: added github.com/bgentry/bgmodtest v0.0.6
go: added github.com/bgentry/bgmodtest/bgmoddriver v0.0.6
go: added github.com/bgentry/bgmodtest/bgmoddriver/bgmodpgxv5 v0.0.6
```

I think what made it work was me removing the `replace` from the cmd’s `go.mod`. The previous `v0.0.5` release just before I did that does _not_ work because of the usual `replace` error. I don’t really get why this works for `gopls`, though my suspicion is that it’s because the main `tools` module there [does _not_](https://github.com/golang/tools/blob/master/go.mod) have any `replace` in it, just [the `gopls` `go.mod`](https://github.com/golang/tools/blob/master/gopls/go.mod) does.

Notably, when releasing new versions of this I did a find/replace across all the `go.mod` files to bump the versions from i.e. `v0.0.5` to `v0.0.6`, committed that and tagged the result as `v0.0.6` before pushing the tags up (including a tag for the cmd module). Does it seem like this would work for us? I think we could bump everything in a single PR and merge them before then tagging that new version. Unless of course that will break CI, which brings us back to finding a way to do this with GitHub Actions (hopefully without a bunch of extra commits though).

brandur:
@bgentry Awesome, thanks a lot for looking into that — as discussed offline, glad you spotted what I didn't, because we were headed down a much worst path I think. As discussed, going to bring this in along with #98 and see if we can successfully cut a release.

I'm going to YOLO a bit with a manual process this time, but if it works, I'll write up some instructions (+ a script possibly), to help with this in the future.

brandur:
Arg, accidentally added an extra `d` to `-f` when pushing my branch, which deleted it and closed this automatically with no obvious way to reopen it. Opened #107 to replace it.

PR #105, opened 2023-12-10
Eager retry for short retry delays

brandur:
An aspect of River that could currently be considered a bug is that
retry delays below five seconds are effectively ignored. This is because
when a job errors it always transitions to `retryable` state, and needs
to wait for the scheduler to run to transition it back to `available`,
and the scheduler run interval is five seconds.

This is in practice a problem because it means that the first retry in a
retry policy isn't actually one second as claimed in docs, etc., but
actually ~5 seconds, and it won't be obvious why.

Here, implement an "eager" retry system such that if we detect that the
retry delay is smaller than the scheduler's run interval, we place the
job immediately into an `available` state, allowing it to be worked
sooner.

To facilitate this we add a predicate on the "get available" query that
checks for `scheduled_at <= now()` so that errored jobs with short
retries can be `available` for a short time without being worked
immediately. Most of the time this should have no effect because this
really only applies to the first retry in any failure (the second is at
16 seconds, which is well above the scheduler's run interval).

### DIFF
--- a/internal/dbsqlc/river_job.sql
+++ b/internal/dbsqlc/river_job.sql
@@ -72,6 +72,7 @@ WITH locked_jobs AS (
   WHERE
     state = 'available'::river_job_state
     AND queue = @queue::text
+    AND scheduled_at <= now()
   ORDER BY
     priority ASC,
     scheduled_at ASC,
@@ -84,7 +85,7 @@ UPDATE
 SET
   state = 'running'::river_job_state,
   attempt = river_job.attempt + 1,
-  attempted_at = NOW(),
+  attempted_at = now(),
   attempted_by = array_append(river_job.attempted_by, @worker::text)
 FROM
   locked_jobs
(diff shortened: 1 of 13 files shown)


internal/dbsqlc/river_job.sql
```diff
@@ -72,6 +72,7 @@ WITH locked_jobs AS (
   WHERE
     state = 'available'::river_job_state
     AND queue = @queue::text
+    AND scheduled_at <= now()
```
bgentry, line 75: This should be totally fine from a query performance perspective because it's part of the compound fetching index and is already being used for sorting.

bgentry (approved):
Excellent improvement :shipit:

internal/dbsqlc/river_job.sql
brandur: Great. Yeah I figured that the extra predicate is on average going to have very little work to do since the vast majority of rows that aren't ready yet will still be `scheduled`.

brandur:
Excellent. Thanks! Added a changelog and pulling this in.

PR #108, opened 2023-12-13
Preparations for 0.0.13 release

brandur:
Here, tee up the necessary changes for an upcoming 0.0.13 release that
I'll cut immediately after this is merged.

We update each submodule's `go.mod` to point to 0.0.13, which doesn't
actually exist yet, but it will very soon. This will leave the repo very
temporarily in an invalid state, but I tried the same change out in a
test project and as soon as the Go module proxy's cache busts, it works.

We comment out all the `replace`'s in `./cmd/river` so that it continues
to be installable with `go install ...@latest`. The `replace`'s are
still in there but commented out so that we can test future CLI changes
without too much trouble.

Finally, update the changelog to add the new section for 0.0.13.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.0.13] - 2023-12-12
+
 ### Added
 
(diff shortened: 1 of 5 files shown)


brandur:
@bgentry Sooo, I don't think there's any way to get this build to pass, but the idea is that immediately after I merge it, I'll cut a new release, and then the build should be okay after that. Unfortunately I'm not sure there's a good way around this as things stand currently.

bgentry (approved):

brandur:
TY.

brandur:
Alright, `v0.0.0.13` is cut. I'll wait a bit and then check to see that everything still works.

PR #109, opened 2023-12-13
Fix development instructions for releasing `riverdriver`

brandur:
Have no idea how my copy/paste skills failed so badly here, but I
somehow managed to add release instructions for `riverdriver` (and only
`riverdriver`) that release new versions at `VERSION` (a non-variable)
instead of `$VERSION`. This caused attempted updates requiring the
driver to fail.

I already removed the bad tag and pushed a new one, but here, update the
instructions to fix future releases.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -33,7 +33,7 @@ queries. After changing an sqlc `.sql` file, generate Go with:
 git checkout master && git pull --rebase
 VERSION=v0.0.x
 git tag cmd/river/$VERSION -m "release cmd/river/$VERSION"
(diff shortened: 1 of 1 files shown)


brandur:
@bgentry Ugh, build still failing due to a mistake I made with releasing the CLI, which I'll send another fix for soon (as soon as the freaking cache is busted). This is a doc-only change though and should be safe.

bgentry (approved):
Ugh sorry I missed this

brandur:
No worries. Can't believe I miscopied/pasted this, heh.

PR #110, opened 2023-12-13
Fix module references for River CLI

brandur:
This one follows up #108 with a fix wherein we add concrete `go.sum`
entries for references to other River modules. This is necessary because
the CLI module no longer has `replace` directives.

This one was my mistake. As in #108, _most_ of the modules in River will
be releasable simultaneously, but the River CLI is special because it
doesn't have `replace` directives so it can stay `go install`able.

In the future, we'll probably set this up such that most of the modules
can be released together, but if the River CLI needs a new River
dependency version, we'll have to make sure to release the top-level
River packages first, then follow up the CLI with its own update with
the version bump.


brandur:
@bgentry Damn, should've seen this one coming, but because `./cmd/river` has no `replace`, we can update its target River version, but the `go.sum` file is still going to need concrete references.

We'll have to be careful when updating the CLI in the future, but this'll fix the problem for now. As you can see, the build is finally succeeding again, which is a good sign. I'll cut a new release when merging this to make sure the CLI has a valid release target.

bgentry (approved):
Yep, this is why in my testing I ended up releasing the cmd with an additional commit and tag after everything else had been released and tagged. Annoying but I think we can find a way to automate it.

brandur:
> Annoying but I think we can find a way to automate it.

Yep, +1. Thanks!

brandur:
Alright, v0.0.14 released. Our lives should be a little easier for now as long as we don't have to make too many CLI releases.

PR #112, opened 2023-12-13
Add Makefile list for tests and CLI + lint fix

brandur:
A side effect of Go submodules in a project is that commands like `go
test ./...` and golangci-lint don't run for the submodules without
`cd`ing into their directory and running those commands explicitly.

Here, add two new Makefile targets for `lint` and `test` that
exhaustively lint and test all submodules in the project. Most of the
time these probably won't be needed as we'll be working inside of the
main River project, but they'll be available in case we're making driver
changes and need something more comprehensive.

Also, fix a lint issue that's been failing on and off for a few days
now.

Lastly, I'm going to propose we remove the `only-new-issues` config
option for golangci-lint in CI. It seems like a useful idea, but it
creates some odd behavior like for the lint problem fixed here in which
lint is in a failing-but-sometimes-not-failing state where it allows
lint to pass in pull requests, but it fails locally and fails the build
when it merges to master. IMO it's better to find out about the problems
sooner rather than later even if it creates a new failure in a pull
request every once in a while.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,35 +151,30 @@ jobs:
       - name: Lint
         uses: golangci/golangci-lint-action@v3
         with:
-          only-new-issues: true # Optional: show only new issues if it's a pull request. The default value is `false`.
           version: ${{ env.GOLANGCI_LINT_VERSION }}
           working-directory: .
 
       - name: Lint cmd/river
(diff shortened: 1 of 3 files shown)


bgentry (approved):

brandur:
ty!

PR #114, opened 2023-12-14
Add script that updates River versions in `go.mod` files for release

brandur:
Here, add a helper script whose job it is to during a release, go into
each `go.mod` submodule in the River project and update the version of
each River dependency to the version found in env `$VERSION`.

This job is made somewhat easier because Go has a package to easily
parse and write `go.mod` files, so changing values within one is a low
risk operation.

A successful run looks like:

    $ export VERSION=v0.0.14
    $ go run ./internal/cmd/update-submodule-versions/main.go
    go.mod
        set version to v0.0.14 for github.com/riverqueue/river/riverdriver
        set version to v0.0.14 for github.com/riverqueue/river/riverdriver/riverdatabasesql
        set version to v0.0.14 for github.com/riverqueue/river/riverdriver/riverpgxv5
    riverdriver/go.mod
        no changes
    riverdriver/riverdatabasesql/go.mod
        set version to v0.0.14 for github.com/riverqueue/river/riverdriver
    riverdriver/riverpgxv5/go.mod
        set version to v0.0.14 for github.com/riverqueue/river/riverdriver

I figure that more automation is probably possible in the future, but
this is an easy way to give us a fast/reliable way today of changing
versions for a release without accidentally forgetting about one of the
`go.mod`s (which, without the script, is quite easy to do).


docs/development.md
```diff
@@ -27,15 +27,27 @@ queries. After changing an sqlc `.sql` file, generate Go with:
...
+### Releasing River CLI
+
+The CLI (`./cmd/river`) is different than other River submodules in that it doesn't use any `replace` directives so that it can stay installable with `go install ...@latest`.
+
+If changes to it don't require updates to its other River dependencies (i.e. they're internal to the CLI only), it can be released normally as shown above.
+
+If updates to River dependencies _are_ required, then a two-phase update is necessary:
+
+1. Release River dependencies with an initial version (e.g. `v0.0.14`).
+2. From `./cmd/river`, `go get` to upgrade to the version from (1), run `go mod tidy`, then release another version (e.g. `v0.0.15`).
```
bgentry, line 53: I think we can tweak this to avoid the double versioning as I did in my testing. All you have to do is tag & push everything else first as above but skip the `cmd/river` tag.

Once that's done, you can follow step 2 here to bump the versions in `cmd/river/go.sum` to the newly released version, commit & push that, then tag it as `cmd/river/v0.0.14`.

The released versions won't have the same commit, but the numbers will at least align this way.

bgentry (approved):
One suggestion but otherwise lgtm :v:

docs/development.md
brandur: Cool. It feels a little weird to have the tags point to different commits, but I guess it doesn't matter. Updated the docs.

bgentry: Indeed. Both of these options are weird, thanks Go 😆

PR #116, opened 2023-12-16
remove redundant internal/rivertest pkg

bgentry:
This was duplicated in a `riverinternaltest` package after we created a public `rivertest` package, but just a few references to it were left around and not updated. This fixes those references and nukes the redundant package.

PR #117, opened 2023-12-16
implement job listing

bgentry:
This implements an API for listing jobs, including filters and cursor-based pagination. I've intentionally kept the options quite minimal to start, though the API design should expand nicely to support more sort and filter options. We will probably need a better way to generate query fragments for the cursor-based pagination if we bring in more options for that.

I recently upgraded sqlc locally and was getting diff here; thought it was best to just get us upgraded so I added a commit for that as well.

Currently based on #116 which I figured should be its own isolated PR.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1108,7 +1108,7 @@ func insertParamsFromArgsAndOptions(args JobArgs, insertOpts *InsertOpts) (*dbad
 	return insertParams, nil
 }
 
-var errInsertNoDriverDBPool = errors.New("driver must have non-nil database pool to use Insert and InsertMany (try InsertTx or InsertManyTx instead")
+var errNoDriverDBPool = errors.New("driver must have non-nil database pool to use non-transactional methods like Insert and InsertMany (try InsertTx or InsertManyTx instead")
 
 // Insert inserts a new job with the provided args. Job opts can be used to
 // override any defaults that may have been provided by an implementation of
@@ -1122,7 +1122,7 @@ var errInsertNoDriverDBPool = errors.New("driver must have non-nil database pool
 //	}
 func (c *Client[TTx]) Insert(ctx context.Context, args JobArgs, opts *InsertOpts) (*rivertype.JobRow, error) {
 	if c.driver.GetDBPool() == nil {
-		return nil, errInsertNoDriverDBPool
+		return nil, errNoDriverDBPool
 	}
 
 	if err := c.validateJobArgs(args); err != nil {
@@ -1201,7 +1201,7 @@ type InsertManyParams struct {
 //	}
 func (c *Client[TTx]) InsertMany(ctx context.Context, params []InsertManyParams) (int64, error) {
 	if c.driver.GetDBPool() == nil {
-		return 0, errInsertNoDriverDBPool
+		return 0, errNoDriverDBPool
 	}
 
 	insertParams, err := c.insertManyParams(params)
@@ -1316,10 +1316,8 @@ func (c *Client[TTx]) JobList(ctx context.Context, params *JobListParams) ([]*ri
 	if params == nil {
 		params = NewJobListParams()
 	}
-	// TODO(bgentry): confirm with Brandur, do we want to error in this scenario?
-	// If so this error is not worded appropriately for these methods.
 	if c.driver.GetDBPool() == nil {
-		return nil, errInsertNoDriverDBPool
+		return nil, errNoDriverDBPool
 	}
 
 	dbParams, err := params.toDBParams()
@@ -1344,12 +1342,6 @@ func (c *Client[TTx]) JobList(ctx context.Context, params *JobListParams) ([]*ri
 //		// handle error
 //	}
 func (c *Client[TTx]) JobListTx(ctx context.Context, tx TTx, params JobListParams) ([]*rivertype.JobRow, error) {
-	// TODO(bgentry): confirm with Brandur, do we want to error in this scenario?
-	// If so this error is not worded appropriately for these methods.
-	if c.driver.GetDBPool() == nil {
-		return nil, errInsertNoDriverDBPool
-	}
-
 	dbParams, err := params.toDBParams()
 	if err != nil {
 		return nil, err
--- a/client.go
+++ b/client.go
@@ -1307,19 +1307,19 @@ func validateQueueName(queueName string) error {
 // provided context is used for the underlying Postgres query and can be used to
 // cancel the operation or apply a timeout.
 //
-//	params := river.JobListParams{}.WithLimit(10).State(river.JobStateCompleted)
+//	params := river.NewJobListParams().WithLimit(10).State(river.JobStateCompleted)
 //	jobRows, err := client.JobList(ctx, params)
 //	if err != nil {
 //		// handle error
 //	}
 func (c *Client[TTx]) JobList(ctx context.Context, params *JobListParams) ([]*rivertype.JobRow, error) {
-	if params == nil {
-		params = NewJobListParams()
-	}
 	if c.driver.GetDBPool() == nil {
 		return nil, errNoDriverDBPool
 	}
 
+	if params == nil {
+		params = NewJobListParams()
+	}
 	dbParams, err := params.toDBParams()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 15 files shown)


client.go
```diff
@@ -1188,3 +1188,76 @@ func validateQueueName(queueName string) error {
...
+//	}
+func (c *Client[TTx]) JobList(ctx context.Context, params *JobListParams) ([]*rivertype.JobRow, error) {
+	if params == nil {
+		params = NewJobListParams()
+	}
+	// TODO(bgentry): confirm with Brandur, do we want to error in this scenario?
+	// If so this error is not worded appropriately for these methods.
+	if c.driver.GetDBPool() == nil {
+		return nil, errInsertNoDriverDBPool
+	}
```
bgentry, line 1209: @brandur left this here to ask you. We will need to error for pseudo-drivers like the current `database/sql` one, right? Although this error seems specifically worded for insert. Should we make it more generalized?

internal/dbadapter/db_adapter.go
```diff
@@ -71,6 +72,29 @@ type JobInsertResult struct {
...
+	SortOrderDesc
+)
+
+type JobListOrderBy struct {
+	Expr  string
+	Order SortOrder
+}
+
+type JobListParams struct {
+	Conditions string
```
brandur, line 89: Just checking here: conditions are meant to be strictly stay an internal construct right? Just wondering if we have to do anything about making sure that SQL injections aren't possible.

job_list_params.go
```diff
@@ -0,0 +1,271 @@
...
+		id:    job.ID,
+		kind:  job.Kind,
+		queue: job.Queue,
+		time:  jobListTimeValue(job),
+	}
+}
+
+// UnmarshalText implements encoding.TextUnmarshaler to decode the cursor from
+// a previously marshaled string.
+func (c *JobListPaginationCursor) UnmarshalText(text []byte) error {
```
brandur, line 36: Just want to make sure I'm following: what's the purpose of this function? Is the idea here that it allows the cursor to be saved to JSON or otherwise persisted? Is that expected to be a feature that people want?

job_list_params.go
```diff
@@ -0,0 +1,271 @@
...
+	"strings"
+	"time"
+
+	"github.com/riverqueue/river/internal/dbadapter"
+	"github.com/riverqueue/river/rivertype"
+)
+
+// JobListPaginationCursor is used to specify a starting point for a paginated
+// job list query.
+type JobListPaginationCursor struct {
```
brandur, line 17: Shed color, but IMO `JobListCursor` might be an adequate name for this that's a little less of a mouthful. IMO use of "cursor" implies pagination.

internal/dbsqlc/job_list.go
```diff
@@ -0,0 +1,118 @@
+package dbsqlc
```
brandur, line 1: So I think my main (big) nit is mostly that this should live somewhere besides `dbsqlc` — I think that it's a little too confusing that this code _looks_ like it's generated, but it's not generated, and that `dbsqlc` shares a mix of generated and non-generated code. IMO, we should put it somewhere else even if it looks almost exactly like `dbsqlc` code.

brandur (commented):
Wicked! A few nits here, but I think it looks great and very close to mergeable.

Sorry about the delay on this — been doing a lot the last week and it's been a bit hard to concentrate, but will have more time next week.

internal/dbadapter/db_adapter.go
bgentry: Yes indeed. The thought is we'll only expose it in ways that do not accept raw user input. Obviously it would be ideal if we had a fully flexible query builder to do this all in a safe way, but barring that we'll just need to be mindful when expanding use of this internal API to ensure we don't add any SQL injection opportunities. Or if we do that would have to be very clearly called out as dangerous, but hopefully we won't find a reason to need to do that.

job_list_params.go
bgentry: `MarshalText` is the fallback for when a type does not implement `MarshalJSON`. This allows it to be marshaled and unmarshaled cleanly as an opaque JSON string, one which River can then ingest to set the cursor for the next page query.

In an API that powers a UI, for example, I'd expose the cursor so that the frontend can send it back with a subsequent request to fetch the next page.

internal/dbsqlc/job_list.go
bgentry: I moved this into `db`, not sure if you have anything better in mind. It's getting a little confusing with `dbsqlc` + `dbadapter` + `db` all in there 🤔

job_list_params.go
bgentry: Thanks, renamed it ✌️

client.go
```diff
@@ -1188,3 +1188,76 @@ func validateQueueName(queueName string) error {
...
+
+	internalJobs, err := c.adapter.JobList(ctx, *dbParams)
+	if err != nil {
+		return nil, err
+	}
+	return dbsqlc.JobRowsFromInternal(internalJobs), nil
+}
+
+// JobListTx inserts many jobs at once using Postgres' `COPY FROM` mechanism,
```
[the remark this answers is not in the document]
bgentry: Oops, thank you, fixed!

client.go
bgentry: I tweaked the name and text to be a little more generalized, lmk your thoughts.

bgentry:
@brandur freshly rebased, and expanded to allow for filtering to match a metadata fragment with `.Metadata("{\"foo\": \"bar\"}")`. I also made the `state` filter not required, though there still is a default value and you have to explicitly choose to clear it out with `.State("")`.

It's definitely pretty ugly internally, but I think it works for these limited use cases. I'd like to clean it up at some point but I think that's lower priority given that the ugliness is well contained here.

job_list_params.go
```diff
@@ -0,0 +1,288 @@
...
+	SortOrderDesc
+)
+
+// JobListOrderByField specifies the field to sort by.
+type JobListOrderByField int
+
+const (
+	// JobListOrderByTime specifies that the sort should be by time. The specific
+	// time field used will vary by job state.
+	JobListOrderByTime JobListOrderByField = iota
```
brandur, line 98: Are you sure about this one? Wouldn't it be better just to have an option for each timestamp field that you'd specify explicitly for what to sort on?

Currently, you have to read source code to understand what this will order by, and even if it were to be documented, it might be faster/easier to have users explicitly specify the timestamp they want.

brandur (commented):
Sorry for the delay on this review. Thx for the fixes.

client.go
```diff
@@ -1197,3 +1197,58 @@ func validateQueueName(queueName string) error {
 	}
 	return nil
 }
+
+// JobList returns a paginated list of jobs matching the provided filters. The
+// provided context is used for the underlying Postgres query and can be used to
+// cancel the operation or apply a timeout.
+//
+//	params := river.JobListParams{}.WithLimit(10).State(river.JobStateCompleted)
```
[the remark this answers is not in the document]
bgentry: These are a mistake, it's meant to always use the constructor. Thanks for catching!

job_list_params.go
bgentry: I'm reasonably sure that this is the behavior I want in the UI, though I could easily extract the logic that maps state to sort field into that external library. On the other hand, it'd be easy for us to add additional field-specific sort options in here as well. Part of the issue is that it doesn't necessarily make sense to sort by i.e. `finalized_at` when listing jobs that aren't yet finalized, so you'd get a bunch of jobs with `NULLS FIRST` at the beginning of the list. You could default to `scheduled_at` but that doesn't seem particularly relevant when listing jobs that have already completed.

Do you specifically think this behavior is bad and confusing to the extent that it shouldn't be included at all, or that we should have a different default, or better documentation, or just that we shouldn't ship this without additional options?

brandur: > Do you specifically think this behavior is bad and confusing to the extent that it shouldn't be included at all, or that we should have a different default, or better documentation, or just that we shouldn't ship this without additional options?

Oh we should definitely have list ordering behavior included — I was thinking more the difference of making it explicit versus implicit. Implicit would be more like you have here where you tell it to sort by time and you get back certain behaviors depending on context. Explicit would be more like you specifically ask to sort on `ScheduledAt`, `FinalizedAt`, etc.

The documentation could be improved, but IMO a case could still be made that this can still be slightly suboptimal because you need to read the docs to know what it's going to do instead of it being pretty obvious.

I see what you mean that this is definitely a sensible thing from the UI's perspective, although I think there's a chance that this should behave more like a dumb ORM and require everything to be explicit from the programmatic API here.

All that said, I guess explicit order by fields could be added in the future and cohabitate peacefully with this one, so having it available isn't harmful. Not strongly against if you want to go this direction.

brandur (approved):
@bgentry Thanks for putting in those feedback changes. Let me know if you want to discuss that order thing more, but don't let me block you on this any further. Marking approved.

bgentry:
Thanks, I suspect we'll revisit the sort params here as soon as people start using it. For now it at least lets me keep moving forward to get this merged.

PR #120, opened 2023-12-17
Fix data race in rescuer

brandur:
Here, the data race described here in the rescuer. It's a little subtle,
but the basic problem is that `defer` statements are LIFO, so what can
happen is that the rescuer can be closed while there's still a log
statement deferred on the stack. Because the log statements feed to
`t.Logf` through a shunt, this can occasionally cause a log line to be
sent to test output after the test has ended, causing the race detector
to detect a race.

The solution is to `defer` the close operation before the log operation.
All the other maintenance services already had this race condition
patched quite some time ago, but the fix didn't make it to the rescuer,
probably due to an outdated branch or the like.

[1] https://github.com/riverqueue/river/pull/116#issuecomment-1858986062

### DIFF
--- a/internal/maintenance/rescuer.go
+++ b/internal/maintenance/rescuer.go
@@ -112,10 +112,12 @@ func (s *Rescuer) Start(ctx context.Context) error {
 	s.CancellableSleepRandomBetween(ctx, JitterMin, JitterMax)
 
 	go func() {
+		// This defer should come first so that it's last out, thereby avoiding
+		// races.
+		defer close(stopped)
+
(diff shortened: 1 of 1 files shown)


bgentry (approved):
Brilliant, thanks. Looks like we already did this in all the other maintenance processes but somehow forgot this one.

Is this worth a changelog entry?

brandur:
Thanks!

> Is this worth a changelog entry?

TBH, probably not. In real life your logger is never going to disappear before or on the moment you stop your client, so I'm not sure you'd ever realistically run into a problem. I'll skip one for now for brevity sake, but lemme know if you disagree.

PR #121, opened 2023-12-19
fix ScheduledAt functionality with InsertManyTx

bgentry:
I stumbled upon this fairly egregious bug in my own test app. It turns out when you insert with `InsertManyTx`, `ScheduledAt` is not respected and we never even added it to the underlying query. Instead, the job is enqueued immediately as available and worked right away.

This fixes the bug by respecting `ScheduledAt` the same on bulk inserts as for individual inserts. I added a test to verify the broken behavior and then fixed it.

I also pulled in a sqlc upgrade commit from #117 to avoid needing to work around having a newer version locally.


brandur (approved):
lgtm!

PR #125, opened 2023-12-21
DOC update migrator creation is no longer returning error as second arg

GGuinea:


GGuinea:
@brandur @bgentry

brandur:
Thanks for the fix!

brandur (approved):

PR #126, opened 2023-12-22
Add changelog header for 0.0.15

brandur:
Seems like a good idea to cut a version that includes #121. Here, add a
banner for 0.0.15 to the changelog so we can cut a release.


brandur:
@bgentry As discussed, going to try and reuse the 0.0.15 release that's in the `go.mod` files already, but isn't tagged.

bgentry (approved):

brandur:
Thanks!

brandur:
Cut v0.0.15 and it's definitely pushed this time.

PR #129, opened 2023-12-23
Expose LISTEN pubsub functionality

bgentry:
We already have a robust LISTEN/NOTIFY implementation internally. It would potentially be useful for a variety of user-space coordination activities.

Expose it by wrapping our internal notifier package in a minimal public API. Ensure that we validate topic names in the process.

TODO: add test coverage

### DIFF
--- a/client.go
+++ b/client.go
@@ -542,6 +542,35 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 	return client, nil
 }
 
+// pgIdentifierRegexp matches all valid PostgreSQL identifier names:
+var pgIdentifierRegexp = regexp.MustCompile("^[a-zA-Z_][a-zA-Z0-9_]*$")
+
+// Listen subscribes to a notification topic. The notifyFunc will be called for
+// any notification that is received on the topic. The returned subscription must be
+// closed with Unlisten when no longer needed.
+//
+// The provided notifyFunc should be fast and not block to avoid dropping other
+// notifications and potentially causing internal issues within River.
+//
+// It returns an error if the specified topic is not a valid Postgres
+// identifier. Panics if notifyFunc is nil.
+func (c *Client[TTx]) Listen(topic string, notifyFunc NotifyFunc) (*ListenSubscription, error) {
+	// Validate that the topic is a valid Postgres identifier:
+	if valid := pgIdentifierRegexp.MatchString(topic); !valid {
+		return nil, fmt.Errorf("invalid topic name %q")
+	}
+	if notifyFunc == nil {
+		panic("notifyFunc is required")
+	}
+
+	wrapperFunc := func(topic notifier.NotificationTopic, payload string) {
+		notifyFunc(string(topic), payload)
+	}
+	return &ListenSubscription{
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -618,6 +647,11 @@ func (c *Client[TTx]) Start(ctx context.Context) error {
 
 		c.wg.Add(2)
 		go func() {
+			// TODO(bgentry): is this the wrong context? Should the notifier actually
+			// use the `workCtx` so that it doesn't get shut down before existing jobs
+			// have finished / had their contexts cancelled? This would preserve the
+			// ability to cancel an individual job's context during the initial
+			// shutdown phase.
 			c.notifier.Run(fetchNewWorkCtx)
 			c.wg.Done()
```
bgentry, line 656: I noticed this while I was in here and wanted to ask about it. If we shut down the notifier at the same time as we stop fetching new work, that means the notifier is no longer available for ongoing jobs (including potentially for the individualized cancellation of ongoing jobs). That seems wrong. Perhaps we should use the `workCtx` instead? That will require a bit of refactoring for the `Stop` flow because we'd want to call `workCancel` as soon as the producers finish but before waiting on the notifier. Currently, the notifier is bundled together into a waitgroup with producers.

client.go
```diff
@@ -542,6 +542,35 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
...
+// any notification that is received on the topic. The returned subscription must be
+// closed with Unlisten when no longer needed.
+//
+// The provided notifyFunc should be fast and not block to avoid dropping other
+// notifications and potentially causing internal issues within River.
+//
+// It returns an error if the specified topic is not a valid Postgres
+// identifier. Panics if notifyFunc is nil.
+func (c *Client[TTx]) Listen(topic string, notifyFunc NotifyFunc) (*ListenSubscription, error) {
+	// Validate that the topic is a valid Postgres identifier:
```
bgentry, line 558: @brandur should we also forbid the use of `river_` prefix topics? Or do you think there's no real risk in allowing it? I've gone back and forth but atm I'm not concerned.

brandur: Hmm, doesn't hurt to start with constraints first and relax them later if it seems necessary.

client.go
brandur: Not much to add, but +1 — seems wrong as is. I'm a little scared to add another stop step since it's already somewhat difficult to understand the progression of each step as is. IMO, might not be a bad idea to try and refactor stopping at some point so that each step is easily visible in sequence (i.e. most of the logic lives in one place and you can easily scroll through each phase).

brandur (commented):
@bgentry I'd be a little reluctant to expose this on the client since it doesn't have much to do with River's core job working functionality. If a user wants to listen/notify on a channel, wouldn't it be better to push them towards using their own pgx primitives instead of going through River?

If this is for use with some of our other planned features, I wonder if we should try to put this in the driver API which is a little out of the way, and which we've marked as clearly subject to change. Thoughts?

geekodour:
Will adding this feature allow us to be notified of a completion of a job, I want to send a realtime update of the status of the job to the fronetend and to other parts of the backend. Think this should be already possible right? Does this add something else?

brandur:
@geekodour Yeah, exactly. The subscriber interface is the way to go for that.

PR #134, opened 2023-12-30
fix typo in leadership resignation query

bgentry:
`s/river_leaderhip/river_leadership` 🤦‍♂️

### DIFF
--- a/internal/dbsqlc/river_leader.sql
+++ b/internal/dbsqlc/river_leader.sql
@@ -40,7 +40,7 @@ WITH currently_held_leaders AS (
 ),
 notified_resignations AS (
   SELECT
-    pg_notify('river_leaderhip', json_build_object('name', name, 'leader_id', leader_id, 'action', 'resigned')::text),
+    pg_notify('river_leadership', json_build_object('name', name, 'leader_id', leader_id, 'action', 'resigned')::text),
     currently_held_leaders.name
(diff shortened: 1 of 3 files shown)


internal/dbsqlc/river_leader.sql
```diff
@@ -40,7 +40,7 @@ WITH currently_held_leaders AS (
 ),
 notified_resignations AS (
   SELECT
-    pg_notify('river_leaderhip', json_build_object('name', name, 'leader_id', leader_id, 'action', 'resigned')::text),
+    pg_notify('river_leadership', json_build_object('name', name, 'leader_id', leader_id, 'action', 'resigned')::text),
```
bgentry, line 43: Actually, another way to fix this would be to pass in the notification name using the in-code constant. That would ensure we're using the same value everywhere.

Thoughts?

brandur: Yeah, good idea. It's already referenced in code anyway, so couldn't hurt just to use the same ref here.

brandur (approved):
Doh! LGTM.

internal/dbsqlc/river_leader.sql
bgentry: I’ll do this in a follow up PR, there are a couple we can get rid of

PR #135, opened 2023-12-30
Expose metadata

bgentry:
I'm running into multiple experiments where I'd like to make use of this field, so I think it's time to do so. One of those is job cancellation, where I'd like to write a value to the metadata that can be used by the rescuer to avoid retrying a stuck `running` job that has already been manually cancelled (but never finished or got marked as such). That PR is just about ready, so it seemed like the right time to bring this in.


brandur (approved):
Cool. WFM.

PR #136, opened 2023-12-31
Pull most notification topic names from code

bgentry:
Rather than hardcoding these into the query, pulling them from code reduces the number of distinct references to these values that must be kept in sync.

Small refactor coming out of https://github.com/riverqueue/river/pull/134#discussion_r1438727909.


brandur (approved):
Excellent. Yep, LGTM.

PR #137, opened 2023-12-31
elector should use slog

pmenglund:
the elector uses plain `log` but should use `slog` instead


brandur (approved):
Yep, looks right. Thanks!

PR #138, opened 2023-12-31
return error when calling stop without first calling start

pmenglund:
If you call `Stop()` without first calling `Start()` you get a null pointer panic. Returning an error is much more user friendly.

### DIFF
--- a/client.go
+++ b/client.go
@@ -681,7 +681,7 @@ func (c *Client[TTx]) signalStopComplete(ctx context.Context) {
 // There's no need to call this method if a hard stop has already been initiated
 // by cancelling the context passed to Start or by calling StopAndCancel.
 func (c *Client[TTx]) Stop(ctx context.Context) error {
-	if c.fetchNewWorkCancel != nil {
+	if c.fetchNewWorkCancel == nil {
 		return errors.New("client not started")
 	}
 
--- a/client.go
+++ b/client.go
@@ -681,11 +681,11 @@ func (c *Client[TTx]) signalStopComplete(ctx context.Context) {
 // There's no need to call this method if a hard stop has already been initiated
 // by cancelling the context passed to Start or by calling StopAndCancel.
 func (c *Client[TTx]) Stop(ctx context.Context) error {
-	c.baseService.Logger.InfoContext(ctx, c.baseService.Name+": Stop started")
 	if c.fetchNewWorkCancel != nil {
 		return errors.New("client not started")
 	}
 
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -682,6 +682,10 @@ func (c *Client[TTx]) signalStopComplete(ctx context.Context) {
 // by cancelling the context passed to Start or by calling StopAndCancel.
 func (c *Client[TTx]) Stop(ctx context.Context) error {
 	c.baseService.Logger.InfoContext(ctx, c.baseService.Name+": Stop started")
+	if c.fetchNewWorkCancel != nil {
+		return errors.New("client not started")
+	}
```
brandur, line 687: @pmenglund What do you think about moving this to be above the log info line? I think it makes more sense for the log line to only be written in cases where a stop really is starting (i.e. not a false start).

client_test.go
```diff
@@ -389,6 +390,15 @@ func Test_Client_Stop(t *testing.T) {
 		}
 	}
 
+	t.Run("not started", func(t *testing.T) {
+		t.Parallel()
+		client := newTestClient(ctx, t, newTestConfig(t, nil))
+
+		err := client.Stop(ctx)
+		require.Error(t, err)
+		assert.Equal(t, "client not started", err.Error())
```
brandur, line 399: See lint error here, but we've tried to standardize around using `require` over `assert` everywhere.

brandur (commented):
Good idea! Left a couple comments here, but I like it.

client_test.go
```diff
@@ -389,6 +389,15 @@ func Test_Client_Stop(t *testing.T) {
 		}
 	}
 
+	t.Run("not started", func(t *testing.T) {
+		t.Parallel()
+		client := newTestClient(ctx, t, newTestConfig(t, nil))
+
+		err := client.Stop(ctx)
+		require.Error(t, err)
+		require.Equal(t, "client not started", err.Error())
```
pmenglund, line 398: another option for checking the error string, would be to expose it as a sentinel error, e.g.
`NotStartedErr`, but I saw that there are three defined in `client.go` but they are not exported, so I opted for this instead

brandur: Yeah. I think it's okay not to give them a full error type until we're more sure that someone will need it. This case is pretty clearly a misuse problem that'll rarely occur outside of dev and not need normal handling.

pmenglund: that was exactly how I found it - I messed up my tests setup and got a panic

PR #141, opened 2024-01-01
add Cancel API to cancel a running job

bgentry:
This adds a new `Client` method `Cancel` which can cancel a running job. It has these behaviors:

* Jobs which are still scheduled, retryable, or available are immediately cancelled and will not be run again.
* Jobs that are already finalized (completed, cancelled, or discarded) are not touched.
* Jobs that are actively running are marked for cancellation, and cancellation is attempted using a `LISTEN`/`NOTIFY` pubsub message which is picked up by the client and producer running that job.

Because Go offers no way to interrupt a running goroutine, actively running jobs cannot be immediately halted and cancelled, so we can only cancel the job's context. Once the cancellation signal is received by the client running the job, any error returned by that job will result in it being cancelled permanently and not retried. However if the job returns no error, it will be completed as usual.

In the event the running job finishes executing _before_ the cancellation signal is received, the job will be completed or retried as usual without regard to the cancellation. @brandur Realized as I was writing this that it was solvable, do you think we should try to fix this with a smarter completion query that uses the metadata value we set as part of the `Cancel` query?

Additional documentation to follow once we're satisfied with the shape and behavior of this API.

### DIFF
--- a/client.go
+++ b/client.go
@@ -312,6 +312,11 @@ func (ts *clientTestSignals) Init() {
 }
 
 var (
+	// ErrNotFound is returned when a query by ID does not match any existing
+	// rows. For example, attempting to cancel a job that doesn't exist will
+	// return this error.
+	ErrNotFound = errors.New("not found")
+
 	errMissingConfig                 = errors.New("missing config")
 	errMissingDatabasePoolWithQueues = errors.New("must have a non-nil database pool to execute jobs (either use a driver with database pool or don't configure Queues)")
 	errMissingDriver                 = errors.New("missing database driver (try wrapping a Pgx pool with river/riverdriver/riverpgxv5.New)")
@@ -933,6 +938,106 @@ func (c *Client[TTx]) runProducers(fetchNewWorkCtx, workCtx context.Context) {
 			producer.Run(fetchNewWorkCtx, workCtx, c.monitor.SetProducerStatus)
 		}()
 	}
+}
+
+// Cancel cancels the job with the given ID. If possible, the job is cancelled
+// immediately and will not be retried. The provided context is used for the
+// underlying Postgres update and can be used to cancel the operation or apply a
+// timeout.
+//
+// If the job is still in the queue (available, scheduled, or retryable), it is
+// immediately marked as cancelled and will not be retried.
+//
+// If the job is already finalized (cancelled, completed, or discarded), no
+// changes are made.
+//
+// If the job is currently running, it is not immediately cancelled, but is
+// instead marked for cancellation. The client running the job will also be
+// notified (via LISTEN/NOTIFY) to cancel the running job's context. Although
+// the job's context will be cancelled, since Go does not provide a mechanism to
+// interrupt a running goroutine the job will continue running until it returns.
+// As always, it is important for workers to respect context cancellation and
+// return promptly when the job context is done.
+//
+// Once the cancellation signal is received by the client running the job, any
+// error returned by that job will result in it being cancelled permanently and
+// not retried. However if the job returns no error, it will be completed as
+// usual.
+//
+// In the event the running job finishes executing _before_ the cancellation
+// signal is received but _after_ this update was made, the behavior depends
+// on which state the job is being transitioned into:
+//
+//   - If the job completed successfully, was cancelled from within, or was
+//     discarded due to exceeding its max attempts, the job will be updated as
+//     usual.
+//   - If the job was snoozed to run again later or encountered a retryable error,
+//     the job will be marked as cancelled and will not be attempted again.
+//
+// Returns the up-to-date JobRow for the specified jobID if it exists. Returns
+// ErrNotFound if the job doesn't exist.
+func (c *Client[TTx]) Cancel(ctx context.Context, jobID int64) (*rivertype.JobRow, error) {
+	job, err := c.adapter.JobCancel(ctx, jobID)
+	if err != nil {
+		if errors.Is(err, riverdriver.ErrNoRows) {
+			return nil, ErrNotFound
+		}
+		return nil, err
+	}
+
+	return dbsqlc.JobRowFromInternal(job), nil
+}
+
+// CancelTx cancels the job with the given ID within the specified transaction.
+// This variant lets a caller cancel a job atomically alongside other database
+// changes. An cancelled job doesn't take effect until the transaction commits,
+// and if the transaction rolls back, so too is the cancelled job.
+//
+// If possible, the job is cancelled immediately and will not be retried. The
+// provided context is used for the underlying Postgres update and can be used
+// to cancel the operation or apply a timeout.
+//
+// If the job is still in the queue (available, scheduled, or retryable), it is
+// immediately marked as cancelled and will not be retried.
+//
+// If the job is already finalized (cancelled, completed, or discarded), no
+// changes are made.
+//
+// If the job is currently running, it is not immediately cancelled, but is
+// instead marked for cancellation. The client running the job will also be
+// notified (via LISTEN/NOTIFY) to cancel the running job's context. Although
+// the job's context will be cancelled, since Go does not provide a mechanism to
+// interrupt a running goroutine the job will continue running until it returns.
+// As always, it is important for workers to respect context cancellation and
+// return promptly when the job context is done.
+//
+// Once the cancellation signal is received by the client running the job, any
+// error returned by that job will result in it being cancelled permanently and
+// not retried. However if the job returns no error, it will be completed as
+// usual.
+//
+// In the event the running job finishes executing _before_ the cancellation
+// signal is received but _after_ this update was made, the behavior depends
+// on which state the job is being transitioned into:
+//
+//   - If the job completed successfully, was cancelled from within, or was
+//     discarded due to exceeding its max attempts, the job will be updated as
+//     usual.
+//   - If the job was snoozed to run again later or encountered a retryable error,
+//     the job will be marked as cancelled and will not be attempted again.
+//
+// Returns the up-to-date JobRow for the specified jobID if it exists. Returns
+// ErrNotFound if the job doesn't exist.
+func (c *Client[TTx]) CancelTx(ctx context.Context, tx TTx, jobID int64) (*rivertype.JobRow, error) {
+	job, err := c.adapter.JobCancelTx(ctx, c.driver.UnwrapTx(tx), jobID)
+	if errors.Is(err, riverdriver.ErrNoRows) {
+		return nil, ErrNotFound
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	return dbsqlc.JobRowFromInternal(job), nil
 }
 
 func insertParamsFromArgsAndOptions(args JobArgs, insertOpts *InsertOpts) (*dbadapter.JobInsertParams, error) {
--- a/client.go
+++ b/client.go
@@ -965,8 +965,8 @@ func (c *Client[TTx]) runProducers(fetchNewWorkCtx, workCtx context.Context) {
 // usual.
 //
 // In the event the running job finishes executing _before_ the cancellation
-// signal is received but _after_ this update was made, the behavior depends
-// on which state the job is being transitioned into:
+// signal is received but _after_ this update was made, the behavior depends on
+// which state the job is being transitioned into (based on its return error):
 //
 //   - If the job completed successfully, was cancelled from within, or was
 //     discarded due to exceeding its max attempts, the job will be updated as
@@ -1017,8 +1017,8 @@ func (c *Client[TTx]) Cancel(ctx context.Context, jobID int64) (*rivertype.JobRo
 // usual.
 //
 // In the event the running job finishes executing _before_ the cancellation
-// signal is received but _after_ this update was made, the behavior depends
-// on which state the job is being transitioned into:
+// signal is received but _after_ this update was made, the behavior depends on
+// which state the job is being transitioned into (based on its return error):
 //
 //   - If the job completed successfully, was cancelled from within, or was
 //     discarded due to exceeding its max attempts, the job will be updated as
--- a/client_test.go
+++ b/client_test.go
@@ -216,7 +216,7 @@ func Test_Client(t *testing.T) {
 		riverinternaltest.WaitOrTimeout(t, workedChan)
 	})
 
-	t.Run("JobCancel", func(t *testing.T) {
+	t.Run("JobCancelErrorReturned", func(t *testing.T) {
 		t.Parallel()
 
 		client, bundle := setup(t)
@@ -245,7 +245,7 @@ func Test_Client(t *testing.T) {
 		require.WithinDuration(t, time.Now(), *updatedJob.FinalizedAt, 2*time.Second)
 	})
 
-	t.Run("JobSnooze", func(t *testing.T) {
+	t.Run("JobSnoozeErrorReturned", func(t *testing.T) {
 		t.Parallel()
 
 		client, bundle := setup(t)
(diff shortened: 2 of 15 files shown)


client.go
```diff
@@ -935,6 +935,46 @@ func (c *Client[TTx]) runProducers(fetchNewWorkCtx, workCtx context.Context) {
...
+//
+// Once the cancellation signal is received by the client running the job, any
+// error returned by that job will result in it being cancelled permanently and
+// not retried. However if the job returns no error, it will be completed as
+// usual.
+//
+// In the event the running job finishes executing _before_ the cancellation
+// signal is received, the job will be completed or retried as usual without
+// regard to the cancellation. TODO: fix this w/ smarter completion query that
+// uses metadata value?
```
bgentry, line 965: See the question in the PR description.

client_test.go
```diff
@@ -274,6 +274,94 @@ func Test_Client(t *testing.T) {
...
+	t.Run("CancelNonExistentJob", func(t *testing.T) {
+		t.Parallel()
+
+		client, _ := setup(t)
+		startClient(ctx, t, client)
+
+		// Cancel an unknown job ID:
+		cancelled, err := client.Cancel(ctx, 0)
+		// TODO(bgentry): should we try to make this return an error even though the
+		// query was successfully a no-op? or is it fine that it returns false, nil?
```
bgentry, line 360: I initially wrote this test expecting the test to error when the row didn't exist, but of course that didn't happen. We could in theory rework the implementation to cause it to error like with an `ErrNoRows` type of thing, but I'm not sure that's desirable or necessary. Thoughts?

I think I'm inclined to just keep it a quiet no-op and don't rely on `Cancel` to tell you whether or not a job exists (we should expose a "get" method for that).

client.go
brandur: IMO: if the job was about to be completed, just let that happen without the cancellation superseding it. However, if a job that was cancelled were to be set as retryable with the possibility of running again, that definitely seems wrong and should be fixed.

client_test.go
brandur: Hmm, I definitely would've leaned toward returning an error on a job that doesn't exist at all. The trouble is that it's quite easy in Go to pass the wrong value since all types are only "moderately strong".

e.g. You might be holding an `int64` that you think is a job ID, but is actually the primary ID of a different entity in your DB, pass that in expecting a cancellation, but never being the wiser that the wrong thing is happening because you're expecting a cancellation to be somewhat of an edge case anyway that's often a no-op.

internal/maintenance/rescuer.go
```diff
@@ -146,10 +146,15 @@ func (s *Rescuer) Start(ctx context.Context) error {
 }
 
 type rescuerRunOnceResult struct {
+	NumJobsCancelled int64
 	NumJobsDiscarded int64
 	NumJobsRetried   int64
 }
 
+type metadataWithCancelledByQuery struct {
+	CancelledByQuery bool `json:"cancelled_by_query"`
```
brandur, line 155: What do you mean exactly by "cancelled by query"? Does that just mean that it was cancelled and cancellation happens by a query so that it was cancelled by a query?

producer.go
```diff
@@ -150,12 +155,47 @@ func (p *producer) Run(fetchCtx, workCtx context.Context, statusFunc producerSta
...
 	p.fetchAndRunLoop(fetchCtx, workCtx, fetchLimiter, statusFunc)
 	statusFunc(p.config.QueueName, componentstatus.ShuttingDown)
 	p.executorShutdownLoop()
 	statusFunc(p.config.QueueName, componentstatus.Stopped)
 }
 
+type jobControlAction string
+
+const (
+	jobControlActionCancel jobControlAction = "cancel"
```
brandur, line 190: OOC, are there any other actions you were expecting to come in?

client.go
```diff
@@ -935,6 +935,90 @@ func (c *Client[TTx]) runProducers(fetchNewWorkCtx, workCtx context.Context) {
...
+//
+// Returns true if the job was cancelled or cancellation was initiated, and
+// false if this was a no-op because the job was already finalized.
+func (c *Client[TTx]) Cancel(ctx context.Context, jobID int64) (bool, error) {
```
brandur, line 969: I kinda would've leaned toward having this return a job row representing the updated job for purposes of logging and additional checking. Thoughts?

A side benefit would be that it'd also allow you to disambiguate between the different edge conditions too. i.e. If the job was completed before it was cancelled, you could check its final state to see that.

brandur (commented):
Awesome. Looking forward to see this coming in!

A couple high level thoughts:

* I know we discussed this somewhat on a call the other week, but given that cancellation is quite a core feature, I think I would've leaned towards bringing it right into the `river_job` schema rather than deferring it to the `metadata` K/V blob. Hard schema is generally preferable IMO because not only is it safe against potential name conflicts in `metadata`, but it's just easier to query and update etc.
* In my experience in `metadata` or out, it's usually better to prefer timestamps like `cancelled_at` rather than straight up booleans like `is_cancelled` because the former providers strictly more information — you can check the cancellation state on either a timestamp or a boolean, but the timestamp adds context to the equation by letting you know when it was cancelled, which is helpful in some situations.

client.go
bgentry: Yes, I agree. That would require changes only to the queries which would mark the job as errored, because in that case you'd want to realize that the cancellation had been attempted and you'd want that to take precedence.

client.go
bgentry: I started with that initially and probably should have kept on in that direction. The difficulty is it makes the query just a bit trickier because you have to lock the specified row no matter what, and return it whether or not it was cancelled/updated. That's only a minor challenge, though.

It probably does make it more important to offer a first-class way to tell that a cancellation has been attempted (not merely a metadata field per your other comment).

> A side benefit would be that it'd also allow you to disambiguate between the different edge conditions too. i.e. If the job was completed before it was cancelled, you could check its final state to see that.

You wouldn't necessarily get this in all cases, at least not as a result of this API. The notification to cancel the job doesn't get emitted until the transaction here is committed, and there's a lag before the notification is received and processed by the producer running the job. But after-the-fact, you would be able to look at the job's row and see that the cancellation was attempted but that the job was still completed. A timestamp (as you suggested) makes this more useful because you'd see that these events were within milliseconds of each other.

client_test.go
bgentry: Yeah, I think that's doable, although we don't have any such "not found" error defined yet. I'm not actually sure where else it would make sense—perhaps a soon-to-be-added `JobGet` API because you similarly wouldn't want to return `nil, nil` there?

internal/maintenance/rescuer.go
bgentry: It's a little tricky because of the inherent raciness of this operation. The cancel query itself is guaranteed to be completed here, but the job isn't actually cancelled yet at the moment this is set because that's an asynchronous operation which requires the producer to receive the signal, act on it, and the job to then exit. And there are the edge cases where the job completes or gets discarded before the cancel signal is received.

Perhaps `cancel_attempted_at` or `cancel_initiated_at` is a suitable name for this attr and/or a column?

producer.go
bgentry: I don't have anything else in mind that's specific to a single in-progress job. You could retry a single job immediately (no matter its state), but that doesn't require a notification because it's just a matter of updating the job row.

bgentry:
> * I know we discussed this somewhat on a call the other week, but given that cancellation is quite a core feature, I think I would've leaned towards bringing it right into the `river_job` schema rather than deferring it to the `metadata` K/V blob. Hard schema is generally preferable IMO because not only is it safe against potential name conflicts in `metadata`, but it's just easier to query and update etc.

Yeah, I think my hesitation here is I'm a bit reluctant to add columns generally because of the potential impact on table bloat and performance. One additional timestamp is probably fine, although you could argue that it's being used to cover some fairly rare edge cases and that alone may not be worth the potential impact of adding the column. Plus, you'd need a migration to do this, and we have to decide if that's worth doing for this particular purpose.

That said, if we go this route, I have a few other unrelated things in other branches that we'd want to bundle together with it (non-null on `args`, adding a `pending` state).

> * In my experience in `metadata` or out, it's usually better to prefer timestamps like `cancelled_at` rather than straight up booleans like `is_cancelled` because the former providers strictly more information — you can check the cancellation state on either a timestamp or a boolean, but the timestamp adds context to the equation by letting you know when it was cancelled, which is helpful in some situations.

I agree a timestamp would have been better and I should have thought to do that! Even just for debugging some of these race conditions it'd be useful to see that the job was completed/finalized mere ms after the cancellation attempt.

client.go
bgentry: I started working on this, but quickly realized it's a bit more nuanced than I expected. [This query](https://github.com/riverqueue/river/blob/f3fabdff6c2010b93938c0bd1602d8f14ec5fb09/internal/dbsqlc/river_job.sql#L241-L259) is the one making the update from the completer. I wanted to confirm the desired outcome for each of the potential states it could transition into:

* `completed`: this one's easy, just let it complete / no changes
* `cancelled`: also easy, let it cancel / no changes
* `discarded`: since the job is going to finalize and not retry again, I think this one is also fine to not change?
* `retryable`: we don't want additional retries to proceed, so we should (a) set a `finalized_at` and (b) override the `state` to `cancelled`. But we won't have a `@finalized_at` input here because the caller hadn't intended to finalize it, so I guess I'd just use `now()` in the query?
* `scheduled`: I guess similar to above, you would also want to override this one to set a `finalized_at` and a `state` of `cancelled`.

I implemented the above behavior just now, please take a look. I haven't yet added tests because I want to make sure this all sounds reasonable to you first (the tests may be tricky).

internal/dbsqlc/river_job.sql
```diff
@@ -240,18 +285,23 @@ RETURNING *;
...
-        finalized_at = CASE WHEN @finalized_at_do_update::boolean THEN @finalized_at                       ELSE finalized_at END,
+    SET
+        finalized_at = CASE WHEN @finalized_at_do_update::boolean THEN @finalized_at
+          WHEN should_cancel THEN now()
+          ELSE finalized_at END,
+        errors       = CASE WHEN @error_do_update::boolean        THEN array_append(errors, @error::jsonb) ELSE errors       END,
         max_attempts = CASE WHEN @max_attempts_update::boolean    THEN @max_attempts                       ELSE max_attempts END,
         scheduled_at = CASE WHEN @scheduled_at_do_update::boolean THEN @scheduled_at                       ELSE scheduled_at END,
-        state = @state
+        state        = CASE WHEN should_cancel THEN 'cancelled'::river_job_state                           ELSE @state::river_job_state END
```
bgentry, line 304: TODO: these changes to `JobSetStateIfRunning` are not tested. Waiting for feedback before adding full coverage for the new bits.

bgentry: j/k I added coverage for this and reworked the query to fix a minor bug I found. It now avoids touching any fields unnecessarily when overriding with a `should_cancel` _except_ for `errors`, which still gets appended to if there's an error to write.

bgentry:
@brandur I think I've made all the changes here except for the potential dedicated DB column. Interested in your thoughts on that based on my above points. If we go that route, we'll want to ship a single migration that includes a handful of changes batched together, so we'll need to coordinate on that.

client_test.go
bgentry: Added a top level `ErrNotFound` and updated `Cancel`/`CancelTx` to rescue `riverdriver.ErrNoRows` and instead return the new error.

client_test.go
```diff
@@ -274,6 +275,121 @@ func Test_Client(t *testing.T) {
...
+	t.Run("CancelNonExistentJob", func(t *testing.T) {
+		t.Parallel()
+
+		client, _ := setup(t)
+		startClient(ctx, t, client)
+
+		// Cancel an unknown job ID:
+		jobAfter, err := client.Cancel(ctx, 0)
+		// TODO(bgentry): do we want to expose a different error type for this
+		// externally since we don't yet want to expose riverdriver publicly?
```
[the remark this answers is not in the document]
bgentry: Updated to return a new `ErrNotFound` from the `Client` when `Cancel` or `CancelTx` gets back a `riverdriver.ErrNoRows` from the adapter call.

client_test.go
```diff
@@ -274,6 +274,130 @@ func Test_Client(t *testing.T) {
...
+
+		AddWorker(client.config.Workers, WorkFunc(func(ctx context.Context, job *Job[JobArgs]) error {
+			jobStartedChan <- job.ID
+			<-ctx.Done()
+			return ctx.Err()
+		}))
+
+		statusUpdateCh := client.monitor.RegisterUpdates()
+		startClient(ctx, t, client)
+		waitForClientHealthy(ctx, t, statusUpdateCh)
```
bgentry, line 298: I got a flaky failure on these tests in [this run](https://github.com/riverqueue/river/actions/runs/7476889516/job/20348329965). There wasn't anything particularly helpful in the failure logs, other than that the job kept running and didn't get cancelled. My guess, however, is that the notifier wasn't up and running before the cancel signal was sent.

I tried reproducing it with thousands of runs locally but couldn't get any failures. Ultimately I added this logic to wait for internal components to come up and be healthy prior to proceeding with the test.

It makes me wonder if there's a more systemic issue here; should we be waiting for some components to come up before returning from `Start()`? Or should our `startClient` test helper always be waiting for the client to become healthy before returning?

brandur (approved):
> Yeah, I think my hesitation here is I'm a bit reluctant to add columns generally because of the potential impact on table bloat and performance. One additional timestamp is probably fine, although you could argue that it's being used to cover some fairly rare edge cases and that alone may not be worth the potential impact of adding the column. Plus, you'd need a migration to do this, and we have to decide if that's worth doing for this particular purpose.
>
> That said, if we go this route, I have a few other unrelated things in other branches that we'd want to bundle together with it (non-null on `args`, adding a `pending` state).

K. The one counter is that in my experience at least, schema is basically always better because it's more visible, more discoverable, provides tab complete, etc., so if we're not going that direction, we should make sure it's for the right reasons, rather than just because it's easier.

I don't think table bloat/perf is a concern if I understand the Postgres page layout. Because this would be a nullable field, a value wouldn't even be present for most rows, and it'd just require one extra bit in the NULL bitmask. Read this and see if I'm out to lunch on that:

https://www.postgresql.org/docs/current/storage-page-layout.html#STORAGE-TUPLE-LAYOUT

It seems okay to start with less schema and then batch up schema changes together into one migration. We'd have to be careful to try and avoid and migration queries that are too expensive (i.e. we might not even bother converting older cancelled job rows to the new format if the table is too large), but should be doable.

client_test.go
bgentry: We could easily update `startClient` to wait for the client to become healthy. That should reduce flakiness and shouldn't impact test timing much as everything starts quickly. However since there is no client health exposed externally yet, it's purely an internal solution for testing.

There's the goal of making sure the client is healthy and fully booted before proceeding; this may be desirable _sometimes_, but other times users may wish to allow the client to finish booting asynchronously while their app does other initialization work. But then there's the separate goal of monitoring the client's health as it operates, and any of these internal components could in theory have issues or connectivity issues during operation.

I'm sure there are other libraries which try to tackle these problems, but one particular instance I have recent experience with is [LaunchDarkly's SDK](https://docs.launchdarkly.com/sdk/features/monitoring#go). It's a bit poorly documented and feels overly abstracted / Java-like, but the concepts there are reasonably well-solved and that's what I was intending to do when I initially built the client monitor stuff internally. I don't claim that the design is perfect or finished but at least the general problem is somewhat solved by it.

We should probably take a look at this both for our own tests but also for users' sake in follow up PRs.

bgentry:
@brandur I just added an example here demonstrating remote job cancellation. Notably, it suffers from the exact problem we were discussing around not being able to ensure that the client is ready and needing to sleep to get around that 😕 

Feel free to review, we can always improve the example after this PR gets merged (which I'll do as soon as CI goes green).

PR #142, opened 2024-01-04
add UniqueSkippedAsDuplicate field to JobRow

chrboe:
When inserting a job, the insert may be rejected because the configured uniqueness constraints for the job are not met.
With the current API, this happens silently and without a way for the caller to check for this condition.

Since we already record the UniqueSkippedAsDuplicate in the internal dbadapter.JobInsertResult, we just have to propagate this value to the external client API.

Callers can now check the UniqueSkippedAsDuplicate field of JobRow to check for the condition described above.

---

This is an RFC PR to keep the discussion on #86 going. I am pretty sure this is the "least breaking" API change possible to implement this.

@bgentry @brandur please advise if you like this approach or if you have other suggestions.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1066,7 +1066,9 @@ func (c *Client[TTx]) InsertTx(ctx context.Context, tx TTx, args JobArgs, opts *
 		return nil, err
 	}
 
-	return dbsqlc.JobRowFromInternal(res.Job), nil
+	jobRow := dbsqlc.JobRowFromInternal(res.Job)
+	jobRow.UniqueSkippedAsDuplicate = res.UniqueSkippedAsDuplicate
+	return jobRow, nil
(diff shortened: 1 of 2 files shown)


brandur:
Thanks @chrboe!

Hmm, let me chew on this a bit longer. I think this is good, and I realistically I wouldn't expect this to cause any practical problems, but something feels a little off to me about putting a non-job-row property onto the `JobRow` struct, which is otherwise a close reflection of the DB schema. I'd prefer if we could find a good way of making this available "out of band" in some way like an alternate return value or something of that nature.

brandur:
@chrboe Going to close this out in favor of the similar, but slightly different approach in #292. Thanks for taking a stab at this!

PR #145, opened 2024-01-11
add Client to work context, helpers to extract it

bgentry:
This change adds the `*Client` to an unexported context key within the work context when the Client is started, making it available to the context for every job that's worked by that Client.

Additionally, two new helper functions are added to retrieve the Client from the context within a Worker: `ClientFromContext` and `ClientFromContextSafely`. The only difference between these is that the first variant panics if the client isn't found in the context, whereas the latter returns an error. However if these are used on the context provided to a Worker's `Work()` method (or one derived from it) they should never panic/error.

Inspired by #87 and https://github.com/riverqueue/river/discussions/144.

If we want to move forward with this, it still needs:
* [x] A changelog entry
* [ ] Website documentation

### DIFF
--- a/context_test.go
+++ b/context_test.go
@@ -1,0 +1,31 @@
+package river
+
+import (
+	"context"
+	"testing"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/stretchr/testify/require"
+)
+
+func TestClientFromContext(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+	client := &Client[pgx.Tx]{}
+	ctx = withClient(ctx, client)
+
+	require.Equal(t, client, ClientFromContext[pgx.Tx](ctx))
+
+	result, err := ClientFromContextSafely[pgx.Tx](ctx)
+	require.NoError(t, err)
+	require.Equal(t, client, result)
+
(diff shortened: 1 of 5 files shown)


context_test.go
```diff
@@ -0,0 +1,31 @@
...
+	client := &Client[pgx.Tx]{}
+	ctx = withClient(ctx, client)
+
+	require.Equal(t, client, ClientFromContext[pgx.Tx](ctx))
+
+	result, err := ClientFromContextSafely[pgx.Tx](ctx)
+	require.NoError(t, err)
+	require.Equal(t, client, result)
+
+	require.Panics(t, func() {
```
brandur, line 24: What do you think about switching to the use of `require.PanicsWithError` here? Just makes sure we for sure got the right panic and didn't trigger some other unexpected one by accident.

brandur (approved):
Sweet, lgtm.

context_test.go
bgentry: Another handy `require` helper I didn't yet know about, thanks! Changed this test to assert about the specific error throughout, and also added a changelog entry.kj

bgentry:
I'm working on a docs PR but no reason not to merge this in the mean time.

PR #149, opened 2024-01-16
fix InsertMany with unspecified ScheduledAt

bgentry:
Due to a missing `IsZero()` check, `InsertMany` was inserting the Go zero time into the database, resulting in `ScheduledAt` times in the distant past. This didn't impact single inserts because that code path had a check of this sort and also relies on insertion-time coalescing & defaults.


brandur (approved):
Damn, I feel like I've fixed this one before. Must have been something else or in another branch.

bgentry:
It's tricky because we're using the same helper `insertParamsFromArgsAndOptions` for both the `Insert` and `InsertMany` functions, but the way they get handled varies internally due to some fields that are nullable in the normal insert query but not in the copy query. Thinking about how to avoid divergence in these longer term 🤔

PR #151, opened 2024-01-16
Mark additional helpers in `rivertest` as such

CGA1123:
`rivertest` is reporting itself as the site of a failure.

Looks like `Helper()` doesn't traverse down the callstack, so each intermediate
helper needs to be marked as such in order to push the report up to the actual
calling (user) test.


bgentry (approved):
Thank you @CGA1123 ! This turns out to be quite an annoying behavior of the `testing` package 😑

CGA1123:
Thanks @bgentry! 🙇

PR #152, opened 2024-01-16
rename Client.Cancel to Client.JobCancel

### LINKED ISSUE #150: API naming inconsistency
I realized while writing up the changelog entry for https://github.com/riverqueue/river/pull/117 that I had named that API `JobList()`, while I named the API from #141 `Cancel()` with no `Job` prefix. We have other job-related query APIs to add like `Get`, `RetryImmediately`, etc. These should probably all either _have_ a prefix or _not_ have one, but without inconsistency.

Thoughts @brandur? Should be an easy fix as we are still prerelease on the new API.

bgentry:
This was added in #141 / 0aaeee8. Meanwhile another job-related query API was added called `JobList` (#117), but `Cancel` and `CancelTx` were not given the `Job*` prefix.

This renames `Cancel` to `JobCancel` and `CancelTx` to `JobCancelTx` for consistency. Other upcoming job query APIs will also use the prefix.

Fixes #150.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added `Cancel` and `CancelTx` to the `Client` to enable cancellation of jobs. [PR #141](https://github.com/riverqueue/river/pull/141).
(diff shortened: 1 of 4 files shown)


brandur (approved):
Great, WFM!

PR #160, opened 2024-01-17
modified to go 1.20 in weavelab.xyz

mauza:


mauza:
Sorry, wrong org :smile:

bgentry:
@mauza not sure if you opened this on purpose but it's not something we can merge :)

Btw you should be able to take advantage of go.mod `replace` directives to point `github.com/riverqueue/river` at your own fork and avoid needing to change the import paths everywhere.

PR #162, opened 2024-01-19
smarter short snooze times

bgentry:
Normally, snoozed jobs are set `scheduled` for the future and it's the scheduler's job to set them back to `available` so they can be reworked. Just as with retryable jobs, this isn't friendly for short snooze times so we instead make the job immediately `available` if the snooze time is smaller than the scheduler's run interval.

We have an identical behavior for errored/retryable jobs for the same reason.


brandur (approved):
Sweet. Yep, makes sense.

PR #168, opened 2024-01-22
Respect job uniqueness when inserting periodic jobs

### LINKED ISSUE #161: Unique periodic jobs
The documentation states it is possible to combine unique and periodic jobs:

> Fortunately, many of these concerns can be addressed by combining periodic jobs with [unique jobs](https://riverqueue.com/docs/unique-jobs) and the RunOnStart option. For example, a job which is configured to be unique at the hourly level will only enqueue once in that hour no matter how many times it's attempted.

Unfortunately I was not able to make it work. After some source code studying I found out that periodic jobs are inserted using `JobInsertMany`. That explains why uniqueness does not take effect, as the docs say:

> ...which is why job uniqueness is not supported on [batch insertion](https://riverqueue.com/docs/batch-job-insertion).

Is it something that needs to be fixed, or am I just missing something?

brandur:
As reported in #161, unique jobs combined with periodic jobs doesn't
work as expected. This is because the periodic job enqueueing service
inserts jobs in bulk as an optimization in case many periodic jobs need
inserting, and bulk insertion doesn't support uniqueness.

Here, check for job uniqueness and make sure to use the single job
insertion path when a periodic job is unique. I left the bulk job
insertion path in as well, for use when job uniqueness isn't required.
The optimization still feels like a reasonable trade off to me because
it doesn't make the code that much harder to read, and I expect
non-unique periodic jobs to be the common case.

Fixes #161.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed incorrect default value handling for `ScheduledAt` option with `InsertMany` / `InsertManyTx`. [PR #149](https://github.com/riverqueue/river/pull/149).
+- Fixed problem where job uniqueness wasn't being respected when used in conjuction with periodic jobs. [PR #168](https://github.com/riverqueue/river/pull/168).
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


bgentry (approved):

brandur:
Thanks!

dhermes:
Amazing! Just saw this in the [release notes][1]. Thanks for the fix.

(I toyed with the uniqueness feature but only in the context of periodic jobs. I walked away thinking I hadn't figured it out yet because I couldn't get it to work the way I expected. Now maybe my puny human brain is only 90% to blame 😆!)

[1]: https://github.com/riverqueue/river/blob/master/CHANGELOG.md#0017---2024-01-22

PR #170, opened 2024-01-22
Add `river validate` command to check for unapplied migrations

brandur:
This one's a feature request from a few months from [1] in which we add
a `river validate` command which fails with a non-zero exit code in case
there are unapplied migrations in the target database. As described in
the original discussion, this could be used in a project's CI to detect
and fail in the presence of a new migration that was part of a River
version update (giving people an easy way to check programmatically so
that upgrades aren't as dependent on a user successfully seeing a new
migration in the changelog).

[1] https://github.com/riverqueue/river/discussions/68


cmd/river/go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/riverqueue/river/cmd/river
 
 go 1.21.4
 
-// replace github.com/riverqueue/river => ../..
+replace github.com/riverqueue/river => ../..
```
brandur, line 5: Uncommented to demonstrate that the whole package works, but I'll have to unwind this as I'm cutting a release. We have steps in `development.md` that'll theoretically work for this situation ...

bgentry (approved):
Looks great! Thanks :shipit:

brandur:
Great, thanks!

PR #172, opened 2024-01-22
Tee up `go.mod` and `CHANGELOG.md` for version 0.0.17

brandur:
Here, as proposed by the new release instructions in #169, tee up some
changes for 0.0.17, updating `CHANGELOG.md` with the new section and
modifying `go.mod` files to point to the new release.

Unfortunate note no. 1: the `go.mod` files change from 0.0.15 to 0.0.17
instead of starting at 0.0.16 because I accidentally botched the last
release by not pushing the `go.mod` changes. Hopefully this new process
will correct that so it doesn't happen again.

Unfortunate note no. 2: As noted in #169, we expect the build to fail
here because `go.mod` files point to a version that doesn't have a tag
yet.

The new `river validate` is included here (#170) so I'm also going to
try a two-phase release by tagging all submodules except for
`./cmd/river` with 0.0.17 first, then modifying that submodule to point
to the new 0.0.17 in other modules, then do a cut and release for it
separately. Kind of a pain, but as discussed at length before, this
seems to be the least bad of a bunch of suboptimal options.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,9 +13,9 @@ require (
 	github.com/jackc/pgx/v5 v5.5.2
 	github.com/jackc/puddle/v2 v2.2.1
 	github.com/oklog/ulid/v2 v2.1.0
-	github.com/riverqueue/river/riverdriver v0.0.15
-	github.com/riverqueue/river/riverdriver/riverdatabasesql v0.0.15
-	github.com/riverqueue/river/riverdriver/riverpgxv5 v0.0.15
+	github.com/riverqueue/river/riverdriver v0.0.17
+	github.com/riverqueue/river/riverdriver/riverdatabasesql v0.0.17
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -13,9 +13,9 @@ require (
 	github.com/jackc/pgx/v5 v5.5.2
 	github.com/jackc/puddle/v2 v2.2.1
 	github.com/oklog/ulid/v2 v2.1.0
-	github.com/riverqueue/river/riverdriver v0.0.15
```
brandur, line 16: See note in PR description, but doh, 0.0.16 got a little botched. My fault.

brandur:
@bgentry Sooo ... see notes above, but the build failure on this one is hard to work around. Mind taking a glance anyway?

bgentry (approved):
:shipit:

brandur:
Thanks!

PR #173, opened 2024-01-23
Also upgrade River CLI to 0.0.17

brandur:
Follow up #172 by also updating the River CLI to version 0.0.17. This
must be done as a separate step because the River CLI module depends on
the main module, and therefore the 0.0.17 tag for the main module must
exist before we can lock to it from the River CLI.

The rough process here was to:

* Remove the `replace` statement added in #170 to demonstrate that the
  new `river validate` command worked.

* Modify `./cmd/river/go.mod`'s River dependency to use v0.0.17.

* Run `go mod tidy` to update `go.sum` and bump the version of the other
  River subdependencies too.

After merging this, I'll add and push one last tag:

    export VERSION=v0.0.17
    git tag cmd/river/$VERSION -m "release cmd/river/$VERSION"
    git push --tags


bgentry (approved):

brandur:
@bgentry Thanks! FWIW, I think we've finally got the release process more or less figured out ...

PR #176, opened 2024-01-26
Remove debug statement in periodic job enqueuer

brandur:
This one's my fault -- left a debug statement in #168 as part of the
periodic job enqueuer. Here, remove it.

Also remove a similar, unused commented out debug statement in
`client_test.go`.

PR #178, opened 2024-01-26
Update `go.mod`s for `v0.0.18` release

brandur:
A small release containing only the fix from #176.


bgentry (approved):

brandur:
Thanks! `v0.0.18` cut and pushed.

PR #179, opened 2024-01-27
Specify specific cancel cause on shutdown and use it to determine error logging

brandur:
Follow up to a discussion [1] in which an additional error log was
causing example tests to fail after uses of `log` were replaced with
`slog` and no longer suppressed.

Here, use `WithCancelCause` to send a specific cancellation error on
shutdown, which can be handled specially for instances where we'd only
want to log an error under unusual circumstances.

[1] https://github.com/riverqueue/river/pull/140#discussion_r1439099623


internal/notifier/notifier.go
```diff
@@ -135,18 +136,7 @@ func (n *Notifier) getConnAndRun(ctx context.Context) {
 
 	conn, err := n.establishConn(ctx)
 	if err != nil {
-		if errors.Is(err, context.Canceled) {
```
brandur, line 138: One thing I wasn't quite able to figure out is why this condition wasn't detecting the cancellation previously. I would've thought that even without a special cancellation error, this still would've been enough to intercept a problem before falling through to logging.

brandur:
@bgentry This one's a follow up from the conversation in https://github.com/riverqueue/river/pull/140#discussion_r1439099623 — roughly what you had in mind? I _think_ this solves the logging problem that was failing example tests before. If I recall correctly, the failures we were seeing before the special patch in #140 would reproduce quite reliably. I reran CI a few times here to try and make a failure happen, but it seems to be fairly stable as written.

client_test.go
```diff
@@ -459,12 +460,7 @@ func Test_Client(t *testing.T) {
 		default:
 		}
 
-		stopCtx, stopCancel := context.WithTimeout(ctx, 5*time.Second)
-		t.Cleanup(stopCancel)
```
brandur, line 463: I took this out because with the addition of `require.ErrorIs` above, it revealed that it occasionally causes an intermittent failure as it kicks in and supersedes the shutdown's cancellation. I don't think this is needed anyway because if shutdown's context cancellation doesn't succeed, I don't think we could expect this one to succeed either. Also, not having keeps the test more succinct.

internal/notifier/notifier.go
bgentry: Hmm, hard to tell without digging into the code, but maybe it's possible pgx doesn't surface this source error correctly? `establishConn` is a simple method that just returns the result of `pgx.ConnectConfig`

bgentry (approved):
Thanks! I've been meaning to take a pass through some places like this and add specific cancellation causes where possible (ever since we started using those APIs).

internal/notifier/notifier.go
brandur: Ah yeah, that's probably it actually. The error being checked is now coming from a different place (context), which would explain why this version works.

brandur:
Thanks!

I'll skip a changelog entry on this one since as written, it's not a user-facing change.

PR #184, opened 2024-01-31
validate queue name on insert, allow hyphen queue name separators

### LINKED ISSUE #183: run validateQueueName on both insert and worker startup?
Hi there! Just started trying out river and really like it so far. However, I ran into this situation which seems like it could be improved upon:

1. I created a job with args with this method `func (MyJobArgs) InsertOpts { return river.InsertOpts{Queue: "my-queue"} }`
2. I deployed my application and successfully inserted a job with that queue value, but realized I'd forgotten to add that queue to my worker config (i did the insert using an insert-only client)
3. I deployed again with an entry in the worker config like `"my-queue": {}`
4. The application failed to start with the error `queue name is invalid, see documentation: "my-queue"`

Personally I like hyphens better than underscores but that's not important 😄 it is surprising though that the `Insert` API doesn't enforce the same validation rules around queue names as the `NewClient` API. Shouldn't the library validate the queue name in both situations using the same logic? I see that there's a `validateJobArgs` that only checks job kinds against registered workers, and only does that if workers are registered (which makes sense), but it feels like it should be fine to also have it always check that the job's configured queue _could_ have a workers attached to it.

(Tangential issue: while the error says "see documentation" I didn't really find anything relevant in the docs; I found the validation rules instead by searching for the error in this repo.)

Happy to submit a PR if you agree but don't have the bandwidth ... or maybe there's a reason to enforce in one situation but not the other?

bgentry:
Validates queue names on `Insert`/`InsertMany` in the same fashion as on client initialization. Also loosens the queue name validation logic to allow hyphen separators in addition to underscore.

Fixes #183.

I thought briefly about implementing a cache to save from re-validating the same queue name many times. We could still do that, but it's a fairly trivial optimization that could be added afterward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Validate queue name on job insertion. [PR #184](https://github.com/riverqueue/rive/pull/184).
+
 ## [0.0.18] - 2024-01-25
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Validate queue name on job insertion. [PR #184](https://github.com/riverqueue/rive/pull/184).
+- Validate queue name on job insertion. Allow queue names with hyphen separators in addition to underscore. [PR #184](https://github.com/riverqueue/river/pull/184).
 
 ## [0.0.18] - 2024-01-25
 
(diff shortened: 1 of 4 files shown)


bgentry:
@brandur lol, we had some tests that used hyphenated queue names on insert. Thoughts on loosening the validation to allow either `-` or `_` separators?

brandur:
> @brandur lol, we had some tests that used hyphenated queue names on insert. Thoughts on loosening the validation to allow either `-` or `_` separators?

I was just about to say from the report in #183 — IMO in this case hyphens vs underscores is an aesthetic choice and we should allow either.

bgentry:
Also changed to allow hyphens in queue names.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Validate queue name on job insertion. Allow queue names with hyphen separators in addition to underscore. [PR #184](https://github.com/riverqueue/rive/pull/184).
```
brandur, line 12: Typo here: "/rive/"

brandur (approved):
lgtm!

CHANGELOG.md
bgentry: thanks, actually clicked it this time to verify 🤦‍♂️

PR #186, opened 2024-02-01
Add JobGet + JobGetTx methods to Client, use within tests

bgentry:
This adds some basic APIs for fetching a single job from the database. I updated all the tests in `client_test.go` to use it, except for some lower level ones that were using dbsqlc types and didn't seem worth converting.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -1310,6 +1310,50 @@ func Test_Client_InsertManyTx(t *testing.T) {
 	})
 }
 
+func Test_Client_JobGet(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+
+	type testBundle struct{}
+
+	setup := func(t *testing.T) (*Client[pgx.Tx], *testBundle) {
+		t.Helper()
+
+		config := newTestConfig(t, nil)
+		client := newTestClient(ctx, t, config)
+
+		return client, &testBundle{}
+	}
+
+	t.Run("FetchesAnExistingJob", func(t *testing.T) {
+		t.Parallel()
+
+		client, _ := setup(t)
+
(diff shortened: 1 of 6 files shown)


bgentry:
Lol, messed up my commits, fixing.

client_test.go
```diff
@@ -1310,6 +1308,50 @@ func Test_Client_InsertManyTx(t *testing.T) {
...
+		require.Equal(t, newJob.ID, job.ID)
+		require.Equal(t, newJob.State, job.State)
+	})
+
+	t.Run("ReturnsErrNotFoundIfJobDoesNotExist", func(t *testing.T) {
+		t.Parallel()
+
+		client, _ := setup(t)
+
+		job, err := client.JobGet(ctx, 999999)
```
brandur, line 1348: Hah, very unlikely, but small chance this fails locally because our test DB sequences don't get reset between runs. Probably will never happen, but I guess zero might be a slighter safer alternative.

brandur (approved):
Looks great!

Should we cut a new release after merging this one? We may not ship anything else interesting for another week or two.

client_test.go
bgentry: Should only be an issue if a row actually exists with this ID, right? And we either truncate after each test or use a txn that gets rolled back.

bgentry:
> Should we cut a new release after merging this one? We may not ship anything else interesting for another week or two.

I don't think so quite yet. I will have another new API up soon (something like `JobRetryNow` to immediately schedule any job that's not currently running) so maybe after that we can cut a new one. Plus I'd like to test both of these further in my UI repo just to be sure.

PR #190, opened 2024-02-05
add JobRetry to Client to enable immediate retry of a job

bgentry:
JobRetry and JobRetryTx will immediately make a job available to be retried again, regardless of the job's current state. The exception is for `running` jobs which will not be altered. All other jobs will be marked as `available` and with a `scheduled_at` no later than `now()` (jobs with a past `scheduled_at` will keep it to hold their place in the queue).

The job's max attempts field is incremented if necessary.


brandur (approved):
Nice! LGTM.

PR #192, opened 2024-02-07
Test against Go 1.22 + 1.21, specify minimum version 1.21

bgentry:
Go 1.22 [has now been released](https://go.dev/doc/go1.22), so we can begin our intended "2 latest major go versions" support policy inline with Go's own policy. We don't foresee any reason for River itself to require anything from Go 1.22 anyway (although the opt-in for loop semantics are going to be ✨).

This PR begins running CI against both 1.22 and 1.21. It also alters the Go version specifier in all our `go.mod` files from `1.21.4` to `1.21.0`, because we don't have any reason to require a specific patch release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,7 @@ jobs:
       matrix:
         postgres-version: [16, 15, 14]
         go-version:
+          - "1.22"
           - "1.21"
       fail-fast: false
     timeout-minutes: 5
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,10 +18,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        postgres-version: [16, 15, 14]
         go-version:
           - "1.22"
           - "1.21"
+        postgres-version: [16, 15, 14]
       fail-fast: false
     timeout-minutes: 5
 
--- a/cmd/river/go.mod
+++ b/cmd/river/go.mod
@@ -2,6 +2,8 @@ module github.com/riverqueue/river/cmd/river
 
 go 1.21.4
 
+toolchain go1.21.6
+
 // replace github.com/riverqueue/river => ../..
 
 // replace github.com/riverqueue/river/riverdriver => ../../riverdriver
--- a/cmd/river/go.mod
+++ b/cmd/river/go.mod
@@ -2,8 +2,6 @@ module github.com/riverqueue/river/cmd/river
 
 go 1.21.4
 
-toolchain go1.21.6
-
 // replace github.com/riverqueue/river => ../..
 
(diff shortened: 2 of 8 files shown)


internal/cmd/update-submodule-versions/main_test.go
```diff
@@ -12,7 +12,7 @@ import (
 
 const sampleGoMod = `module github.com/riverqueue/river
 
-go 1.21.4
```
bgentry, line 15: Oops, didn't mean to change this line, but I'm not sure it matters?

cmd/river/go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/riverqueue/river/cmd/river
 
 go 1.21.4
```
bgentry, line 3: `go mod tidy` wanted to revert some of these submodules back to the `go 1.21.4` specifier. I'm not really sure why and didn't spend much time looking into it, but we can do that if you think it's worthwhile.

.github/workflows/ci.yml
```diff
@@ -18,9 +18,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        postgres-version: [16, 15, 14]
         go-version:
+          - "1.22"
           - "1.21"
+        postgres-version: [16, 15, 14]
```
bgentry, line 24: Flipped these because I find the actions naming better as `(GO_VERSION, PG_VERSION)` than the other way around.

cmd/river/go.mod
brandur: I think it's okay. One thing I've noticed is that to pick up a toolchain from a version, you need to have the patch version in there. i.e. `go 1.22` wont mind 1.22, but `go 1.22.0` will.

cmd/river/go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/riverqueue/river/cmd/river
...
 
+toolchain go1.21.6
```
brandur, line 5: Can we take `toolchain` out? Go should be able to pick it up from the version above.

internal/cmd/update-submodule-versions/main_test.go
brandur: Won't matter as far as the test is concerned.

brandur (approved):
Cool. Looks good!

One thing ia that 2 x 3 = 6 is getting to be a largish matrix. Any chance that's related to why two of the builds here didn't start?

bgentry:
> One thing ia that 2 x 3 = 6 is getting to be a largish matrix. Any chance that's related to why two of the builds here didn't start?

Oh, they all started just fine, but I broke the required status checks by changing the matrix order. I agree it's large... what do you think we should do? We could remove the middle pg version because that seems unnecessary 🤔

PR #193, opened 2024-02-08
Remove pg 15 from CI matrix

bgentry:
Shoot, forgot I had automerge on in #192. This alters the CI matrix to only test the latest Postgres and the one from 2 versions ago, which I feel like gives us sufficient coverage on the "does it still work with 3 major PG versions" question.

Feel free to merge if it looks good.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
         go-version:
           - "1.22"
           - "1.21"
-        postgres-version: [16, 15, 14]
+        postgres-version: [16, 14]
       fail-fast: false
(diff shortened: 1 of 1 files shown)


brandur (approved):
I'm good with or without this one. I know I said a 2x3 matrix is somewhat large, but I guess we could also try it for a bit first to see whether it causes any major issues or us. The logic here makes sense (if it works in 14 and 16 it probably works in 15), but it is also a bit weird to skip one version. I guess it's more sensible if the rationale is more that we always keep it as `[lower_postgres_bound, upper_postgres_bound]`.

bgentry:
Closing for now!

PR #199, opened 2024-02-14
Fix leadership reelection query

### LINKED ISSUE #197: Leadership re-election
Hi, I was reading through the following SQL query for leadership re-election:

```sql
-- name: LeadershipAttemptReelect :execrows
INSERT INTO river_leader(name, leader_id, elected_at, expires_at)
  VALUES (@name::text, @leader_id::text, now(), now() + @ttl::interval)
ON CONFLICT (name)
  DO UPDATE SET
    expires_at = now() + @ttl::interval
  WHERE
    EXCLUDED.leader_id = @leader_id::text;
```

Please correct me if I'm understanding incorrectly, but does the WHERE clause always evaluate to true, which could cause non-leaders to extend the expiry of the current leader?

Should the intent be to compare against the latest leader_id value instead as follows?

```sql
...
WHERE
    river_leader.id = EXCLUDED.leader_id;
```

bgentry:
This adds thorough test cases around leadership election queries, including one which demonstrates the issue reported in #197. Not only does this bug mean that the timestamp would be updated even when a non-leader attempts to re-elect itself, but also that non leader would have been tricked into thinking it was still the leader (due to the rows altered count being >0).

`EXCLUDED` refers specifically to the params we _attempted_ to insert before hitting a conflict, so this condition was always evaluating to true.

Fixes #197. Thank you @joshchoo 🙏🏻

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix a leadership re-election query bug that would cause past leaders to think they were continuing to win elections. [PR #199].
+
 ## [0.0.19] - 2024-02-10
 
 ### Added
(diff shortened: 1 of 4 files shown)


brandur (approved):
Nice! I'll cut a new release with the fix.

PR #204, opened 2024-02-16
add client in log when starting client

TC5027:
related to issue 203


brandur:
Seems reasonable. Thanks!

Issue #203, opened 2024-02-16
Expose client id to provide use of column attempted_by in table river_job


TC5027:
In table river_job there is a column attempted_by which corresponds to which client processed the job. Value inside is the client id. Problem is that this client id is not exposed / can't be accessed making attempted_by column kinda useless, non-informative.

TC5027:
https://github.com/riverqueue/river/pull/204

bgentry:
@TC5027 also see #206 to expose the client ID as a config option and make it programmatically accessible via a new `ID()` method on the `Client` type. Thanks for bringing this up!

PR #206, opened 2024-02-16
expose client ID in config and on client

bgentry:
Inspired by #203, I wanted to make the client IDs configurable by the user and also programmatically accessible (in the event a user decides to rely on auto-generated client IDs). This PR achieves both.

### DIFF
--- a/client.go
+++ b/client.go
@@ -101,6 +101,15 @@ type Config struct {
 	// Defaults to 1 second.
 	FetchPollInterval time.Duration
 
+	// ID is the unique identifier for this client. If not set, a random ULID will
+	// be generated.
+	//
+	// This is used to identify the client in job attempts and for leader election.
+	// This value must be unique across all clients in the same database and
+	// schema and there must not be more than one process running with the same
+	// ID at the same time.
+	ID string
+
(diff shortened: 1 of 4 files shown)


TC5027:
really neat thanks! :)

client.go
```diff
@@ -101,6 +101,15 @@ type Config struct {
 	// Defaults to 1 second.
 	FetchPollInterval time.Duration
 
+	// ID is the unique identifier for this client. If not set, a random ULID will
```
brandur, line 104: What do you think about saying "random identifier" instead of "random ULID" here?

I'm kind of thinking that we shouldn't commit to the ULID algorithm specifically — it'd be nice to get rid of that dependency, and also I think a more human friendly combination of say host name plus start time might be a better default for operations anyway. Not something we need to do right now, but if we changed it, it'd a minor alteration of the contract above.

brandur (approved):

client.go
bgentry: I made that change and also relocated some relevant language that we had stashed on `JobRow.AttemptedBy` so that the full explanation is only given on the `Config.ID` field now.

PR #219, opened 2024-02-19
Elector: Add TTL padding on initial election as well

brandur:
Follows up #217 to add TTL padding on the elector's initial election
attempt in addition to adding it to reelection attempts as was done
previously. We move the TTL calculation to a struct variable for easier
and more foolproof access, and to be able put a comment on it.


bgentry (approved):
:shipit:

brandur:
TY.

PR #221, opened 2024-02-20
Reuse timer on `DebouncedChan` + put in some comments + continous test

brandur:
I wanted to understand a little more clearly how `DebouncedChan` works,
so I went in to look for a few refactoring opportunities. There was a
TODO in the file to potentially reuse the internal timer so it doesn't
need to be reallocated every debounce cycle, so here we implement that.

To make sure that the reused timer is really working we also add a new
"continuous" test that checks that the channel will in fact fire many
times in a row. We also go through and add a few comments and rename a
couple things for clarity, and bring in a `TestMain` wrapper so that the
package tests are checked for goroutine leaks.

### DIFF
--- a/internal/util/chanutil/debounced_chan.go
+++ b/internal/util/chanutil/debounced_chan.go
@@ -11,45 +11,68 @@ import (
 // subsequent calls are delayed until the cooldown period has elapsed and are
 // also coalesced into a single call.
 type DebouncedChan struct {
-	done     <-chan struct{}
 	c        chan struct{}
 	cooldown time.Duration
+	ctxDone  <-chan struct{}
 
-	mu                   sync.Mutex
-	timer                *time.Timer
-	sendWhenTimerExpired bool
+	// mu protects variables in group below
+	mu                 sync.Mutex
+	sendOnTimerExpired bool
+	timer              *time.Timer
+	timerDone          bool
 }
 
 // NewDebouncedChan returns a new DebouncedChan which sends on the channel no
 // more often than the cooldown period.
 func NewDebouncedChan(ctx context.Context, cooldown time.Duration) *DebouncedChan {
 	return &DebouncedChan{
-		done:     ctx.Done(),
+		ctxDone:  ctx.Done(),
 		c:        make(chan struct{}, 1),
 		cooldown: cooldown,
 	}
 }
 
+// C is the debounced channel. Multiple invocations to Call during the cooldown
+// period will deduplicate to a single emission on this channel on the period's
+// leading edge, and one more on the trailing edge.
 func (d *DebouncedChan) C() <-chan struct{} {
 	return d.c
 }
 
+// Call invokes the debounced channel, and is the call which will be debounced.
+// If multiple invocations of this function are made during the cooldown period,
+// they'll be debounced to a single emission on C on the period's leading edge,
+// and one more on the trailing edge.
 func (d *DebouncedChan) Call() {
 	d.mu.Lock()
 	defer d.mu.Unlock()
-	if d.timer != nil {
-		d.sendWhenTimerExpired = true
+
+	// A timer has already been initialized and hasn't already expired. (If it
+	// has expired, we'll reset it below.) Set to signal when it does expire.
+	if d.timer != nil && !d.timerDone {
+		d.sendOnTimerExpired = true
 		return
 	}
 
-	// TODO: The design of this can probably be simplified to not rely on creating
-	// a new timer every time.
-	d.timer = time.NewTimer(d.cooldown)
+	// No timer had been started yet, or the last one running was expired and
+	// will be reset. Send immediately. (i.e. On the leading edge of the
+	// debounce period.)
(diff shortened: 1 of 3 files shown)


internal/util/chanutil/debounced_chan_test.go
```diff
@@ -77,3 +80,63 @@ func TestDebouncedChan_OnlyBuffersOneEvent(t *testing.T) {
...
+	case <-time.After(3 * time.Second):
+		require.FailNow(t, "Timed out waiting for goroutine to finish")
+	}
+
+	// Expect number of signals equal to number of cooldown periods that fit
+	// into our total test time, multiplied by two, because the debounced chan
+	// fires at the beginning and end of a bounce period. +1 for the last period
+	// that fires on the leading edge, but which we don't give time for the
+	// timer to fully expire. (We've chosen numbers so that test time doesn't
+	// divide by cooldown perfectly.)
```
brandur, line 132: @bgentry Something I noticed while looking into the implementation is that for every cooldown period, given relatively constant `Call` signaling throughout, the debounce channel will actually fire twice for every period — once at the beginning, then once at the end as the timer fires. Assuming the signaling continues, it then fires immediately again. So the number of times it fires for N periods is actually 2*N.

It's not too bad of a problem, but OOC was that on purpose? I think there might be an alternate version of this that's more of a "rolling" debounce channel that fires once immediately, but then only fires again once at the end of every cooldown period.

bgentry: Nope, doesn't sound intentional at all. I think the purpose was for it to be a leading edge debounce algorithm (fire immediately, then do not fire again until (a) another signal comes in, and (b) at least `cooldown` has elapsed).

internal/util/chanutil/debounced_chan.go
```diff
@@ -11,45 +11,66 @@ import (
...
-	mu                   sync.Mutex
-	timer                *time.Timer
-	sendWhenTimerExpired bool
+	c                  chan struct{}
+	cooldown           time.Duration
+	ctxDone            <-chan struct{}
+	mu                 sync.Mutex
+	sendOnTimerExpired bool
+	timer              *time.Timer
+	timerDone          bool
```
bgentry, line 20: Do you want to preserve the traditional mutex organization where its declaration precedes all the things that it protects? I think it happens to actually be alphabetical here, but maybe we should keep the line gap and add a comment about `// protects the following` or w/e

internal/util/chanutil/debounced_chan.go
```diff
@@ -11,45 +11,66 @@ import (
...
-		done:     ctx.Done(),
+		ctxDone:  ctx.Done(),
 		c:        make(chan struct{}, 1),
 		cooldown: cooldown,
 	}
 }
 
+// C is the debounced channel. Multiple invocations to Call during the cooldown
+// period will deduplicate to a single emission on this channel on the period's
+// leading edge, and one more on the trailing edge.
```
bgentry, line 35: Yeah, definitely doesn't sound like what I intended here. I wanted it to be leading edge, but if no other calls come in after the initial one nothing should happen, it should just be ready to fire again immediately on the next one.

However if a 2nd call comes in immediately after the 1st one, my thought was that we should still fire immediately a 2nd time at the end of the cooldown. And then repeat, i.e. only fire a 3rd time if yet another call comes in before `cooldown` completes (or whenever it comes in at some later time after `cooldown`).

Hopefully that makes sense? Could easily diagram it if that helps.

bgentry (commented):
Do you think we should try to fix any bugs in this before merging the PR? If not, I can give a full review right away (after I give my brain a rest)

internal/util/chanutil/debounced_chan.go
brandur: Wait, that's a thing??

internal/util/chanutil/debounced_chan.go
brandur: kk, yeah, your explanation here is exactly how I would've intuitively expected it to work too.

brandur:
> Do you think we should try to fix any bugs in this before merging the PR? If not, I can give a full review right away (after I give my brain a rest)

Just since I already wrote the comments/commit message, do you mind if we merge this one mostly as is, and I'll follow up with another PR that tweaks the chan's behavior and shows the correction via changes in the tests?

internal/util/chanutil/debounced_chan.go
bgentry: It's a convention in the Go source code, yeah 😆  there are even linters for it https://github.com/go-critic/go-critic/issues/465

bgentry (approved):

internal/util/chanutil/debounced_chan.go
brandur: Wow, okay that's a total TIL for me! I changed it back, but also added a comment.

brandur:
Thanks! Will follow up with another fix.

PR #222, opened 2024-02-22
DebouncedChan: When channel is firing continously, only fire once per period

brandur:
This one's a follow up to #221. `DebouncedChan` previously had a bit of
a bug where if `Call` was being invoked continuously, it was firing more
often than expected: once for an initial `Call`, and then again when the
timer elapsed.

Here, we modify the implementation so that under a continuous fire
situation, `DebouncedChan` will fire once initially, and then once every
time the timer elapses at the end of each period. This reduces the
number of emits on `C` from 2N+1 to the more expected N+1 (the +1 being
the initial fire).

We accomplish this by entering a loop that waits on the timer when
receiving an initial `Call`, with the loop continuously resetting and
waiting on the timer after each time it fires, sending on `C` each time
the period elapses where a `Call` invocation came in. If a period
elapses without a new `Call` coming in, the timer stops, the loop
returns and `DebouncedChan` returns to its initial state.

### DIFF
--- a/internal/util/chanutil/debounced_chan.go
+++ b/internal/util/chanutil/debounced_chan.go
@@ -34,7 +34,8 @@ func NewDebouncedChan(ctx context.Context, cooldown time.Duration) *DebouncedCha
 
 // C is the debounced channel. Multiple invocations to Call during the cooldown
 // period will deduplicate to a single emission on this channel on the period's
-// leading edge, and one more on the trailing edge.
+// leading edge, and one more on the trailing edge for as many periods as
+// invocations continue to come in.
 func (d *DebouncedChan) C() <-chan struct{} {
 	return d.c
 }
@@ -42,7 +43,10 @@ func (d *DebouncedChan) C() <-chan struct{} {
 // Call invokes the debounced channel, and is the call which will be debounced.
 // If multiple invocations of this function are made during the cooldown period,
 // they'll be debounced to a single emission on C on the period's leading edge,
-// and one more on the trailing edge.
+// and then one fire on the trailing edge of each period for as long as Call
+// continues to be invoked. If a timer period elapses without an invocation on
+// Call, the timer is stopped and behavior resets the next time Call is invoked
+// again.
 func (d *DebouncedChan) Call() {
 	d.mu.Lock()
 	defer d.mu.Unlock()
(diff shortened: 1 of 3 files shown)


internal/util/chanutil/debounced_chan_test.go
```diff
@@ -125,18 +125,13 @@ func TestDebouncedChan_ContinuousOperation(t *testing.T) {
...
-	// +/-4 to allow the channel to be off by two cycles (again, one cycle
-	// signals once at leading edge of the period and once at trailing, so 2
-	// cycles * 2 signals = 4) in either direction. By running at `-count 1000`
-	// or so I can usually reproduce an off-by-one-or-two cycle.
-	expectedNumSignal := int(math.Round(float64(testTime)/float64(cooldown)))*2 + 1
+	// This almost always lands right on the expected number, but allow a delta
+	// of +/-2 to allow the channel to be off by two cycles in either direction.
+	// By running at `-count 1000` or so I can usually reproduce an
+	// off-by-one-or-two cycle.
+	expectedNumSignal := int(math.Round(float64(testTime)/float64(cooldown))) + 1
```
brandur, line 134: This is the key test change here: notice how we drop the `*2` expected invocations.

internal/util/chanutil/debounced_chan.go
```diff
@@ -79,25 +83,52 @@ func (d *DebouncedChan) nonBlockingSendOnC() {
 	}
 }
 
-func (d *DebouncedChan) waitForTimer() {
+// Waits for the timer to be fired, and loops as long as Call invocations comes
```
bgentry, line 86:
```suggestion
// Waits for the timer to be fired, and loops as long as Call invocations come
```

bgentry (approved):
Looks fantastic 🙌  :shipit:

bgentry:
Oh, do you think this is changelog worthy? I would list it as a minor fix in there.

internal/util/chanutil/debounced_chan.go
brandur: Oops, thanks. Fixed!

brandur:
> Oh, do you think this is changelog worthy? I would list it as a minor fix in there.

Yeah I was having a hard time deciding whether to mention it or not. Added an entry now.

brandur:
Thanks!

PR #226, opened 2024-02-25
Fix intermittent test problem in `rivertest` + more transactional use

### LINKED ISSUE #225: Intermittent test failure: `TestRequireInsertedTx/FailsOnInsertOpts`
New test failure on [this run](https://github.com/riverqueue/river/actions/runs/8033452341/job/21943962479):

```
--- FAIL: TestRequireInsertedTx (0.00s)
    --- FAIL: TestRequireInsertedTx/FailsOnInsertOpts (0.23s)
        rivertest_test.go:908: 
                River assertion failure:
                Job with kind 'job2' max attempts 78 not equal to expected 77
            
        rivertest_test.go:908: 
                River assertion failure:
                Job with kind 'job2' priority 2 not equal to expected 3
            
        rivertest_test.go:908: 
                River assertion failure:
                Job with kind 'job2' queue 'another_queue' not equal to expected 'wrong-queue'
            
        rivertest_test.go:908: 
                River assertion failure:
                Job with kind 'job2' scheduled at 2023-10-30T10:[45](https://github.com/riverqueue/river/actions/runs/8033452341/job/21943962479#step:8:46):23.000123Z not equal to expected 2023-10-30T10:48:46.000246Z
            
        rivertest_test.go:908: 
                River assertion failure:
                Job with kind 'job2' state 'available' not equal to expected 'cancelled'
            
        rivertest_test.go:374: 
            	Error Trace:	/home/runner/work/river/river/rivertest/rivertest_test.go:374
            	Error:      	Not equal: 
            	            	expected: "\n    River assertion failure:\n    Job with kind 'job2' state 'scheduled' not equal to expected 'cancelled'\n\n"
            	            	actual  : "\n    River assertion failure:\n    Job with kind 'job2' state 'available' not equal to expected 'cancelled'\n\n"
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -2,3 +2,3 @@
            	            	     River assertion failure:
            	            	-    Job with kind 'job2' state 'scheduled' not equal to expected 'cancelled'
            	            	+    Job with kind 'job2' state 'available' not equal to expected 'cancelled'
            	            	 
            	Test:       	TestRequireInsertedTx/FailsOnInsertOpts
FAIL
FAIL	github.com/riverqueue/river/rivertest	1.2[57](https://github.com/riverqueue/river/actions/runs/8033452341/job/21943962479#step:8:58)s
```

Just reporting this for now because I'm on a plane. You may also be more familiar with these tests and able to diagnose the failure quickly @brandur.

brandur:
This one's aimed at fixing #225, in which I think what was happening is
that because we were inserting a test job on a non-transactional pool
with a state of `scheduled`, in some cases the queue maintainer's
scheduler service was actually changing it back to `available` before
the test case would check it with a `RequireInserted`.

As I was moving through the test suite, I realized that we could be
making a lot more use of `InsertTx` variants pretty much everywhere
since we were mostly testing the transactional variables of
`RequireInserted`, so that I changed those over.

Then I realize that even when we were testing the non-transactional
variants, we didn't actually need a started client to do so, so I took
that code out. Then I realized that since we're not starting the client,
we didn't actually need to specify `Queues` or `Workers` (I believe the
idea of an insert-only came about only after I wrote these tests
originally). And if we didn't need to specify `Workers`, then we didn't
even need the worker structs, so I got rid of those too.

Overall, I think we simplify things quite a bit, and because the client
doesn't need to be started, probably reduces the possibility of future
intermittency problems like we observed in #225.

Fixes #225.

### DIFF
--- a/rivertest/rivertest_test.go
+++ b/rivertest/rivertest_test.go
@@ -25,23 +25,11 @@ type Job1Args struct {
 
 func (Job1Args) Kind() string { return "job1" }
 
-type Job1Worker struct {
-	river.WorkerDefaults[Job1Args]
-}
-
-func (w *Job1Worker) Work(ctx context.Context, job *river.Job[Job1Args]) error { return nil }
-
 type Job2Args struct {
 	Int int `json:"int"`
 }
 
 func (Job2Args) Kind() string { return "job2" }
-
-type Job2Worker struct {
-	river.WorkerDefaults[Job2Args]
-}
-
-func (w *Job2Worker) Work(ctx context.Context, job *river.Job[Job2Args]) error { return nil }
 
 // The tests for this function are quite minimal because it uses the same
 // implementation as the `*Tx` variant, so most of the test happens below.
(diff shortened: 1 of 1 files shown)


bgentry:
> Then I realize that even when we were testing the non-transactional
> variants, we didn't actually need a started client to do so, so I took
> that code out. Then I realized that since we're not starting the client,
> we didn't actually need to specify `Queues` or `Workers` (I believe the
> idea of an insert-only came about only after I wrote these tests
> originally). And if we didn't need to specify `Workers`, then we didn't
> even need the worker structs, so I got rid of those too.

Yep, all of this correct. Primarily the fix here is to not run maintenance processes for tests that don't need or want them. In the course of investigating #213, I've also realized we have the scheduler running frequently in most tests, which is another thing we don't need most of the time and probably adds up to a fair bit of overhead. I'm sure there are other similar issues & inefficiencies throughout the tests to clear out.

bgentry (approved):
Great fixes, thank you!

brandur:
Great, thanks!

PR #227, opened 2024-02-25
Improve ListJob flexibility - Allow for querying based on a list of states and kinds

thatjos:
This modifies the JobListParams such that the user can pass in any number of job states or job kinds to filter against. Before this, the JobList API only exposed a way to grab jobs with a single type of state at a time, and filtering based on job kind was not possible. 

This also allows the user to specify a specific timestamp field to sort by in case multiple job states are requested.

Happy to update / improve based on your feedback! Happy River user so far.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -1407,12 +1407,12 @@ func Test_Client_JobList(t *testing.T) {
 		job2 := insertJob(ctx, client.driver.GetDBPool(), insertJobParams{State: dbsqlc.JobStateAvailable})
 		job3 := insertJob(ctx, client.driver.GetDBPool(), insertJobParams{State: dbsqlc.JobStateRunning})
 
-		jobs, err := client.JobList(ctx, NewJobListParams().State(JobStateAvailable))
+		jobs, err := client.JobList(ctx, NewJobListParams().States(JobStateAvailable))
 		require.NoError(t, err)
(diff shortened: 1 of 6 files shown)


thatjos:
Ah just noticed you added a bunch of this already, let me revisit my work in light of that!

brandur:
Thanks for the patch.

> Happy River user so far.

Excellent! That's great to hear.

> Ah just noticed you added a bunch of this already, let me revisit my work in light of that!

Yeah. Oof, your timing on this one is quite unlucky — I just landed a big refactor today that moved a lot of things around, and also randomly added filtering loads by kind after I happened to notice it was missing.

So much has shifted that it might actually be easier to start your patch over, using this one as a reference, rather than try to rebase and resolve the conflicts (which I'd expect to be somewhat painful). One note to help on that is that previously the list logic was broken apart, with about half in `dbadapter` and half in `dblist`, now it's basically all in `dblist`.

@bgentry There's a minor breaking change here that'd allow filtering by multiple states (so `State(...)` would become `States(...)`), which is probably what it should've been before. We might have gotten it early enough that few people are depending on the list API so far. Thoughts on the change?

thatjos:
I'm working on a new PR now yeah, should have something up shortly. Thanks so much for the response!

thatjos:
Closing in favor of https://github.com/riverqueue/river/pull/227

PR #228, opened 2024-02-25
Consolidate on `.yaml` extension for YAML files

brandur:
This one is extremely minor, but we use the `.yml` extension for GitHub
Actions and golangci-lint, and the `.yaml` extension for our sqlc
configurations. I find that I occasionally type the full "yaml" when
doing a file quick search in places like VSCode, and it's a bit better
to have the full `.yaml` extension for that because typing "yml" will
find either `.yml` or `.yaml`, but typing "yaml" finds only the latter.

Both GitHub Actions and golangci-lint [1] are agnostic to the file
extension in use.

[1] https://golangci-lint.run/usage/configuration/


brandur:
@bgentry The only good news about these frequent intermittent failures is that they're annoying enough that it's going to make us get pretty serious about fixing them fast, lol.

Tiny change here — okay with you?

bgentry (approved):

brandur:
Ty.

PR #229, opened 2024-02-25
Drop sqlc overrides for `JobState`

brandur:
This is another small one, but I noticed while doing driver work that
since we put a layer of abstraction between sqlc and any public-facing
APIs, the `JobState*` state constants in sqlc are not really used
anymore. Even the `JobState` type is used to infrequently that golfing
"River" off the front also does very little.

Here, I propose that we just get rid of the sqlc overrides. They're not
a big deal, but not having them means less sqlc configuration that needs
to be shared around between YAML files, and less opportunity for error
there.

### DIFF
--- a/riverdriver/riverdatabasesql/internal/dbsqlc/sqlc.yaml
+++ b/riverdriver/riverdatabasesql/internal/dbsqlc/sqlc.yaml
@@ -22,13 +22,6 @@ sql:
 
         rename:
           river_job_state: "JobState"
-          river_job_state_available: "JobStateAvailable"
-          river_job_state_cancelled: "JobStateCancelled"
-          river_job_state_completed: "JobStateCompleted"
-          river_job_state_discarded: "JobStateDiscarded"
-          river_job_state_retryable: "JobStateRetryable"
-          river_job_state_running: "JobStateRunning"
-          river_job_state_scheduled: "JobStateScheduled"
           ttl: "TTL"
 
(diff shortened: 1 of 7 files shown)


brandur:
@bgentry Another small one. Seem okay?

riverdriver/riverdatabasesql/internal/dbsqlc/sqlc.yaml
```diff
@@ -22,13 +22,6 @@ sql:
 
         rename:
           river_job_state: "JobState"
-          river_job_state_available: "JobStateAvailable"
-          river_job_state_cancelled: "JobStateCancelled"
-          river_job_state_completed: "JobStateCompleted"
-          river_job_state_discarded: "JobStateDiscarded"
-          river_job_state_retryable: "JobStateRetryable"
-          river_job_state_running: "JobStateRunning"
-          river_job_state_scheduled: "JobStateScheduled"
```
bgentry, line 31: one less thing to have to fix when I get that pending state added soon.

bgentry (approved):

riverdriver/riverdatabasesql/internal/dbsqlc/sqlc.yaml
brandur: Yep, totally. Should keep future changes of that sort a bit easier.

PR #230, opened 2024-02-25
Emit pointers for struct parameters in sqlc as well as results

brandur:
By default, sqlc emits non-pointer structs for both parameters and
results. Since pretty much the beginning we've had this configuration to
emit pointers instead of structs for results:

    emit_result_struct_pointers: true

Here, I propose that we add a similar configuration that also makes
parameters pointers:

    emit_params_struct_pointers: true

It's not a big deal either way, but this fits a little better with the
`riverdriver` interface, which uses pointers for everything already, it
also makes it so we can do a direct pointer conversion in a couple
places, probably leading to a very, very nominal efficiency gain. e.g.:

    err := e.queries.JobRescueMany(ctx, e.dbtx, (*dbsqlc.JobRescueManyParams)(params))

### DIFF
--- a/riverdriver/riverdatabasesql/internal/dbsqlc/pg_misc.sql.go
+++ b/riverdriver/riverdatabasesql/internal/dbsqlc/pg_misc.sql.go
@@ -27,7 +27,7 @@ type PGNotifyParams struct {
 	Payload string
 }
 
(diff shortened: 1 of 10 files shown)


brandur:
@bgentry Yet another very nominal change.

bgentry (approved):

brandur:
ty.

PR #232, opened 2024-02-25
Move `JobCompleteTx` to its own `.go` file

brandur:
Another small one, but I always have a hard time remembering where
`JobCompleteTx` lives because it doesn't feel intuitive to me that it
should belong in `job.go` with the `Job` definition.

Here, break it out into its own separate file with a clear-as-day name
associated with it. Also give it a basic test suite that pairs with the
new file. This'll be somewhat duplicative with some cases in
`client_test.go`, but that shouldn't be harmful.

### DIFF
--- a/job.go
+++ b/job.go
@@ -1,13 +1,17 @@
 package river
 
 import (
-	"context"
-	"encoding/json"
-	"errors"
-	"time"
-
-	"github.com/riverqueue/river/riverdriver"
 	"github.com/riverqueue/river/rivertype"
+)
+
+const (
+	JobStateAvailable = rivertype.JobStateAvailable
+	JobStateCancelled = rivertype.JobStateCancelled
+	JobStateCompleted = rivertype.JobStateCompleted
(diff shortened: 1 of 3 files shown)


job.go
```diff
@@ -1,15 +1,19 @@
...
 
+const (
+	JobStateAvailable = rivertype.JobStateAvailable
+	JobStateCancelled = rivertype.JobStateCancelled
+	JobStateCompleted = rivertype.JobStateCompleted
+	JobStateDiscarded = rivertype.JobStateDiscarded
+	JobStateRetryable = rivertype.JobStateRetryable
+	JobStateRunning   = rivertype.JobStateRunning
+	JobStateScheduled = rivertype.JobStateScheduled
+)
```
brandur, line 15: Also hoisted these job state definitions to the top of the file. Feels a little weird that they lived at the bottom.

brandur:
@bgentry Okay with this?

bgentry (approved):

PR #233, opened 2024-02-25
`Rescuer` -> `JobRescuer` + `Scheduler` -> `JobScheduler`

brandur:
Another small one, but it's triggering my OCD in that for services
inside the queue maintainer, the job cleaner is called `JobCleaner`, the
periodic enqueuer is called `PeriodicJobEnqueuer`, but the rescuer is
only `Rescuer` (without a "job" prefix) and scheduler is `Scheduler`.

Here, rename the rescuer and scheduler so they pick up a prefix and
become `JobRescuer` and `JobScheduler`. I think this is nominally
beneficial anyway because it helps telegraph that these are services
that are specific to jobs (unlike say `Reindexer` for example, which is
not related to jobs specifically, and therefore unprefixed).

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -1823,6 +1823,131 @@ func Test_Client_Maintenance(t *testing.T) {
 		require.NotErrorIs(t, err, ErrNotFound) // still there
 	})
 
+	t.Run("JobRescuer", func(t *testing.T) {
+		t.Parallel()
+
+		dbPool := riverinternaltest.TestDB(ctx, t)
+
+		config := newTestConfig(t, nil)
+		config.RescueStuckJobsAfter = 5 * time.Minute
+
+		client := newTestClient(t, dbPool, config)
+		exec := client.driver.GetExecutor()
+
+		now := time.Now()
+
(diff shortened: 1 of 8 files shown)


client_test.go
```diff
@@ -1823,72 +1823,7 @@ func Test_Client_Maintenance(t *testing.T) {
...
-
-		client.testSignals.electedLeader.WaitOrTimeout()
-		svc := maintenance.GetService[*maintenance.Reindexer](client.queueMaintainer)
-		// There are two indexes to reindex by default:
-		svc.TestSignals.Reindexed.WaitOrTimeout()
-		svc.TestSignals.Reindexed.WaitOrTimeout()
-	})
-
-	t.Run("Rescuer", func(t *testing.T) {
+	t.Run("JobRescuer", func(t *testing.T) {
```
brandur, line 1826: No substantial changes in this file despite what the diff suggests — I just hoisted the tests for the rescuer and scheduler up so that they're sorted in the same way as in the client and alphabetically.

bgentry (approved):

Issue #205, opened 2024-02-16
Worker Coordinator fails with PgBouncer due to unsupported statement_timeout


mfrister:
Thanks for the extensive documentation on using river with PgBouncer! 

I’m still having a small issue with river and PgBouncer - when the worker tries to connect, it fails (wrapped for better readability):

    river client start: error making initial connection to database:
    failed to connect to `host=... user=... database=...`:
    server error (FATAL: unsupported startup parameter: statement_timeout (SQLSTATE 08P01))

PgBouncer doesn’t seem to support the `statement_timeout` parameter. Disabling the [statement_timeout in the Notifier](https://github.com/riverqueue/river/blob/896fd0b93ba4c334e3d7e1d35d1b03688e5d3830/internal/notifier/notifier.go#L69) by removing the line makes the worker run. The comment above this line indicates that the timeout is necessary, though:

```
// Rely on an overall statement timeout instead of setting identical context timeouts on every query:
```

Side Note: PgBouncer does seem to have an option to ignore certain startup parameters, but I’m a bit wary of doing so, given that ignoring the statement timeout might have hard-to-notice side effects for other applications. I’d rather have the applications fail while connecting and adapt them however necessary.

My questions:

1. Do you expect river to be negatively impacted by removing the statement timeout? I interpret the comment mentioned above as the statement timeout being required for reliable operation, although I'm wondering whether context-based timeouts would be necessary anyway to properly deal with network issues that would prevent river from receiving a response from Postgres. I found at least one context-based timeout in the Notifier, I'm not sure that's enough to cover all cases, though.
2. If river is able to work reliably without the statement timeout, would you be open to making the timeout optional in some way or replace it with context-based timeouts where necessary?

I’m happy to submit a PR if you think a change in river makes sense.

brandur:
Hi @mfrister, thanks for the report -- we'd intended to do more testing with PgBouncer to make sure everything looked good there, but hadn't gotten around to it yet.

One question before I start: what pooling mode are you using PgBouncer in?

I didn't test it, but reading through the features docs in PgBouncer [1], `SET` should be supported in session pooling mode. And if you _aren't_ using session pooling mode, even if we take out the `statement_timeout`, the client's probably going to fail on something anyway. You can run an insert-only client in other other PgBouncer pooling modes, but when the client is trying to act as work coordinator, it's going to need session pooling or bust.

[...]

Issue #171, opened 2024-01-22
Enqueue problem (PostgreSQL 22P02)


lexor:
Have been playing around with River and I get an error when adding a job to the queue. I define 3 string arg as job args. 

The error I received is as follows: `error getting unique job: ERROR: invalid input syntax for type json (SQLSTATE 22P02)`

```go
type ExampleJobArgs struct {
	Email  string `json:"email"`
	UserID string `json:"user_id"`
}

func (ExampleJobArgs) Kind() string {
	return "activation_email"
}

func (ExampleJobArgs) InsertOpts() river.InsertOpts {
	return river.InsertOpts{
		UniqueOpts: river.UniqueOpts{
			ByArgs:   true,
			ByPeriod: 4 * time.Hour,
		},
	}
}

// ...

```

```go
type api struct { queue *river.Client[pgx.Tx] }

queue, err := river.NewClient(riverpgxv5.New(postgres), &river.Config{})
if err != nil {
	panic(err)
}

// ...

func (a *api) enqueueHandler(w http.ResponseWriter, r *http.Request) {
	ctx, cancel := context.WithCancel(r.Context())
	defer cancel()

	if _, err = a.queue.Insert(ctx, worker.ExampleJobArgs{
		Email:    "foo@bar.tld",
		UserID:  "1",
	}, nil); err != nil {
		// ...
	}

	// ...
}
```

My periodic jobs works correctly, but I only get this error when enqueueing from API.

brandur:
Odd, so the unique query isn't do anything too exotic. Basically just encoding JSON. Here it is:

``` go
			if params.UniqueByArgs {
				advisoryLockHash.Write([]byte("&args="))
				advisoryLockHash.Write(params.EncodedArgs)

				getParams.Args = params.EncodedArgs
				getParams.ByArgs = true
			}
```

Do you want to try doing a `json.Marshal` on your input `ActivationEmailArgs`, print in, and see if anything in there looks odd?

If not, maybe try removing each of the fields one at a time, and see if any one of them seems to be the offender? (It's probably `email`? But just a guess.)

lexor:
@brandur I tried already removed all args but facing same error.

```
_, err := a.queue.Insert(ctx, worker.ExampleJobArgs{}, nil)

...

type ExampleJobArgs struct {}

func (ExampleJobArgs) Kind() string {
	return "example"
}

func (ExampleJobArgs) InsertOpts() river.InsertOpts {
	return river.InsertOpts{
[...]

PR #234, opened 2024-02-25
Proof of concept: River operating with pgx simple protocol

brandur:
We had an issue opened a little while back (#171) wherein it turned out
that River couldn't operate with pgx's simple protocol turned on, which
appeared to be because the simple protocol doesn't know anything about a
target Postgres tuple's makeup and therefore had to guess how to encode
parameters based on input types, which was leading `[]byte` intended for
`jsonb`s to be binary encoded, which isn't allowed.

We closed that issue after I pointed out that PgBouncer now supports
prepared statement pooling, and therefore simple protocol is no longer
necessary.

We recently got another comment [1] where someone had tried to use River
with Supavisor on Supabase, which also does not supported prepared
statements, leading to lots of errors. Normally, pgx's simple protocol
could be used to address this, but as we saw in #171, River doesn't
support that.

Here, a proof of concept demonstrating how River could support simple
protocol by modeling `jsonb` fields in sqlc as strings, which prevents
pgx from encoding to binary with simple procotol, and we get to a point
where all driver tests are passing.

I'm not sure if we should action on this one or not. Having to fall back
to `string` fields a bit gross, although I don't *think* carries
significant performance penalty because it's basically the same
underlying structure as far as Go is concerned.

[1] https://github.com/riverqueue/river/issues/205#issuecomment-1963001669

### DIFF
--- a/internal/riverinternaltest/riverinternaltest.go
+++ b/internal/riverinternaltest/riverinternaltest.go
@@ -73,6 +73,7 @@ func DatabaseConfig(databaseName string) *pgxpool.Config {
 	// are unlikely to succeed even with more time:
 	config.ConnConfig.ConnectTimeout = 2 * time.Second
 	config.ConnConfig.RuntimeParams["timezone"] = "UTC"
+	config.ConnConfig.DefaultQueryExecMode = pgx.QueryExecModeSimpleProtocol
 	return config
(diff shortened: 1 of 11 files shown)


brandur:
Closing this out for now.

PR #235, opened 2024-02-25
Don't disable sleep by default from `riverinternaltest.BaseServiceArchetype`

brandur:
The `riverinternaltest.BaseServiceArchetype` test helper has up until
now return an archetype with sleep disabled, with the idea that
deactivating it by default derisked the possibility that tests would
accidentally introduce sleep invocations that'd make them slow.

(Note this only works for invocations of `BaseService.CancellableSleep`
which a lot of River doesn't use.)

This made more sense when the main use of the archetype was in in the
queue maintenance services, which all start up with a short jittered
sleep so that they don't all go for the database at the same time.

But since then, the use of archetype has become more widespread, and
disabling sleep probably doesn't make sense in most situations,
especially when testing at the client level. It'd have the effect of
making all the queue maintenance services run more aggressively than
they would by default, which isn't desirable, and may even lead to
contention problems.

Here, add a `WithSleepDisabled()` helper to archetype that lets sleep be
disabled easily in tests that want to do that (and add invocations of it
to each of the queue maintainer tests), but keep sleep on everywhere
else.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -125,7 +125,6 @@ func newTestConfig(t *testing.T, callback callbackFunc) *Config {
 		Logger:            riverinternaltest.Logger(t),
 		Queues:            map[string]QueueConfig{QueueDefault: {MaxWorkers: 50}},
 		Workers:           workers,
-		disableSleep:      true,
 		schedulerInterval: riverinternaltest.SchedulerShortInterval,
 	}
 }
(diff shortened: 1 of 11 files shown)


bgentry (approved):

brandur:
Thanks!

This might not be the perfect solution — it may be that a lot of sleeps should not be cancellable, but rather only "advisory" sleeps that are used to prevent thundering herd and that kind of thing. I'll have to find a good way to make a distinction between those two cases though, so this is an improvement for now.

PR #236, opened 2024-02-26
Change JobList ergnomics - add filtering by states, return cursor from JobList function

thatjos:
This supersedes https://github.com/riverqueue/river/pull/227. It adds:

* Ability to filter by any number of states. By default, returns all states.
* Ability to choose a specific sort field. By default, uses created_at
* Updates JobListCursor to include a sort field
* Updates JobList and JobListTx to create the cursor for users and return it with as a separate return value


client.go
```diff
@@ -1360,20 +1360,27 @@ func validateQueueName(queueName string) error {
 //	if err != nil {
 //		// handle error
 //	}
-func (c *Client[TTx]) JobList(ctx context.Context, params *JobListParams) ([]*rivertype.JobRow, error) {
+func (c *Client[TTx]) JobList(ctx context.Context, params *JobListParams) ([]*rivertype.JobRow, *JobListCursor, error) {
```
brandur, line 1363: Agreed that the code needed to get the next cursor is a little overly verbose at the moment, and furthermore you have to make a `len` check on jobs to make sure you're not indexing a 0-length slice.

``` go
jobs, err := client.JobList(ctx, NewJobListParams().After(JobListCursorFromJob(jobs[len(jobs)-1])))
```

If we were going to change this API though, I think I'd be tempted to make it more like a result struct, which is a bit more conventional, and a little more futureproof. e.g.

``` go
type JobListResult struct {
    Jobs []*rivertype.JobRow
    NextCursor *JobListCursor
```

Thoughts?

job_list_params.go
```diff
@@ -15,19 +15,36 @@ import (
...
+		if job.AttemptedAt != nil {
+			time = *job.AttemptedAt
+		}
+	case JobListOrderByFinalizedAt:
+		if job.FinalizedAt != nil {
+			time = *job.FinalizedAt
+		}
+	case JobListOrderByScheduledAt:
+		time = job.ScheduledAt
+	}
```
brandur, line 41: OOC, wasn't this `switch` in `jobListTimeValue` before (and isn't `jobListTimeValue` still in this file?)? Is moving it up just to remove a layer of indirection?

brandur (commented):
Thanks @thatjos! Looking very good.

Did a preliminary review here, but @bgentry, since you wrote the bulk of this code you should take the final review.

Notably: we've got a couple breaking changes in here. I'm on the fence — we're trying not to do that, but on the other hand, the job list API is still quite new, so maybe we could get away with it. What do you think?

brandur:
Oh and @thatjos, you may want to pull down golangci-lint and get that running locally. A few lint errors here, and the local tool makes it much faster to find/fix them.

Depending on your OS flavor, something like:

``` sh
$ brew install golangci-lint
$ golangci-lint run --fix
```

thatjos:
Agreed on all points, I'll clean it up!

job_list_params.go
thatjos: Yeah, unfortunately the cursor is tied to the sorting mechanism that was used for the initial query. If the sorting mechanism was the generic 'time' one (meaning - inferred through the state that was queried), we'd have to do the same thing here. If it wasn't, and the user sorted by a specific timestamp field, that needs become a part of the cursor. The logic is slightly different between this function and `jobListTimeValue`, and I couldn't think of a way to combine the functions. Happy to edit if you have thoughts!

brandur:
@thatjos Thanks! Looking good.

I have a few nit-picky comments, but will wait until @bgentry's able to do a higher level review first so that you don't end up making changes that weren't needed.

bgentry:
@thatjos sorry for the delay on this, I've had to prioritize fixing the memory leaks these past few days. I will get to it soon!

thatjos:
> @thatjos sorry for the delay on this, I've had to prioritize fixing the memory leaks these past few days. I will get to it soon!

No worries, thanks so much for the great work

client.go
```diff
@@ -1351,6 +1351,13 @@ func validateQueueName(queueName string) error {
 	return nil
 }
 
+// JobListResult is the result of a job list operation. It contains a list of
+// jobs and a cursor for fetching the next page of results.
+type JobListResult struct {
+	Jobs   []*rivertype.JobRow
+	Cursor *JobListCursor
```
bgentry, line 1358: Not sure it makes sense to output a single `Cursor` field here. There could be a cursor for _any_ of the entries in the `Jobs` list, so with this name it's not clear if it's the start cursor or end cursor.

I doubt there's much use case for the start cursor, and people can always use the `JobListCursorFromJob` function to create one if they want. Maybe let's just name this `LastCursor` to be unambiguous?

job_list_params.go
```diff
@@ -90,12 +113,20 @@ const (
...
+	// time field used will vary by the first specified job state.
+	JobListOrderByTime JobListOrderByField = "time"
+	// JobListOrderByCreatedAt specifies that the sort should be by created_at.
+	JobListOrderByCreatedAt JobListOrderByField = "created_at"
+	// JobListOrderByScheduledAt specifies that the sort should be by scheduled_at.
+	JobListOrderByScheduledAt JobListOrderByField = "scheduled_at"
+	// JobListOrderByAttemptedAt specifies that the sort should be by attempted_at.
+	JobListOrderByAttemptedAt JobListOrderByField = "attempted_at"
+	// JobListOrderByFinalizedAt specifies that the sort should be by finalized_at.
+	JobListOrderByFinalizedAt JobListOrderByField = "finalized_at"
```
bgentry, line 129: Can we keep these sorted alphabetically?

bgentry (commented):
LGTM aside from a couple minor nits. We'll need a breaking changelog entry for this as well and will want to bump the minor prerelease version to indicate the break.

Thanks for your patience!

thatjos:
Thanks so much @bgentry and apologies for the delay! Fixed up those things and wrote an unreleased changelog entry. Would you be able to help me cut a new release whenever's convenient for you?

brandur:
Added "breaking change" label.

Sorry for the delay here @thatjos. Have spoken about this one with Blake, and we'll get back to you soon. We're going to ship a couple small breaking changes as part of a single release to make sure that we're not making upgrade headaches a regular feature. See also #292.

brandur:
@thatjos Blake ended up pulling your commits into another PR in #304 which we merged in with a few tweaks. We should have a release cut sometime this weekend. Closing this PR out, but thanks for the contribution.

PR #237, opened 2024-02-28
Implement `rivertest.RequireNotInserted` test helper (inverse of `RequireInserted`)

brandur:
Here, introduce a new test helper for `rivertest.RequireNotInserted`,
which fails in cases where a job matching in the input kind and
(optional) properties _was_ inserted rather than was not. Its purpose is
to invert the checks made by the existing `RequireInserted`.

Unlike `RequireInserted`, I _didn't_ implement a batch version of the
helper like `RequireNotInsertedMany`. I was going to do it, but after
thinking about it for a while, I don't think it really makes sense.
`RequireInsertedMany` is useful because it allows you to verify a
sequence of job inserts that occurred in a particular order with
particular options. There's no sequence for jobs that weren't inserted,
so the only thing a `RequireNotInsertedMany` would do is verify that
many different job kinds were all not inserted as part of one operation,
which doesn't really seem that useful. I figure if there's demand we can
think about adding a batch helper, but it may be better to not do so
prospectively to keep the API smaller, and avoiding adding a function
that doesn't end up getting any use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `RequireNotInserted` test helper (in addition to the existing `RequireInserted`) that verifies that a job with matching conditions was _not_ inserted. [PR #237](https://github.com/riverqueue/river/pull/237).
+
 ## [0.5.0] - 2024-05-03
 
 ⚠️ Version 0.5.0 contains a new database migration, version 4. This migration is backward compatible with any River installation running the v3 migration. Be sure to run the v4 migration prior to deploying the code from this release.
--- a/rivertest/rivertest.go
+++ b/rivertest/rivertest.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"slices"
+	"strings"
 	"testing"
 	"time"
 
@@ -29,8 +30,18 @@ type testingT interface {
 	Logf(format string, args ...any)
 }
 
-// Options for RequireInserted or RequireManyInserted including expectations for
-// various queuing properties that stem from InsertOpts.
+// Options for RequireInserted functions including expectations for various
+// queuing properties that stem from InsertOpts.
+//
+// Multiple properties set on this struct increase the specifity on a job to
+// match, acting like an AND condition on each.
+//
+// In the case of RequireInserted or RequireInsertedMany, if multiple properties
+// are set, a job must match all of them to be considered a successful match.
+//
+// In the case of RequireNotInserted, if multiple properties are set, a test
+// failure is triggered only if all match. If any one of them was different, an
+// inserted job isn't considered a match, and RequireNotInserted succeeds.
 type RequireInsertedOpts struct {
 	// MaxAttempts is the expected maximum number of total attempts for the
 	// inserted job.
@@ -78,7 +89,7 @@ type RequireInsertedOpts struct {
 //
 // A RequireInsertedOpts struct can be provided as the last argument, and if it is,
 // its properties (e.g. max attempts, priority, queue name) will act as required
-// assertions in the inserted job row. UniqueOpts is ignored.
+// assertions in the inserted job row.
 //
 // The assertion will fail if more than one job of the given kind was found
 // because at that point the job to return is ambiguous. Use RequireManyInserted
@@ -108,7 +119,7 @@ func requireInserted[TDriver riverdriver.Driver[TTx], TTx any, TArgs river.JobAr
 //
 // A RequireInsertedOpts struct can be provided as the last argument, and if it is,
 // its properties (e.g. max attempts, priority, queue name) will act as required
-// assertions in the inserted job row. UniqueOpts is ignored.
+// assertions in the inserted job row.
 //
 // The assertion will fail if more than one job of the given kind was found
 // because at that point the job to return is ambiguous. Use RequireManyInserted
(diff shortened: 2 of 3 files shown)


rivertest/rivertest.go
```diff
@@ -157,14 +166,117 @@ func requireInsertedErr[TDriver riverdriver.Driver[TTx], TTx any, TArgs river.Jo
...
+//
+// A RequireInsertedOpts struct can be provided as the last argument, and if it
+// is, its properties (e.g. max attempts, priority, queue name) will act as
+// requirements on a found row. If any fields are set, then the test will fail
+// if a job is found that matches any of them (unlike options to RequireInserted
+// which behave like an AND, these are an OR). So for example, if options
+// specify `Priority: 3`, and a joke of the same kind was inserted, but it was
+// `Priority: 2`, RequireNotInserted will not fail. If the inserted job was
+// `Priority: 2` (therefore matching the job), RequireNotInserted does fail.
+func RequireNotInserted[TDriver riverdriver.Driver[TTx], TTx any, TArgs river.JobArgs](ctx context.Context, tb testing.TB, driver TDriver, expectedJob TArgs, opts *RequireInsertedOpts) {
```
brandur, line 193: Okay, so I went back and forth a lot on what `opts` this function should accept.

You can see here that it takes `RequireInsertedOpts`, which is the same as `RequireInserted`. Initially, this felt a little wrong to me, but after implementing it as a different struct, having a totally duplicated set of fields on a different struct just felt wasteful.

I then thought about hacking it a bit with something like:

``` go
type RequireNotInsertedOpts RequireInsertedOpts
```

Which would let us futureproof a bit by changing the type definition to a full struct in the future in case we needed to. But it has the unfortunate downside of being worse for IDE mouseover definitions of the struct, because the properties of `RequireInsertedOpts` aren't directly visible.

Lastly, I considered just not giving this function an `opts` at all. The purpose of the opts is so you could do something like:

``` go
RequireNotInserted(..., &RequireInsertedOpts{Queue: "high-priority"})
```

To verify that a job of that type _did not_ go the high property queue. So if a job the kind went to the default queue, it's okay, but it fails if it went to high priority.

However, one could also imagine asserting on this affirmatively instead by inverting. Something like:

``` go
RequireInserted(..., &RequireInsertedOpts{Queue: "default"})
```

Which might be clearer than the use of `RequireNotInserted` anyway.

That said, if we decide not to give `RequireNotInserted` options, then realize later that we need them for something, it's a breaking change to add them in, which would be nice to avoid.

Any immediate thoughts on all that?

rivertest/rivertest.go
```diff
@@ -29,8 +29,17 @@ type testingT interface {
...
+// queuing properties that stem from InsertOpts.
+//
+// When used with RequiredInserted or RequireInsertedMany, multiple properties
+// set on this struct increase the specifity on a job to match, acting like an
+// AND condition on each. So if multiple properties are set, a job must match
+// all of them to be considered a successful match.
+//
+// When used with RequireNotInserted, multiple properties act like an OR instead
+// of an AND. If an inserted job is found whose properties match any of the set
+// opts properties, a test failure is triggered.
```
bgentry, line 42: I don't this behavior is what I would expect here. My intuition (based on how I've used queue assertions for other libraries/ecosystems) is that I'm allowed to be as specific as I desire here, and that the assertion will ensure that no jobs have been inserted which match all of the conditions I specified. If I want, I can easily make sure that no jobs have been inserted on a certain queue, or if I want to be more specific, I can make sure that no jobs have been inserted on that queue and with that job type.

And in the affirmative case where I want to make sure a job _was_ inserted, it likewise matches _all_ of the criteria I specify, but ignores all those which I don't.

rivertest/rivertest.go
bgentry: See my above comment, I think the behavior on this might be backwards.

But as far as the type thing, I would definitely avoid the aliasing option because it results in poorer usability. If we define these types right next to each other, it wouldn't be that hard to keep them lined up.  Or alternatively if we can get away with just using the same type in both places, that's even better.

brandur:
@bgentry Alright I ended up flipping the conditions on `RequireNotInserted` options so they act like an AND instead of an OR. Also ended up improving the error messages so that if multiple properties on an inserted job row do or don't match, we return all of them instead of just the first. Added additional tests to check everything. Want to take another look?

rivertest/rivertest.go
```diff
@@ -29,8 +30,18 @@ type testingT interface {
...
 
-// Options for RequireInserted or RequireManyInserted including expectations for
-// various queuing properties that stem from InsertOpts.
+// Options for RequireInserted functions including expectations for various
+// queuing properties that stem from InsertOpts.
+//
+// Multiple properties set on this struct increase the specifity on a job to
+// match, acting like an AND condition on each.
+//
+// In the case of RequireInserted or RequireInsertdMany, if multiple properties
```
bgentry, line 39:
```suggestion
// In the case of RequireInserted or RequireInsertedMany, if multiple properties
```

rivertest/rivertest.go
```diff
@@ -291,22 +414,48 @@ func compareJobToInsertOpts(t testingT, jobRow *rivertype.JobRow, expectedOpts R
...
+			if requireNotInserted {
+				failures = append(failures, fmt.Sprintf("max attempts equal to excluded %d", expectedOpts.MaxAttempts))
+			}
+		} else {
+			if requireNotInserted {
+				return true // any one property doesn't match; assertion passes
+			} else {
+				failures = append(failures, fmt.Sprintf("max attempts %d not equal to expected %d", jobRow.MaxAttempts, expectedOpts.MaxAttempts))
+			}
+		}
```
bgentry, line 430: There's definitely a lot of duplication in these attr comparisons. Do you think it'd be worth extracting a generic helper for these comparable attrs (int, string)?

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   A new `river_queue` table is introduced in the v4 migration for this purpose. Upon startup, every producer in each River `Client` will make an `UPSERT` query to the database to either register the queue as being active, or if it already exists it will instead bump the timestamp to keep it active. This query will be run periodically in each producer as long as the `Client` is alive, even if the queue is paused. A separate query will delete/purge any queues which have not been active in awhile (currently fixed to 24 hours).
 
   `QueuePause` and `QueueResume` APIs have been introduced to `Client` pause and resume a single queue by name, or _all_ queues using the special `*` value. Each producer will watch for notifications on the relevant `LISTEN/NOTIFY` topic unless operating in poll-only mode, in which case they will periodically poll for changes to their queue record in the database.
+- `RequireNotInserted` test helper (in addition to the existing `RequireInserted`) that verifies that a job with matching conditions was _not_ inserted. [PR #237](https://github.com/riverqueue/river/pull/237).
```
bgentry, line 22: We missed the release on this so unfortunately this will need to get rebased and pushed to the unreleased section.

bgentry (commented):
Nice work! This should be super useful.

rivertest/rivertest.go
brandur: Thx. Fixed.

CHANGELOG.md
brandur: Moved it out to "unreleased".

rivertest/rivertest.go
brandur: Looked into this one. I'm not against the idea, but the helper would have to take a lot of different parameters to replicate that functionality that's in here now, and even if it's DRYer, I think overall it might obscure rather than clarify. The special cases of `ScheduledAt` and `Tags` would also be much harder to shoehorn in there and would probably have to stay on there own, thereby making the overall value a bit more dubious again.

brandur:
Thx!

brandur:
Actually, I guess that wasn't an approval, but lemme know if you disagree with the above.

bgentry (approved):
Meant to approve, sorry!

brandur:
Oh cool. No worries!

PR #240, opened 2024-02-29
fix memory leak on job execution from baseservice

### LINKED ISSUE #239: Memory leak?
We are trialing River in production. We have a consumer process that consumes a message from Kafka and inserts a River job. We've been running with 1 River worker process (MaxWorkers = 4) that runs separately from the consumer process that has an insert-only client. I've observed memory increasing through relative low volume periods, and it increases quickly during high volume of 30+ messages per second. The dip near the end of the chart is a restart. You can see memory increasing during a relatively low volume period during the overnight hours.

image

image

I'll paste some of the code below:

Insert-only Client creation:
```go
riverClient, err := river.NewClient(riverpgxv5.New(ds.PgxPool()), &river.Config{})
if err != nil {
	log.Errorf("could not initialize new River client: %s", err.Error())
	panic(err)
}
```

River job insertion
```go
_, err := riverClient.Insert(ctx, ShipmentMessage{JSONString: string(message.Value)}, nil)
if err != nil {
	log.Errorf("failed to insert job: %s", err.Error())
} else {
	metrics.Count("shipment_message_worker.message_inserted", 1, []string{}, 1)
}
```

The worker (this process is where the memory may be leaking)
```go
func (ShipmentMessage) Kind() string { return "shipment_message" }

func (rw *ShipmentMessageWorker) Work(ctx context.Context, job *river.Job[ShipmentMessage]) error {
	metrics.Count("shipment_message_worker.message_processed", 1, []string{}, 1)

	return nil
}
```

Worker initialization:
```go
func InitRiverWorkerAndClient(ctx context.Context, ds *postgres.Datastore, cfg Config) {
	// add river worker
	riverWorkers := river.NewWorkers()
	if err := river.AddWorkerSafely(riverWorkers, &ShipmentMessageWorker{
		Logger: log.New(),
	}); err != nil {
		log.Errorf("could not register new River worker: %s", err.Error())
		panic(err)
	}
	log.Infof("new River worker registered successfully")

	// create new river client
	client, err := river.NewClient(riverpgxv5.New(ds.PgxPool()), &river.Config{
                // cfg.MaxRiverWorkers is 4
		Queues:  map[string]river.QueueConfig{river.QueueDefault: {MaxWorkers: cfg.MaxRiverWorkers}},
		Workers: riverWorkers,
	})
	if err != nil {
		log.Errorf("could not initialize new River client: %s", err.Error())
		panic(err)
	}

	// start the client
	if err := client.Start(ctx); err != nil {
		log.Errorf("failed to start River client: %s", err.Error())
		panic(err)
	}

	// ensures the worker doesn't exit until it's terminated by a signal
	<-ctx.Done()
       // Note we use context cancelation to stop the worker on shutdown.
}
```

bgentry:
The `BaseService` type was originally only used for long-running internal services. At some point, we refactored the job execution code to also make use of this type. That exposed a memory leak: every time a new `BaseService` is `Init`'d, a new random source is allocated. That meant for every single job.

This is of course totally unnecessary, and to avoid it we can move the rand source to the `Archetype`, which is only allocated once per client.

Partially fixes #239.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a memory leak caused by allocating a new random source on every job execution. Thank you @shawnstephens for reporting ❤️ [PR #240](https://github.com/riverqueue/river/pull/240).
 - Fix a problem where `JobListParams.Queues()` didn't filter correctly based on its arguments. [PR #212](https://github.com/riverqueue/river/pull/212).
 - Fix a problem in `DebouncedChan` where it would fire on its "out" channel too often when it was being signaled continuousy on its "in" channel. This would have caused work to be fetched more often than intended in busy systems. [PR #222](https://github.com/riverqueue/river/pull/222).
 
--- a/client.go
+++ b/client.go
@@ -24,6 +24,7 @@ import (
 	"github.com/riverqueue/river/internal/maintenance"
 	"github.com/riverqueue/river/internal/notifier"
 	"github.com/riverqueue/river/internal/rivercommon"
+	"github.com/riverqueue/river/internal/util/randutil"
 	"github.com/riverqueue/river/internal/util/sliceutil"
 	"github.com/riverqueue/river/internal/util/valutil"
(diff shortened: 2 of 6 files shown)


brandur (approved):
Awesome. Thanks for the fix!

PR #244, opened 2024-03-01
Add `rivertype` module to release instructions

brandur:
As of #212, `rivertype` became a separate Go module from the mainline
`river`. Here, add it to the release instructions so that it gets
properly tagged along with other all other modules.

Also add it to the `update-submodule-versions` program. This isn't
strictly necessary because `rivertype` doesn't have any submodule
versions to update, but it feels like it should be there for
completeness just in case, and also because it makes the list in the
program a complete list of all Go modules in the project.


bgentry (approved):

brandur:
ty!

PR #245, opened 2024-03-01
Tee up release `v0.0.24`

brandur:
Prepares `v0.0.24`, a small release, but one which contains an important
fix from #243 which resolves a context memory leak that'd cause River's
memory usage to bloat over time.


bgentry (approved):

brandur:
thx.

PR #246, opened 2024-03-01
More robust listener close in `riverpgxv5`

### LINKED ISSUE #248: Possible connection re-use issue in `0.0.24`
We just put `0.0.24` into `production` and it's starting to get into a state where `river` is logging hundreds of times a second with

```json
{
    "level": "error",
    "time": "2024-03-01T23:56:08.679305457Z",
    "notifier": {
        "err": {
            "error": "connection already established",
            "kind": "*errors.errorString",
            "stack": null
        }
    },
    "subsystem": "river",
    "message": "error establishing connection from pool"
}
```

Via:

- https://github.com/riverqueue/river/blob/035ba5919e944c1ad027582b877192cc16516fc2/internal/notifier/notifier.go#L130
- https://github.com/riverqueue/river/blob/035ba5919e944c1ad027582b877192cc16516fc2/riverdriver/riverpgxv5/river_pgx_v5_driver.go#L491

(My apologies for the lack of a stack trace)

This is thrashing CPU significantly (the small bumps are the deployments that haven't rolled out `0.0.24` yet):

Screenshot 2024-03-01 at 5 15 37 PM

I managed to capture the lead up to **ONE** of these bursts and it looks like a TCP error on the PostgreSQL connection started the death spiral

```json
{
    "level": "info",
    "time": "2024-03-01T23:56:07.097802512Z",
    "num_completed_jobs": 40,
    "num_jobs_running": 0,
    "queue": "default",
    "subsystem": "river",
    "message": "producer: Heartbeat"
}

{
    "level": "error",
    "time": "2024-03-01T23:56:08.679126682Z",
    "notifier": {
        "err": {
            "error": "tls: failed to send closeNotify alert (but connection was closed anyway): write tcp 10.122.48.181:44034->10.122.30.240:5432: i/o timeout",
            "kind": "*fmt.wrapError",
            "stack": null
        }
    },
    "subsystem": "river",
    "message": "error closing listener"
}

{
    "level": "error",
    "time": "2024-03-01T23:56:08.679305457Z",
    "notifier": {
        "err": {
            "error": "connection already established",
            "kind": "*errors.errorString",
            "stack": null
        }
    },
    "subsystem": "river",
    "message": "error establishing connection from pool"
}
```

I'm sorry I don't have anything more conclusive right now but wanted to at least get this in your hands in case something rings a bell.

brandur:
As part of #239, we're observing the possibility of the notifier going
into a hot loop where it's trying to reopen a listener connection, but
the listener won't let it because it thinks the connection is already
open.

A suspect is the listener's `Close` implementation, which in the event
of an error, returns the error and fails to release an underlying
connection, putting it into a state where it's never reusable.

Here, modify `Close` so that it always releases and unsets an underlying
connection regardless of the error state returned.

I tried to add a test case for this, but reading through pgx and net
source code, I couldn't find any way to simulate an error from `Close`
(I thought a cancelled context would do it, but it does not), so I had
to leave it.

Thanks @mfrister for pointing this one out!

Fixes #248.

### DIFF
--- a/riverdriver/riverpgxv5/river_pgx_v5_driver.go
+++ b/riverdriver/riverpgxv5/river_pgx_v5_driver.go
@@ -475,12 +475,14 @@ func (l *Listener) Close(ctx context.Context) error {
 		return nil
 	}
 
-	if err := l.conn.Conn().Close(ctx); err != nil {
-		return err
-	}
+	err := l.conn.Conn().Close(ctx)
+
+	// Regardless of the error state returned above, always release and unset
+	// the listener's local connection.
 	l.conn.Release()
 	l.conn = nil
-	return nil
+
+	return err
 }
 
 func (l *Listener) Connect(ctx context.Context) error {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed a problem in `riverpgxv5`'s `Listener` where it wouldn't unset an internal connection if `Close` returned an error, making the listener not reusable. Thanks @mfrister for pointing this one out! [PR #246](https://github.com/riverqueue/river/pull/246).
+
 ## [0.0.24] - 2024-02-29
 
 ### Fixed
--- a/internal/riverinternaltest/riverdrivertest/riverdrivertest.go
+++ b/internal/riverinternaltest/riverdrivertest/riverdrivertest.go
@@ -1617,9 +1617,6 @@ func setupListener[TTx any](ctx context.Context, t *testing.T, getDriverWithPool
 	driver := getDriverWithPool(ctx, t)
 
 	listener := driver.GetListener()
-	t.Cleanup(func() { require.NoError(t, listener.Close(ctx)) })
-
-	require.NoError(t, listener.Connect(ctx))
 
 	return listener, &testListenerBundle[TTx]{
 		driver: driver,
@@ -1629,6 +1626,13 @@ func setupListener[TTx any](ctx context.Context, t *testing.T, getDriverWithPool
 
 func ExerciseListener[TTx any](ctx context.Context, t *testing.T, getDriverWithPool func(ctx context.Context, t *testing.T) riverdriver.Driver[TTx]) {
 	t.Helper()
+
+	connectListener := func(ctx context.Context, t *testing.T, listener riverdriver.Listener) {
+		t.Helper()
+
+		require.NoError(t, listener.Connect(ctx))
+		t.Cleanup(func() { require.NoError(t, listener.Close(ctx)) })
+	}
 
 	requireNoNotification := func(ctx context.Context, t *testing.T, listener riverdriver.Listener) {
 		t.Helper()
@@ -1653,10 +1657,19 @@ func ExerciseListener[TTx any](ctx context.Context, t *testing.T, getDriverWithP
 		return notification
 	}
 
+	t.Run("Close_NoOpIfNotConnected", func(t *testing.T) {
+		t.Parallel()
+
+		listener, _ := setupListener(ctx, t, getDriverWithPool)
+		require.NoError(t, listener.Close(ctx))
+	})
+
 	t.Run("RoundTrip", func(t *testing.T) {
 		t.Parallel()
 
 		listener, bundle := setupListener(ctx, t, getDriverWithPool)
+
+		connectListener(ctx, t, listener)
 
 		require.NoError(t, listener.Listen(ctx, "topic1"))
 		require.NoError(t, listener.Listen(ctx, "topic2"))
@@ -1695,6 +1708,8 @@ func ExerciseListener[TTx any](ctx context.Context, t *testing.T, getDriverWithP
 
 		listener, bundle := setupListener(ctx, t, getDriverWithPool)
 
+		connectListener(ctx, t, listener)
+
 		require.NoError(t, listener.Listen(ctx, "topic1"))
 
 		execTx, err := bundle.exec.Begin(ctx)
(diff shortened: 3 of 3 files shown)


brandur:
@bgentry Seriously tried to add a test case for this, but couldn't find a way to make `Close` on the PG conn return an error. The only way to do it I think would be to add additional test-only interface functions to inject state, which would be kind of gross. Thoughts?

mfrister:
Thanks for the fix!

I've deployed a version of my worker with riverpgxv5 from this branch and will report on Monday whether the issue appeared again. So far it looks good.

Enjoy your weekend!

Rshep3087:
There is this package from pgx https://github.com/jackc/pgx/blob/master/pgxtest/pgxtest.go. You can overwrite the CloseFunc on the test runner I believe.

brandur:
@mfrister Excellent. Thanks!

@Rshep3087 Hm, just reading through this module, I believe `ConnTestRunner` is more of a helper for controlling the flow of test runs rather than something that can act as a mock for a `pgx.Conn`.

riverdriver/riverpgxv5/river_pgx_v5_driver.go
```diff
@@ -475,12 +475,14 @@ func (l *Listener) Close(ctx context.Context) error {
 		return nil
 	}
 
-	if err := l.conn.Conn().Close(ctx); err != nil {
-		return err
-	}
```
bgentry, line 480: Man, I noticed [this change](https://github.com/riverqueue/river/pull/212/files#diff-445bd06e41eab3a8a89576b7cda31b0e834bbefa4fa969af11c1ce17766b1cceL149) when I reviewed the mega-PR, should have said something 😞

bgentry (approved):
The only ways I can think of to test this are (a) use an interface over the conn here, which feels excessive, or (b) somehow force the server to uncleanly terminate the conn or reply with an error upon close. 

I'm fine with not going forward with one of those approaches at this time, they seem to heavy for the benefit at this stage. Thanks for the quick fix :shipit:

riverdriver/riverpgxv5/river_pgx_v5_driver.go
brandur: Doh. Well, I was going to say "next time", but hopefully nothing like that PR ever happens again :)

brandur:
> The only ways I can think of to test this are (a) use an interface over the conn here, which feels excessive, or (b) somehow force the server to uncleanly terminate the conn or reply with an error upon close.

Yeah, I had the same reactions.

I think (a) is plausible, but has the downside that it makes the code a little more painful to use. Rather than being able to easily jump easily into concrete implementations, you now have to jump through the interface, which is always a bit annoying.

And RE: (b), I read through a lot of pgx source around `Close`, and man, as far as I can tell, it is quite difficult to ever get it to return an error. The underlying `net.Conn` under the pgx connection which is under the pgx pool connection will under some cases, but I couldn't find an obvious way to trigger it, especially from this level.

> I'm fine with not going forward with one of those approaches at this time, they seem to heavy for the benefit at this stage. Thanks for the quick fix :shipit:

Sweet. Yeah, that was my conclusion as well. Possible maybe to add some form of testing, but maybe more trouble than it's worth, and getting the fix out the door is the best compromise for the time being.

Thanks!

dhermes:
Regarding unit testing, I have a little bit of experience here. I [hunted down][1] a bug in `lib/pq` a few years back and ended up replicating the bug locally while crafting a fix by using a TCP [proxy][2] that would send an `RST` packet after some signal had been received (but before the actual PostgreSQL connection was closed).

From what I [understand][3], your unit tests hit a **REAL** PostgreSQL instance, so you could run a proxy within a Go unit test and put that proxy in front of the PostgreSQL instance available to the unit tests.

🤞 I'd hope a plain old TCP error over a non-TLS connection would do the trick, but maybe not. In my log capture around the issue that this fixed (#248), it was actually a TLS protocol issue that showed up in the error:

```
{
    "level": "error",
    "time": "2024-03-01T23:56:08.679126682Z",
    "notifier": {
        "err": {
            "error": "tls: failed to send closeNotify alert (but connection was closed anyway): write tcp 10.122.48.181:44034->10.122.30.240:5432: i/o timeout",
            "kind": "*fmt.wrapError",
            "stack": null
        }
    },
    "subsystem": "river",
    "message": "error closing listener"
}
```

However, if I had to guess I'd imagine the **TCP** connection is what had already been closed (e.g. via RST).

[1]: https://blog.bossylobster.com/2020/12/broken-pipe.html
[2]: https://github.com/dhermes/bossylobster-blog/blob/c2017d3220d0814ee45e11085ecafc515464f938/content/code/rst-proxy.go
[3]: https://github.com/riverqueue/river/blob/6b03de1f181272ce953c3f7fabbef0a5ad8e4996/.github/workflows/ci.yaml#L29-L39

brandur:
@dhermes Thanks for the added detail there! I'm sort of hoping to avoid a full-blown proxy for the time being if we can possibly do it (looked at yours — looks good, but definitely a substantial amount of new code). You gave me the idea though of getting something more like an "in code" proxy going, and I found a fairly low touch way of simulating a `Close` error using a combination of pgxpool's `DialFunc` and a lightweight `net.Conn` mock. See #250.

``` go
type connStub struct {
	net.Conn

	closeFunc func() error
}

func newConnStub(conn net.Conn) *connStub {
	return &connStub{
		Conn: conn,

		closeFunc: conn.Close,
	}
}

func (c *connStub) Close() error {
	return c.closeFunc()
}
```

mfrister:
As [promised above](https://github.com/riverqueue/river/pull/246#issuecomment-1973377923):

> I've deployed a version of my worker with riverpgxv5 from this branch and will report on Monday whether the issue appeared again. So far it looks good.

The problem hasn't reappeared on my worker over the weekend, as expected.

brandur:
@mfrister That's great to hear!! Thanks for checking in about it.

PR #247, opened 2024-03-01
`riverdrivertest`: Decrease tolerance required in inserted timestamps

brandur:
Noticed a failure in one of the driver tests where it can take longer
than 100 milliseconds for an inserted row to have its timestamp compared
against the current time (not much longer, 143 ms versus 100 ms):

    $ go test .
    --- FAIL: TestDriverRiverPgxV5_Executor (0.00s)
        --- FAIL: TestDriverRiverPgxV5_Executor/LeaderGetElectedLeader (0.15s)
            riverdrivertest.go:1403:
                    Error Trace:    /Users/brandur/Documents/projects/river/internal/riverinternaltest/riverdrivertest/riverdrivertest.go:1403
                    Error:          Max difference between 2024-03-01 07:05:55.905609 -0800 PST m=+2.603374667 and 2024-03-01 15:05:55.763974 +0000 UTC allowed is 100ms, but difference was 141.635ms
                    Test:           TestDriverRiverPgxV5_Executor/LeaderGetElectedLeader
    FAIL
    FAIL    github.com/riverqueue/river     9.516s
    FAIL

This actually occurred locally rather than in GitHub Actions.

I believe these sorts of slowdowns can occur when you have a lot of
parallel tests running and the scheduling of goroutines becomes somewhat
less predictable. Increase tolerance from 100ms to 500ms.

### DIFF
--- a/internal/riverinternaltest/riverdrivertest/riverdrivertest.go
+++ b/internal/riverinternaltest/riverdrivertest/riverdrivertest.go
@@ -1382,8 +1382,8 @@ func ExerciseExecutorFull[TTx any](ctx context.Context, t *testing.T, driver riv
 			TTL:      leaderTTL,
 		})
 		require.NoError(t, err)
-		require.WithinDuration(t, time.Now(), leader.ElectedAt, 100*time.Millisecond)
-		require.WithinDuration(t, time.Now().Add(leaderTTL), leader.ExpiresAt, 100*time.Millisecond)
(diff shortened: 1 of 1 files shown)


bgentry (approved):
This is one that can really only be tested well with a synthetic clock and tweaks to the query. But this will do for now!

brandur:
> This is one that can really only be tested well with a synthetic clock and tweaks to the query. But this will do for now!

Yep, +1. Thanks!

PR #249, opened 2024-03-02
Tee up version `v0.0.25`

brandur:
Contains an important fix from #246 which resolves a problem wherein a
listener from the `riverpgxv5` driver couldn't be reused in cases where
`Close` returned an error.


bgentry (approved):

brandur:
Thanks!

PR #250, opened 2024-03-02
Add tests for listener `Close` + invoke only pool release

brandur:
Here, follow up #246 by adding a few more tests that verify a listener's
state after `Close` has been invoked, including if it returned an error,
which we're able to simulate by overriding pgx's `DialFunc` and
returning a stand-in stub for an underlying `net.Conn`.

Also, remove the explicit `Close` call on an underlying connection in
favor of just invoking the pool's `Release` function. In case of an
error condition, `Release` will detect that and do the right thing, and
pgx is better tested/vetted to make sure that right thing happens.

### DIFF
--- a/riverdriver/riverpgxv5/river_pgx_v5_driver.go
+++ b/riverdriver/riverpgxv5/river_pgx_v5_driver.go
@@ -475,10 +475,14 @@ func (l *Listener) Close(ctx context.Context) error {
 		return nil
 	}
 
+	// Release below would take care of cleanup and potentially put the
+	// connection back into rotation, but in case a Listen was invoked without a
+	// subsequent Unlisten on the same tpic, close the connection explicitly to
+	// guarantee no other caller will receive a partially tainted connection.
 	err := l.conn.Conn().Close(ctx)
 
-	// Regardless of the error state returned above, always release and unset
-	// the listener's local connection.
+	// Even in the event of an error, make sure conn is set back to nil so that
+	// the listener can be reused.
 	l.conn.Release()
 	l.conn = nil
 
--- a/riverdriver/riverpgxv5/river_pgx_v5_driver_test.go
+++ b/riverdriver/riverpgxv5/river_pgx_v5_driver_test.go
@@ -1,8 +1,13 @@
 package riverpgxv5
 
 import (
+	"context"
 	"errors"
+	"fmt"
+	"net"
+	"os"
 	"testing"
+	"time"
 
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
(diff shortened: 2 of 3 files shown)


riverdriver/riverpgxv5/river_pgx_v5_driver_test.go
```diff
@@ -34,10 +39,144 @@ func TestNew(t *testing.T) {
...
+
+		closeFunc: conn.Close,
+	}
+}
+
+func (c *connStub) Close() error {
+	return c.closeFunc()
+}
+
+func testPoolConfig() *pgxpool.Config {
```
brandur, line 155: Maybe `riverinternaltest` should be its own package so we can reuse this stuff, but ugh, skipped any further refactoring there for now.

riverdriver/riverpgxv5/river_pgx_v5_driver.go
```diff
@@ -475,14 +475,10 @@ func (l *Listener) Close(ctx context.Context) error {
 		return nil
 	}
 
-	err := l.conn.Conn().Close(ctx)
-
-	// Regardless of the error state returned above, always release and unset
-	// the listener's local connection.
```
bgentry, line 481: I'm not actually sure this is a good idea and I don't think we want to return the conn to the pool. If there are `LISTEN` subscriptions active which weren't properly `UNLISTEN`'d, the conn may still receive additional notifications that could interfere with other operations on it.

For pgbouncer compatibility reasons, we will probably also end up going a direction where the listener conn is created from a standalone config or at least a separate pool from the main river db pool.

Or maybe you can elaborate more on why you think this is a necessary change?

riverdriver/riverpgxv5/river_pgx_v5_driver_test.go
bgentry: not more packages 😩

bgentry (commented):
LGTM aside from not closing the conn, lmk if you disagree on that

riverdriver/riverpgxv5/river_pgx_v5_driver_test.go
brandur: Haha, ack.

riverdriver/riverpgxv5/river_pgx_v5_driver.go
brandur: Okay, fair point on that one, I forgot about the potential taint from not correctly using `Unlisten`. I put the `Close` back in and added a comment to remind us why it's written this way. One of the test cases had to go because pgx's implementation is _very_ specific, and `AfterRelease` doesn't get called if a connection was released that was already closed.

Want to take another look?

bgentry (approved):

brandur:
ty.

PR #251, opened 2024-03-02
Periodic job enqueuer: Keep timer paused around inserts to fix intermittency

brandur:
Here, aim to fix the intermittent test problem described in #215, in
which after inserting a periodic batch where periodic 500 ms jobs and
periodic 1500 ms jobs should've lined up, we come up with fewer 500 ms
jobs than expected, suggesting that only a 1500 ms job was inserted
during the run.

    --- FAIL: TestPeriodicJobEnqueuer (0.00s)
        --- FAIL: TestPeriodicJobEnqueuer/EnqueuesPeriodicJobs (1.57s)
            logger.go:225: time=2024-02-19T07:53:42.969Z level=INFO msg="PeriodicJobEnqueuer: Run loop started"
            periodic_job_enqueuer_test.go:112:
                    Error Trace:	/home/runner/work/river/river/internal/maintenance/periodic_job_enqueuer_test.go:82
                                                            /home/runner/work/river/river/internal/maintenance/periodic_job_enqueuer_test.go:112
                    Error:      	"[%!s(*dbsqlc.RiverJob=&{45 [1[23](https://github.com/riverqueue/river/actions/runs/7956225078/job/21716534964?pr=214#step:8:24) 125] 0 <nil> [] {482531000 63843926023 0x107e100} [] <nil> periodic_job_500ms 25 [123 125] 1 default available {482709000 63843926023 0x107e100} []}) %!s(*dbsqlc.RiverJob=&{46 [123 125] 0 <nil> [] {983499000 63843926023 0x107e100} [] <nil> periodic_job_500ms 25 [123 125] 1 default available {983650000 63843926023 0x107e100} []})]" should have 3 item(s), but has 2
                    Test:       	TestPeriodicJobEnqueuer/EnqueuesPeriodicJobs
                    Messages:   	Expected to find exactly 3 job(s) of kind: periodic_job_500ms, but found 2
            logger.go:225: time=20[24](https://github.com/riverqueue/river/actions/runs/7956225078/job/21716534964?pr=214#step:8:25)-02-19T07:53:44.472Z level=INFO msg="PeriodicJobEnqueuer: Run loop stopped"
    FAIL
    FAIL	github.com/riverqueue/river/internal/maintenance	1.761s

As with many of these intermittent problems, this one is quite difficult
to fully confirm a fix on because I'm finding it basically impossible to
reproduce locally, regardless of the number of test iterations I run.
However, after reauditing the code, I think I know what the problem was.

[...]

Issue #215, opened 2024-02-19
Intermittent test failure: `TestPeriodicJobEnqueuer/EnqueuesPeriodicJobs`


brandur:
Another intermittent failure. This time in the periodic job enqueuer.

Sample failure run:

https://github.com/riverqueue/river/actions/runs/7956225078/job/21716534964?pr=214

```
--- FAIL: TestPeriodicJobEnqueuer (0.00s)
    --- FAIL: TestPeriodicJobEnqueuer/EnqueuesPeriodicJobs (1.57s)
        logger.go:225: time=2024-02-19T07:53:42.969Z level=INFO msg="PeriodicJobEnqueuer: Run loop started"
        periodic_job_enqueuer_test.go:112: 
            	Error Trace:	/home/runner/work/river/river/internal/maintenance/periodic_job_enqueuer_test.go:82
            	            				/home/runner/work/river/river/internal/maintenance/periodic_job_enqueuer_test.go:112
            	Error:      	"[%!s(*dbsqlc.RiverJob=&{45 [1[23](https://github.com/riverqueue/river/actions/runs/7956225078/job/21716534964?pr=214#step:8:24) 125] 0 <nil> [] {482531000 63843926023 0x107e100} [] <nil> periodic_job_500ms 25 [123 125] 1 default available {482709000 63843926023 0x107e100} []}) %!s(*dbsqlc.RiverJob=&{46 [123 125] 0 <nil> [] {983499000 63843926023 0x107e100} [] <nil> periodic_job_500ms 25 [123 125] 1 default available {983650000 63843926023 0x107e100} []})]" should have 3 item(s), but has 2
            	Test:       	TestPeriodicJobEnqueuer/EnqueuesPeriodicJobs
            	Messages:   	Expected to find exactly 3 job(s) of kind: periodic_job_500ms, but found 2
        logger.go:225: time=20[24](https://github.com/riverqueue/river/actions/runs/7956225078/job/21716534964?pr=214#step:8:25)-02-19T07:53:44.472Z level=INFO msg="PeriodicJobEnqueuer: Run loop stopped"
FAIL
FAIL	github.com/riverqueue/river/internal/maintenance	1.761s
```

brandur:
Sigh, I saw this again [1]:

```
--- FAIL: TestPeriodicJobEnqueuer (0.00s)
    --- FAIL: TestPeriodicJobEnqueuer/EnqueuesPeriodicJobs (1.60s)
[...]

PR #253, opened 2024-03-07
Make notifier operations more deterministic + fix intermittency

brandur:
I have a hypothesis that a lot of our test intermittency in high level
clien tests stem from deep down in the notifier. It works the vast
majority of the time, but looking into its implementation, it's
currently written to be non-deterministic. When using `Listen` to add a
new subscription, the notifier only attempts to do so asynchronously by
putting it in a queue for eventually processing. Callers don't know
whether adding a subscription succeeded, and more importantly (as far as
our intermittency problems are concerned), don't know _when_ it may have
succeeded.

Here, we make quite a few cosmetic changes, but the most important one
is to the API and implementation of `Listen`:

    sub, err := notifier.Listen(...)

Notably, it:

1. Now returns an error. If there was a problem listening, the caller is
   told about it.

2. Operations synchronously. `Listen` waits until a subscription could
   be confirmed, and then returns it, with the returned subscription
   known to be working immediately, as opposed to an indeterminate time
   in the future.

The synchronous works by having `Listen` schedule an "interrupt", so if
the service is it its main loop blocking on `WaitForNotification`, the
wait is cancelled, the loop processes any pending listens and unlistens
in interrupts, then it reenters `WaitForNotification` once again.

We add a much more comprehensive test suite to help verify various
behavioral edges, and suss out problems via use of `-race`. I believe I
found quite a few other smaller bugs in the notifier while doing this,
but admittedly things got a bit muddied because I surely introduced many
of those bugs during my main refactor.

A few other miscellaneous changes:

* The notifier becomes a `startstop`. Along with benefits around
  preventing double starts and many concurrency problems, it also
  changes to be able to error quickly in case of an obvious immediate
  problem like an unusable connection pool.

* We implement a true exponential backoff wait in case of a bad error in
  the main loop.

* Try to improve logging so that in case of future test failures, we
  should have a fairly good trace to see what happened when leading up
  to a failure.

To confirm/disconfirm resolution of the intermittency problems, I've
been running the infamous:

    $ go test . -run Test_Client_InsertTriggersImmediateWork -race -count 5000

It's had about 10 successes now. I did run into one intermittent failure
on one run, but it was a brand new one in the client that's not the
usual we see in CI, and I think unrelated to any changes made here.


internal/baseservice/base_service.go
```diff
@@ -106,6 +127,53 @@ func (s *BaseService) CancellableSleepRandomBetween(ctx context.Context, sleepDu
...
 }
 
+// CancellableSleepRandomBetween sleeps for a random duration between the given
+// bounds (max bound is exclusive), but returns early if context has been
+// cancelled. If the service's `DisableSleep` option is set (usually appropriate
+// in tests), sleep is skipped automatically.
+//
+// This variant returns a channel that should be waited on and which will be
+// closed when either the sleep or context is done.
+func (s *BaseService) CancellableSleepRandomBetweenC(ctx context.Context, sleepDurationMin, sleepDurationMax time.Duration) <-chan struct{} {
```
brandur, line 137: Okay, so a couple of these new functions aren't used here yet, but gonna ask you to take some of it on faith. What happened is that I was working on this for a different branch, then started working on this project, and realized I wanted one of the functions for this branch. Instead of pulling in the one specific function I wanted, I just pulled in all the base service changes so I didn't have to tease apart all the tests I'd added and everything.

internal/leadership/elector.go
```diff
@@ -114,8 +114,18 @@ func (e *Elector) Run(ctx context.Context) {
...
 
-	subscription := e.notifier.Listen(notifier.NotificationTopicLeadership, handleNotification)
-	defer subscription.Unlisten()
+	sub, err := e.notifier.Listen(ctx, notifier.NotificationTopicLeadership, handleNotification)
+	if err != nil {
+		if !errors.Is(err, context.Canceled) {
+			// TODO(brandur): We shouldn't be silently swallowing major start up
+			// problems like this. Unfortunately, Run doesn't return an error, so
+			// fixing this will be a big refactor, so don't do it right now.
+			e.logger.Error(e.name+": Error listening for leadership changes", "err", err, "topic", notifier.NotificationTopicJobControl)
```
brandur, line 123: See TODO here, but I don't like this and will fix it, but can't right now. I'd really like to see both the elector and producer return errors when problems happen on start up instead of silently swallowing errors.

It'll need some refactoring to get there, but once I do that this will return an error instead of producing a log statement with a TODO.

internal/maintenance/job_cleaner_test.go
```diff
@@ -71,7 +72,7 @@ func TestJobCleaner(t *testing.T) {
 		cleaner.Logger = riverinternaltest.LoggerWarn(t) // loop started/stop log is very noisy; suppress
 		cleaner.TestSignals = JobCleanerTestSignals{}    // deinit so channels don't fill
 
-		runStartStopStress(ctx, t, cleaner)
```
brandur, line 74: Just moved this into a test package so we can use it across multiple packages to test start/stop services.

internal/maintenance/startstop/start_stop.go
```diff
@@ -5,6 +5,23 @@ import (
 	"sync"
 )
 
+// Service is a generalized interface for a service that starts and stops,
+// usually one backed by embedding BaseStartStop.
+type Service interface {
```
brandur, line 10: Move in from the maintenance package so it can be reused. Required by `startstoptest.Stress`, but will also likely be useful more broadly.

producer_test.go
```diff
@@ -71,7 +71,7 @@ func Test_Producer_CanSafelyCompleteJobsWhileFetchingNewOnes(t *testing.T) {
 	}))
 
 	ignoreNotifierStatusUpdates := func(componentstatus.Status) {}
-	notifier := notifier.New(archetype, listener, ignoreNotifierStatusUpdates, riverinternaltest.Logger(t))
```
brandur, line 74: Logger now comes in from the archetype instead of another parameter that's really just a duplicate.

riverdriver/riverpgxv5/river_pgx_v5_driver.go
```diff
@@ -507,31 +507,31 @@ func (l *Listener) Connect(ctx context.Context) error {
 }
 
 func (l *Listener) Listen(ctx context.Context, topic string) error {
-	l.mu.RLock()
-	defer l.mu.RUnlock()
+	l.mu.Lock()
+	defer l.mu.Unlock()
```
brandur, line 511: I missed it before until adding more tests to stress it, but even `LISTEN` keeps the connection busy, so every listener operation should actually take a full mutex lock instead of allowing multiple readers.

brandur:
@bgentry Holy crap, I just reran this PR's test suite 10x+ times in a row, each including a matrix of 6x test suite runs. I did observe 2x random failures of #215 [1], but ZERO instances of #213. I'm fairly confident this change fixes it.

I know a lot of LOCs in the notifier change, but I added quite a comprehensive test suite which I was running the whole time with `-race -count 100` to try and suss out all possible problems, race or otherwise. I think test coverage is _much_ better than before, and more specific too so we don't have to catch all problems with broad tests like `Test_Client_InsertTriggersImmediateWork`. Mind taking a look?

---

[1] That's out of > 60 runs, so not too bad, and I'm 99% sure the failure isn't related to these changes since they don't really touch anything in the maintainer. My fix in #251 helped decrease the frequency of this problem, but apparently didn't resolve it completely.

internal/notifier/notifier.go
```diff
@@ -24,189 +28,327 @@ const (
...
+	topic        NotificationTopic
+	unlistenOnce sync.Once
 }
 
-func (s *Subscription) Unlisten() {
+func (s *Subscription) Unlisten(ctx context.Context) {
 	s.unlistenOnce.Do(func() {
-		s.notifier.unlisten(s)
+		// Unlisten uses background context in case of cancellation.
+		if err := s.notifier.unlisten(context.Background(), s); err != nil { //nolint:contextcheck
```
bgentry, line 40: Is it safe to use a background context with no timeout here? I'm guessing not, because this underlying code:

https://github.com/riverqueue/river/blob/35ee45ee57e15b91eb1f722c4bf6ff69c003779d/riverdriver/riverpgxv5/river_pgx_v5_driver.go#L524-L530

Could hang forever if the context never times out and there isn't some other mechanism to force it to give up.

I get that you don't want to inherit from `ctx` because it may/will already be cancelled at the time this runs, but it probably needs at least _some_ timeout to make sure it doesn't hang forever.

internal/notifier/notifier.go
```diff
@@ -220,122 +362,175 @@ func (n *Notifier) runOnce(ctx context.Context) error {
...
+			if err := n.sendInterruptAndReceiveResult(func() error { return n.listenerListen(ctx, topic) }); err != nil {
+				removeSub()
+				return nil, err
+			}
+		} else {
+			var justConnected bool
+
+			if !n.isConnected {
+				if err := n.listenerConnect(ctx, true); err != nil {
+					removeSub()
```
bgentry, line 468: I don't think this behavior aligns with how the users of the notifier are designed today. I believe they all expect that once they call `Listen(topic)`, the notifier will do its best to make sure that `topic` is listened to until the subscriber calls `Unlisten` on the subscription. This is the case regardless of what errors/reconnects/etc happen on the notifier in the interim.

What's here seems fundamentally different. The listen attempt can _fail inline_, in which case that topic will not be listened to at all unless the caller tries to `Listen()` again. Because those subscribers are just swallowing the errors as of now (due to them being designed to consume the previous API which did not surface such errors), they will just carry on operating but without a functional listener. That doesn't seem like a desirable outcome.

bgentry (commented):
I still need to review the notifier tests, but I've looked at everything else here. These are my initial thoughts & concerns.

It's awesome this fixes the flakiness! I don't think it's quite ready to ship though, at least not without addressing the fact that expectations have changed for notifier subscribers. I would also feel more comfortable about the concurrency situation if we had some test coverage that actually exercised various concurrent access patterns (as discussed offline).

internal/leadership/elector.go
brandur: > I think the challenge is that it's probably desirable to have the client fail to start if any of its critical subprocesses fail to start correctly at initial boot, however if those subprocesses fail at any other point they should enter into a graceful exponential backoff loop to try and get healthy again.

Ah yeah, as mentioned on Slack, this is exactly how the new notifier is written to work: we fail fast in case of a major problem so the user can find out about it, but subsequent failures after start are in internalized, and services will do their best to make sure they get healthy again. We also now have tests in the notifier to prove that this works, helping to take the guesswork out of it. (See `WaitErrorAndBackoff`, `BackoffSleepCancelledOnStop`, and `StillFunctionalAfterMainLoopFailure`.)

internal/notifier/notifier.go
```diff
@@ -24,189 +28,327 @@ const (
...
+		n.statusChangeFunc(componentstatus.Initializing)
+		var wg sync.WaitGroup
+
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			n.deliverNotifications(ctx)
+		}()
+
+		for attempt := 0; ; attempt++ {
```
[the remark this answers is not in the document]
brandur: This is taken care of by the base service:

``` go
// Reset exponential backoff counters after a certain point so that sleep
// durations aren't flung into a ridiculously distant future.
const exponentialBackoffAttemptResetAfter = 10

func (s *BaseService) exponentialBackoffSecondsWithoutJitter(attempt int) float64 {
	// We use a different exponential backoff algorithm here compared to the
	// default retry policy (2**N versus N**4) because it results in more
	// retries sooner. When it comes to exponential backoffs in services we
	// never want to sleep for hours/days, unlike with failed jobs.
	return math.Pow(2, float64(attempt%exponentialBackoffAttemptResetAfter))
}
```

I went back and forth on where best to implement it, but putting it in the base has the major advantages of (1) it being in one place and impossible to forget, and (2) more easily testable to make sure the right thing is happening.

internal/notifier/notifier.go
brandur: I'm still not sure that this is something that happens IRL, but in the interest of safety, just added timeouts to all listener operations.

internal/notifier/notifier.go
brandur: I don't think it's too different in practice to how it worked before — after `Listen` has returned successfully once, then any service that calls it is considered listening, and any subsequent failures in the machinery will cause the notifier to do its best to make sure that those subscriptions stay connected.

It may be possible (although very unlikely) currently for the producer + elector to fail on that initial `Listen` connection, but I have branches for both of them that refactor them to services that can propagate errors on `Start` like I'm proposing the notifier will here. The eventual (and soon to be reached) destination will be an initial `Listen` that fails will propagate  its error, but assuming it can succeed, will become reliably error tolerant, just as things are currently. This will also allow us to remove these sort of hacking checks that are currently required on startup because nothing propagates an error:

https://github.com/riverqueue/river/blob/4bfe88a21a2a73dce03a0a35e56970f6f11ca61e/client.go#L586-L596

brandur:
@bgentry I also add a new stress test that does lots of listens/unlistens while simultaneously pounding the notifier with notifications.

internal/notifier/notifier.go
bgentry: In the event I haven't recounted the outage at Mux that was caused by a lock taken within a transaction where an operation hung (likely some k8s/aws networking blip) and there was no txn or context timeout, remind me to do so 🤬

internal/notifier/notifier.go
bgentry: Ok, I think the issue is that there's currently no retry loop on callers to `Listen()`. This was fine previously, because no matter what, the notifier would always keep track of things that were supposed to be listened to after that initial call, regardless of any errors encountered. However with this current design, the `Listen()` call can error, the notifier will then _not_ stay subscribed to that topic, and the caller will not retry.

If you're saying you have a fast follow PR that will fix these issues, then I don't think this needs to be held up from merging, but we should probably get those additional fixes in before cutting a new release just so we're not shipping a regression like that.

bgentry (approved):

internal/notifier/notifier.go
brandur: Okay, I just pushed a change that puts all listens into a retry loop with exponential backoff and a cancel check.

I don't think this is the right answer longer term, but seems like a reasonable precaution in the interim before we can start returning errors from elector/producer.

internal/notifier/notifier.go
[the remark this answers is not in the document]
brandur: Ah, the max before reset idea is that it it'll backoff totally normally, but after hitting 10 attempts, the sleep schedule resets. So there's no quick burst — it sleeps on a normal schedule, but makes sure to reset before the sleep durations get too obscene.

Just to make this more obvious, I've retooled the API to now look like this:

``` go
s.CancellableSleepExponentialBackoff(ctx, attempt, baseservice.MaxAttemptsBeforeResetDefault)
```

So the reset is still built in, but it's now visible in the function invocation that a reset is a thing, which will hopefully improve legibility. Not necessarily the be all end all if this turns out to be suboptimal — will keep an eye on it to make sure it continues to be reasonable.

brandur:
Thanks!

PR #255, opened 2024-03-08
Generate default client IDs using host and client start time

brandur:
Here, move away from ULIDs for generating client IDs, and over to a
custom ID generator that uses a combination of host and client start
time. This has two benefits:

* Host/start time are more descriptive for humans. It should give an
  operator a rough idea as to where the name of a client originated
  from, and if it's a client that's likely long dead (if the timestamp
  is very old).

* Lets us drop our ULID package dependency, which we don't use for
  anything else.

Generated IDs look like this for example:

    miniml2_local_2024_03_07T04_39_12

I've tried to arrange them so that on most systems you'll be able to
double click on an ID to select the whole thing, which might be helpful
while copying information or debugging a production problem. Characters
like colons (`:`) are less friendly for double-click-to-select, so I've
avoided them. Hyphens and dots aren't universally friendly for double
click either (depends on program and configuration), so I've avoided
those as well.


bgentry:
> Generated IDs look like this for example:
> 
> ```
> miniml2.local_2024-03-07T04-39-12
> ```
> 
> I've tried to arrange them so that on most systems you'll be able to double click on an ID to select the whole thing, which might be helpful while copying information or debugging a production problem. Characters like colons (`:`) are less friendly for double-click-to-select, so I've avoided them. Hyphens aren't universally friendly for double click either (depends on program and configuration), but they're more so.

Hmm, macOS terminal and Chrome both want to break on the `.` and `-` characters in the above, so unfortunately it's pretty far from double-clickable 😕

bgentry (approved):
LGTM but probably warrants a changelog entry :v:

brandur:
> Hmm, macOS terminal and Chrome both want to break on the `.` and `-` characters in the above, so unfortunately it's pretty far from double-clickable 😕

OOC, are you actually using macOS terminal? I would've thought most River devs on Mac would be using iTerm, which does allow double-click across hyphens.

It's definitely very system/app dependent though.

brandur:
@bgentry Changed this to be all underscores instead like:

```
miniml2_local_2024_03_07T04_39_12
```

And added changelog.

bgentry:
> Changed this to be all underscores instead like:
> 
> ```
> miniml2_local_2024_03_07T04_39_12
> ```
> 
> And added changelog.

I think that's good at least to preserve double-click selectability in ~all platforms. People can customize this to their heart's desire so we don't need to stress too much about it IMO. They can use pod/container names, dyno names, etc.

brandur:
> I think that's good at least to preserve double-click selectability in ~all platforms. People can customize this to their heart's desire so we don't need to stress too much about it IMO. They can use pod/container names, dyno names, etc.

Yep, 100%. My thoughts too.

PR #257, opened 2024-03-09
Remove Go `x/exp` dependency; using `comparable` instead

brandur:
We were previously using the `x/exp` package for access to constraints
like `constraints.Integer`, but it turns out that we weren't using
anything that wouldn't be better served by Go's built in `comparable`
keyword instead. Here, in the interest of minimizing unnecessary
dependencies, drop `x/exp`.


brandur:
@bgentry Just a small clean up.

bgentry (approved):

PR #258, opened 2024-03-09
Batch completer + additional completer test suite and benchmarks

brandur:
Here, add a new completer using a completion strategy designed to be
much faster than what we're doing right now. Rather than blindly
throwing completion work into goroutine slots, it accumulates "batches"
of completions to be carried out, and using a debounced channel to fire
periodically (currently, up to every 100 milliseconds) and submit entire
batches for completion at once up to 2,000 jobs.

For the purposes of not grossly expanding the `riverdriver` interface,
the completer only batches jobs being set to `completed`, which under
most normal workloads we expect to be the vast common case. Jobs going
to other states are fed into a member `AsyncCompleter`, thereby allowing
the `BatchCompleter` to keeps implementation quite simple.

According to in-package benchmarking, the new completer is in the range
of 3-5x faster than `AsyncCompleter` (the one currently in use by River
client), and 10-15x faster than `InlineCompleter`.

    $ go test -bench=. ./internal/jobcompleter
    goos: darwin
    goarch: arm64
    pkg: github.com/riverqueue/river/internal/jobcompleter
    BenchmarkAsyncCompleter_Concurrency10/Completion-8                 10851            112318 ns/op
    BenchmarkAsyncCompleter_Concurrency10/RotatingStates-8             11386            120706 ns/op
    BenchmarkAsyncCompleter_Concurrency100/Completion-8                 9763            116773 ns/op
    BenchmarkAsyncCompleter_Concurrency100/RotatingStates-8            10884            115718 ns/op
    BenchmarkBatchCompleter/Completion-8                               54916             27314 ns/op
    BenchmarkBatchCompleter/RotatingStates-8                           11518            100997 ns/op
    BenchmarkInlineCompleter/Completion-8                               4656            369281 ns/op
    BenchmarkInlineCompleter/RotatingStates-8                           1561            794136 ns/op
    PASS
    ok      github.com/riverqueue/river/internal/jobcompleter       21.123s

Along with the new completer, we also add a vastly more thorough test
suite to help tease out race conditions and test edges that were
previously being ignored completely. For most cases we drop the heavy
mocking that was happening before, which was having the effect of
minimizing the surface area under test, and producing misleading timing
that wasn't realistic.

Similarly, we bring in a new benchmark framework to allow us to easily
vet and compare completer implementations relative to each other. The
expectation is that this will act as a more synthetic proxy, with the
new benchmarking tool in #254 providing a more realistic end-to-end
measurement.


brandur:
@bgentry Hoping we can bring in #254 so I can run an end-to-end benchmark on this, but results from more synthetic benchmarks are currently looking _very_ good. I suspect that the completer is currently our bottleneck because even with the async completer, it's easy for all 100 errgroup slots to become saturated, and once they do, completion probably isn't that much faster than just completing inline.

brandur:
Actually, I ended up hacking my benchmarker into this branch. 5x+ improvement at >35k jobs/sec on my commodity MBA (compared to 6k jobs/sec previously).

Before:

``` sh
$ go run main.go bench --database-url $DATABASE_URL --duration 1m
bench: jobs worked [          0 ], inserted [      50000 ], job/sec [        0.0 ] [0s]
bench: jobs worked [       3634 ], inserted [       4000 ], job/sec [     1817.0 ] [2s]
bench: jobs worked [       4230 ], inserted [       4000 ], job/sec [     2115.0 ] [2s]
bench: jobs worked [       4238 ], inserted [       4000 ], job/sec [     2119.0 ] [2s]
bench: jobs worked [       4309 ], inserted [       6000 ], job/sec [     2154.5 ] [2s]
bench: jobs worked [       4251 ], inserted [       4000 ], job/sec [     2125.5 ] [2s]
bench: jobs worked [       4246 ], inserted [       4000 ], job/sec [     2123.0 ] [2s]
bench: jobs worked [       4202 ], inserted [       4000 ], job/sec [     2101.0 ] [2s]
bench: jobs worked [       3982 ], inserted [       4000 ], job/sec [     1991.0 ] [2s]
bench: jobs worked [       4055 ], inserted [       4000 ], job/sec [     2027.5 ] [2s]
bench: jobs worked [       4060 ], inserted [       4000 ], job/sec [     2030.0 ] [2s]
bench: jobs worked [       4100 ], inserted [       4000 ], job/sec [     2050.0 ] [2s]
bench: jobs worked [       3985 ], inserted [       4000 ], job/sec [     1992.5 ] [2s]
bench: jobs worked [       4002 ], inserted [       4000 ], job/sec [     2001.0 ] [2s]
bench: jobs worked [       4016 ], inserted [       4000 ], job/sec [     2008.0 ] [2s]
bench: jobs worked [       3834 ], inserted [       4000 ], job/sec [     1917.0 ] [2s]
bench: jobs worked [       3775 ], inserted [       4000 ], job/sec [     1887.5 ] [2s]
bench: jobs worked [       3815 ], inserted [       4000 ], job/sec [     1907.5 ] [2s]
bench: jobs worked [       3617 ], inserted [       4000 ], job/sec [     1808.5 ] [2s]
bench: jobs worked [       9917 ], inserted [      10000 ], job/sec [     4958.5 ] [2s]
bench: jobs worked [       7186 ], inserted [       6000 ], job/sec [     3593.0 ] [2s]
bench: jobs worked [      25564 ], inserted [      26000 ], job/sec [    12782.0 ] [2s]
bench: jobs worked [      24033 ], inserted [      24000 ], job/sec [    12016.5 ] [2s]
bench: jobs worked [      27612 ], inserted [      26000 ], job/sec [    13806.0 ] [2s]
bench: jobs worked [      26952 ], inserted [      28000 ], job/sec [    13476.0 ] [2s]
bench: jobs worked [      27083 ], inserted [      28000 ], job/sec [    13541.5 ] [2s]
bench: jobs worked [      27501 ], inserted [      26000 ], job/sec [    13750.5 ] [2s]
bench: jobs worked [      27636 ], inserted [      28000 ], job/sec [    13818.0 ] [2s]
bench: jobs worked [      27914 ], inserted [      28000 ], job/sec [    13957.0 ] [2s]
bench: jobs worked [      28998 ], inserted [      30000 ], job/sec [    14499.0 ] [2s]
bench: total jobs worked [     366572 ], total jobs inserted [     416000 ], overall job/sec [     6109.5 ], running 1m0.00017125s
```

After:

``` sh
$ go run main.go bench --database-url $DATABASE_URL --duration 1m
bench: jobs worked [          0 ], inserted [      50000 ], job/sec [        0.0 ] [0s]
bench: jobs worked [      50005 ], inserted [      52000 ], job/sec [    25002.5 ] [2s]
bench: jobs worked [      79998 ], inserted [      74000 ], job/sec [    39999.0 ] [2s]
bench: jobs worked [      70013 ], inserted [      76000 ], job/sec [    35006.5 ] [2s]
bench: jobs worked [      92621 ], inserted [      88000 ], job/sec [    46310.5 ] [2s]
bench: jobs worked [      87398 ], inserted [      86000 ], job/sec [    43699.0 ] [2s]
bench: jobs worked [      61233 ], inserted [      66000 ], job/sec [    30616.5 ] [2s]
bench: jobs worked [      90732 ], inserted [      76000 ], job/sec [    45366.0 ] [2s]
bench: jobs worked [      62004 ], inserted [      78000 ], job/sec [    31002.0 ] [2s]
bench: jobs worked [      93999 ], inserted [      86000 ], job/sec [    46999.5 ] [2s]
bench: jobs worked [      83998 ], inserted [      82000 ], job/sec [    41999.0 ] [2s]
bench: jobs worked [      73999 ], inserted [      78000 ], job/sec [    36999.5 ] [2s]
bench: jobs worked [      71581 ], inserted [      76000 ], job/sec [    35790.5 ] [2s]
bench: jobs worked [      76939 ], inserted [      78000 ], job/sec [    38469.5 ] [2s]
bench: jobs worked [      73105 ], inserted [      68000 ], job/sec [    36552.5 ] [2s]
bench: jobs worked [      80354 ], inserted [      86000 ], job/sec [    40177.0 ] [2s]
bench: jobs worked [      70004 ], inserted [      60000 ], job/sec [    35002.0 ] [2s]
bench: jobs worked [      70011 ], inserted [      76000 ], job/sec [    35005.5 ] [2s]
bench: jobs worked [      84107 ], inserted [      84000 ], job/sec [    42053.5 ] [2s]
bench: jobs worked [      65873 ], inserted [      70000 ], job/sec [    32936.5 ] [2s]
bench: jobs worked [      80005 ], inserted [      72000 ], job/sec [    40002.5 ] [2s]
bench: jobs worked [      70009 ], inserted [      74000 ], job/sec [    35004.5 ] [2s]
bench: jobs worked [      80015 ], inserted [      78000 ], job/sec [    40007.5 ] [2s]
bench: jobs worked [      70016 ], inserted [      76000 ], job/sec [    35008.0 ] [2s]
bench: jobs worked [      69993 ], inserted [      70000 ], job/sec [    34996.5 ] [2s]
bench: jobs worked [      80020 ], inserted [      80000 ], job/sec [    40010.0 ] [2s]
bench: jobs worked [      90008 ], inserted [      80000 ], job/sec [    45004.0 ] [2s]
bench: jobs worked [      60011 ], inserted [      70000 ], job/sec [    30005.5 ] [2s]
bench: jobs worked [      70566 ], inserted [      70000 ], job/sec [    35283.0 ] [2s]
bench: jobs worked [      79411 ], inserted [      80000 ], job/sec [    39705.5 ] [2s]
bench: total jobs worked [    2268151 ], total jobs inserted [    2314000 ], overall job/sec [    37802.2 ], running 1m0.000476583s
```

event.go
```diff
@@ -69,7 +69,7 @@ func jobStatisticsFromInternal(stats *jobstats.JobStatistics) *JobStatistics {
 
 // The maximum size of the subscribe channel. Events that would overflow it will
 // be dropped.
-const subscribeChanSize = 100
+const subscribeChanSize = 50_000
```
brandur, line 72: Had to substantially increase this number of the benchmarker since it's now possible to insert 100 jobs to the channel almost instantaneous and even a fast channel listener can lose some.

brandur:
Benchmark settings tweaks discussed in #254 improved these numbers even more — now regularly seeing > 40k jobs/sec with an average ~37k jobs/sec.

brandur:
With the tweaks from #259 these got even better, showing >50k jobs/sec, although oddly the longer the benchmark goes on, the more degraded the numbers become, and this is reproducible. I'm fairly sure that's not related to the completer changes here, but something to take a look into as we try to optimize. Also, the fact that some of these iterations performed _exactly_ 50k jobs/sec is also suspicious and needs more investigation.

``` sh
$ go run main.go bench --database-url $DATABASE_URL --duration 1m
bench: jobs worked [          0 ], inserted [      50000 ], job/sec [        0.0 ] [0s]
bench: jobs worked [      82080 ], inserted [      82000 ], job/sec [    41040.0 ] [2s]
bench: jobs worked [      89920 ], inserted [      90000 ], job/sec [    44960.0 ] [2s]
bench: jobs worked [     100000 ], inserted [      92000 ], job/sec [    50000.0 ] [2s]
bench: jobs worked [      99034 ], inserted [      94000 ], job/sec [    49517.0 ] [2s]
bench: jobs worked [      74966 ], inserted [      88000 ], job/sec [    37483.0 ] [2s]
bench: jobs worked [     100000 ], inserted [     100000 ], job/sec [    50000.0 ] [2s]
bench: jobs worked [      96000 ], inserted [      96000 ], job/sec [    48000.0 ] [2s]
bench: jobs worked [     109049 ], inserted [      94000 ], job/sec [    54524.5 ] [2s]
bench: jobs worked [      70065 ], inserted [      86000 ], job/sec [    35032.5 ] [2s]
bench: jobs worked [      73861 ], inserted [      74000 ], job/sec [    36930.5 ] [2s]
bench: jobs worked [      75025 ], inserted [      74000 ], job/sec [    37512.5 ] [2s]
bench: jobs worked [      74023 ], inserted [      76000 ], job/sec [    37011.5 ] [2s]
bench: jobs worked [      49014 ], inserted [      48000 ], job/sec [    24507.0 ] [2s]
bench: jobs worked [      94971 ], inserted [      94000 ], job/sec [    47485.5 ] [2s]
bench: jobs worked [      59105 ], inserted [      60000 ], job/sec [    29552.5 ] [2s]
bench: jobs worked [      55060 ], inserted [      52000 ], job/sec [    27530.0 ] [2s]
bench: jobs worked [      48911 ], inserted [      50000 ], job/sec [    24455.5 ] [2s]
bench: jobs worked [      47193 ], inserted [      48000 ], job/sec [    23596.5 ] [2s]
bench: jobs worked [      45858 ], inserted [      46000 ], job/sec [    22929.0 ] [2s]
bench: jobs worked [      41228 ], inserted [      42000 ], job/sec [    20614.0 ] [2s]
bench: jobs worked [      35031 ], inserted [      34000 ], job/sec [    17515.5 ] [2s]
bench: jobs worked [      29986 ], inserted [      30000 ], job/sec [    14993.0 ] [2s]
bench: jobs worked [      29058 ], inserted [      28000 ], job/sec [    14529.0 ] [2s]
bench: jobs worked [      27033 ], inserted [      30000 ], job/sec [    13516.5 ] [2s]
bench: jobs worked [      29033 ], inserted [      28000 ], job/sec [    14516.5 ] [2s]
bench: jobs worked [      26030 ], inserted [      24000 ], job/sec [    13015.0 ] [2s]
bench: jobs worked [      28020 ], inserted [      30000 ], job/sec [    14010.0 ] [2s]
bench: jobs worked [      25026 ], inserted [      26000 ], job/sec [    12513.0 ] [2s]
bench: jobs worked [      27019 ], inserted [      26000 ], job/sec [    13509.5 ] [2s]
bench: total jobs worked [    1767497 ], total jobs inserted [    1818000 ], overall job/sec [    29457.5 ], running 1m0.001600541s
```

event.go
bgentry: I was wondering earlier if this would be an issue with the way you were now measuring. It’s unlikely someone would hit this in practice but I suppose it’s still a bit of a wart with the current fixed size buffered channel for subscriptions.

bgentry:
`1 / .02 * 1000 = 50,000`, the exact max theoretical throughput when you’re fetching 1000 jobs every 2ms 😁

brandur:
> `1 / .02 * 1000 = 50,000`, the exact max theoretical throughput when you’re fetching 1000 jobs every 2ms 😁

Ah dang it. Good call :)

event.go
```diff
@@ -69,7 +69,7 @@ func jobStatisticsFromInternal(stats *jobstats.JobStatistics) *JobStatistics {
 
 // The maximum size of the subscribe channel. Events that would overflow it will
 // be dropped.
-const subscribeChanSize = 100
+const subscribeChanSizeDefault = 50_000
```
brandur, line 72: Okay, so unfortunately, it seems that Go eagerly allocates channel size [1], and this 50k number is probably way more than most uses will ever need. However, we do need a number at least this high for benchmarking (50k is only ~1s worth of benching which you could imagine falling behind).

@bgentry Thoughts on possibly exposing a second `client.Subscribe` variant that allows a channel size to be injected? (Or actually, it might take a `SubscribeConfig` object just for better future compatibility.) Not perfect, but it may be the best option.

I already had to add an internal subscribe variant so I could fix the test case that checked that subscribe channels can overflow.

---

[1] https://stackoverflow.com/questions/65417280/how-and-when-does-go-allocate-memory-for-bounded-queue-channels

brandur: BTW, I went ahead and did this (add a second `Subscribe` function that takes `SubscribeConfig`). Feels pretty clean IMO and resolves this problem. Also bumped the default size to 1_000 which feels a little more reasonable.

cmd/river/go.mod
```diff
@@ -2,19 +2,20 @@ module github.com/riverqueue/river/cmd/river
 
 go 1.21.4
 
-// replace github.com/riverqueue/river => ../..
+replace github.com/riverqueue/river => ../..
```
brandur, line 5: Have to temporarily uncomment the `replace` because the River API changed. I'll unwind this as I'm cutting a release.

brandur:
I made a last minute change to this one and accidentally introduced a context bug in the client, but I think I've got it squashed now.

internal/jobcompleter/job_completer.go
```diff
@@ -56,125 +64,290 @@ type InlineJobCompleter struct {
...
-	})
+	}
+
+	// Use a single `finalized_at` value for the whole batch. Not the greatest
+	// thing maybe, but makes things much easier.
+	var sampleFinalizedAt time.Time
+	for _, setState := range setStateBatch {
+		sampleFinalizedAt = *setState.Params.FinalizedAt
+		break
+	}
```
bgentry, line 261: Unfortunately I think this change is a pretty substantial degradation. With this PR we're now debouncing up to 100ms, meaning jobs could have up to 100ms of randomly distributed padding added to their `finalized_at`, making it so `attempted_at - finalized_at` is no longer representative of their duration (at least not for reasonably fast jobs).

I'm sure it will be a bit less efficient, but I think we'd be better off keeping a `finalized_at` for each job as reported at the time it finished executing. Then your `JobSetCompleteIfRunningManyParams` can be a list of two arrays: the IDs, and the corresponding timestamps for each.

I also wanted to take this moment to raise another question. As I've progressed on the UI, I don't really love the limitations of trying to deduce job statistics from the handful of available timestamps (`inserted_at`, `attempted_at`, `finalized_at`, and those in `errors`). It's too complicated and leaves odd undefined gaps. I can paper over some of those at the UI level, but it has held me back from some stuff I'd like to ship like a more reliable execution timeline.

I've started wondering whether it'd be better if we tracked some sort of job stats object that we could keep as an array on the `JobRow` and as a jsonb array in Postgres. Possibly it could be merged together with `errors` so there's a single large jsonb ala `executions` or `attempts`. It would allow for some nice built-in stats even without hooking up an external monitoring system to consume all job stats in real time. I'm not sure how this will impact performance in the happy path though, because you'd then always be setting/updating a jsonb TOAST value as well as setting a timestamp.

Thoughts on both these fronts?

brandur: > Unfortunately I think this change is a pretty substantial degradation. With this PR we're now debouncing up to 100ms, meaning jobs could have up to 100ms of randomly distributed padding added to their `finalized_at`, making it so `attempted_at - finalized_at` is no longer representative of their duration (at least not for reasonably fast jobs).

Fair enough. I just pushed up a change that also sets `finalized_at` specifically for each job, and a new set of tests.

> I also wanted to take this moment to raise another question. As I've progressed on the UI, I don't really love the limitations of trying to deduce job statistics from the handful of available timestamps (`inserted_at`, `attempted_at`, `finalized_at`, and those in `errors`). It's too complicated and leaves odd undefined gaps. I can paper over some of those at the UI level, but it has held me back from some stuff I'd like to ship like a more reliable execution timeline.

OOC, if you excluded errors/snoozes, are there any particular timings you were thinking of that you can't calculate using the existing fields?

Seems plausible, although I'd definitely try top get away with either (1) consolidating to one jsonb field, or (2) stuff more stats into the `errors` field that we already have. Having to look at two fields that may have a different number of elements and try to match them up against each sounds awful.

I'll also give my typical response that jsonb isn't very ergonomic. It's harder to use in queries, it's more obfuscated in query results, and you don't get autocomplate on names in psql. So as I mentioned with putting cancels in metadata, IMO the less we have of it the better. We'd have to put stats for retries in jsonb, but if we could get away with it only for errors, that might be better.

Ditching the name `errors` should sure make things easier for use in Sequel/ActiveRecord in Ruby land though. The hacks needed to work around this naming are kind of crazy.

client.go
```diff
@@ -656,18 +677,21 @@ func (c *Client[TTx]) signalStopComplete(ctx context.Context) {
...
+		//
+		// TODO: there's a risk here that the completer is stuck on a job that
+		// won't complete. We probably need a timeout or way to move on in those
+		// cases.
+		c.services,
 
-	c.queueMaintainer.Stop()
+		// Will only be started if this client was leader, but can tolerate a stop
+		// without having been started.
+		c.queueMaintainer,
```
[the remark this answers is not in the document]
bgentry: oh yeah, there's probably no reason for this not to stop earlier. Well, maybe it should wait until after the elector has given up any leadership it may hold? 🤔 I guess it's fine if not, worst case they skip a cycle that's about to happen but somebody else will pick it up, right? Is that true of all maintenance services?

client.go
```diff
@@ -762,7 +785,41 @@ func (c *Client[TTx]) Stopped() <-chan struct{} {
...
+	// If new event kinds are added, callers will have to explicitly add them to
+	// their requested list and esnure they can be handled correctly.
+	Kinds []EventKind
+}
+
+// Special internal variant that lets us inject an overridden size.
+func (c *Client[TTx]) SubscribeConfig(config *SubscribeConfig) (<-chan *Event, func()) {
+	if config.ChanSize < 0 {
+		panic("SubscribeConfig.ChanSize must be greater or equal to 1")
+	}
```
bgentry, line 817: I guess we don't care about max? if you want to be dumb go right ahead...?

cmd/river/riverbench/river_bench.go
```diff
@@ -132,13 +145,30 @@ func (b *Benchmarker[TTx]) Run(ctx context.Context) error {
...
+				switch {
+				case event.Kind == river.EventKindJobCancelled:
+					fallthrough
+
+				case event.Kind == river.EventKindJobCompleted:
+					fallthrough
+
+				// Only count a job as complete if it failed for the last time.
+				// We don't expect benchmark jobs to ever fail, so this extra
+				// attention to detail is here, but shouldn't be needed.
```
bgentry, line 163: should we count and log these for debug purposes?

bgentry (commented):
I have just the completer and its tests left to review almost there 🔜  Looking great now.

client.go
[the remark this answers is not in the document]
brandur: Yeah, my rationale is basically: the maintenance services are going to be spending most of their time in wait loops anyway, and we expect all maintenance services to be spending most of their time in waits between loops under typical operation. Shutdown should take < 1s once it hits this point, so a new elector and maintenance services should come up quite quickly.

The only service that comes to mind is the periodic job enqueuer because it's essentially stateless now. If it was 59 minutes into about to working a once an hour job, that wait will have to start again on a client elsewhere. But that said, the difference in stopping the queue maintainer a second later will be negligible.

The queue maintainer is a constellation of ~5 services so I did find that having it start to stop in parallel along with these other services did help somewhat in making the top level client tests a little faster in aggregate (since the client is stopped on ~every test). I think overall the tiny tradeoff (if there is one even), is worth it.

cmd/river/riverbench/river_bench.go
brandur: Sure. Added this as an error log since it should never, ever happen. The errors will be visible even without passing the verbose flag.

client.go
brandur: Yeah, not sure it makes sense to try and enforce a cap. Any Go user can try to `make(chan struct{}, 1_000_000_000_000)` anytime they want.

internal/jobcompleter/job_completer.go
```diff
@@ -56,125 +63,299 @@ type InlineJobCompleter struct {
...
+func (c *AsyncCompleter) JobSetStateIfRunning(ctx context.Context, stats *jobstats.JobStatistics, params *riverdriver.JobSetStateIfRunningParams) error {
+	// Start clock outside of goroutine so that the time spent blocking waiting
+	// for an errgroup slot is accurately measured.
+	start := c.TimeNowUTC()
 
-		job, err := withRetries(&c.BaseService, f)
+	c.errGroup.Go(func() error {
+		job, err := withRetries(ctx, &c.BaseService, func(ctx context.Context) (*rivertype.JobRow, error) {
+			return c.exec.JobSetStateIfRunning(ctx, params)
+		})
```
bgentry, line 136: Hmm, is this an appropriate context to be passing down and respecting the deadline on? Isn't this `ctx` the job's execution context, which may be cancelled (including by a remote cancellation attempt)?

I feel like the only place it would make sense to be used is in the inline completer. The async one should probably be started with a Client-derived context that is in no way tied to the job's execution.

internal/jobcompleter/job_completer.go
```diff
@@ -56,125 +63,299 @@ type InlineJobCompleter struct {
...
 
 		stats.CompleteDuration = c.TimeNowUTC().Sub(start)
+		c.sendJobToSubscription(job, stats)
+
+		return nil
+	})
+	return nil
+}
+
+func (c *AsyncCompleter) Start(ctx context.Context) error { return nil }
```
bgentry, line 149: This context is unused, but do you think it should maybe in some way be a parent to any completion attempts in here? Actually, I'm second-guessing myself in realtime. Even in the case of an aggressive shutdown, we still don't want to shut down until the completer has finished its work. So IMO these completion attempts should not inherit any deadlines whatsoever from parent contexts.

internal/jobcompleter/job_completer.go
```diff
@@ -56,125 +63,299 @@ type InlineJobCompleter struct {
...
+		}
+		return batchIDs, batchFinalizedAt
+	}
+
+	// Tease apart enormous batches into sub-batches.
+	//
+	// All the code below is concerned with doing that, with a fast loop that
+	// doesn't allocate any additional memory in case the entire batch is
+	// smaller than the sub-batch maximum size (which will be the common case).
+	const oneOperationMax = 2_000
```
bgentry, line 286: Is it preferable to wait until the end of the timer window once we have enough jobs to fill a batch, or should we dispatch it ~immediately when we hit this threshold and reset the timer?

Doesn't have to be part of this PR, but might be worth refactoring if you think it could be straightforward to do so. IMO if we already have enough jobs to fill a batch, may as well get rid of them ASAP and avoid just stacking up more jobs until we hit the timer.

Maybe it ends up being harder to manage multiple in-flight batches and isn't worth it as a result.

internal/jobcompleter/job_completer.go
```diff
@@ -56,125 +63,299 @@ type InlineJobCompleter struct {
...
+func (c *AsyncCompleter) Stop() {
+	if err := c.errGroup.Wait(); err != nil {
+		c.Logger.Error("Error waiting on async completer: %s", err)
+	}
+}
+
+type batchCompleterSetState struct {
+	Params    *riverdriver.JobSetStateIfRunningParams
+	Stats     *jobstats.JobStatistics
+	WaitingAt time.Time // went job was submitted for completion
```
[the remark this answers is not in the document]
bgentry: Actually, is this still being set anywhere after the per-job timestamp refactor? Maybe you can delete the field?

internal/jobcompleter/job_completer.go
```diff
@@ -56,125 +63,299 @@ type InlineJobCompleter struct {
...
+			select {
+			case <-stopCtx.Done():
+				// Try to insert last batch before leaving. Note we use the
+				// original context so operations aren't immediately cancelled.
+				if err := c.handleBatch(ctx); err != nil {
+					c.Logger.Error(c.Name+": Error completing batch", "err", err)
+				}
+
+				return
+			case <-c.debounceChan.C():
```
bgentry, line 220: Doesn't this `DebouncedChan` type fire on the leading edge? So that means we will always submit a batch of 1 before then waiting to collect additional jobs? Doesn't seem ideal if so.

internal/jobcompleter/job_completer.go
bgentry, line 286: Maybe you want to make this a configurable field on the completer type? I guess we don't have any batch completer specific testing atm, but it might make such tests easier to write.

internal/jobcompleter/job_completer.go
```diff
@@ -56,125 +63,299 @@ type InlineJobCompleter struct {
...
+	}()
+
+	if len(setStateBatch) < 1 {
 		return nil
-	})
+	}
+
+	// Insert a sub-batch with retries. Also helps reduce visual noise and
+	// increase readability of loop below.
+	insertSubBatch := func(batchID []int64, batchFinalizedAt []time.Time) ([]*rivertype.JobRow, error) {
```
bgentry, line 256: Slightly misnamed IMO because it's not _inserting_ anything. Maybe submit, update, or handle sub batch?

internal/jobcompleter/job_completer.go
```diff
@@ -56,125 +63,299 @@ type InlineJobCompleter struct {
...
+		jobRows                   []*rivertype.JobRow
+	)
+	if len(setStateBatch) > oneOperationMax {
+		jobRows = make([]*rivertype.JobRow, 0, len(setStateBatch))
+		for i := 0; i < len(setStateBatch); i += oneOperationMax {
+			endIndex := min(i+oneOperationMax, len(batchID)) // beginning of next sub-batch or end of slice
+			jobRowsSubBatch, err := insertSubBatch(batchID[i:endIndex], batchFinalizedAt[i:endIndex])
+			if err != nil {
+				return err
+			}
```
bgentry, line 299: so if a single sub batch fails after 4 attempts, we immediately return an error and give up on all remaining jobs in the batch? That feels a bit risky 🤔 

Also is there any kind of backpressure to stop the client from fetching more jobs if the completer is filling up and can't complete all the jobs it already has lined up?

internal/jobcompleter/job_completer.go
```diff
@@ -56,125 +63,299 @@ type InlineJobCompleter struct {
...
+		return c.asyncCompleter.JobSetStateIfRunning(ctx, stats, params)
+	}
 
-	c.subscribeFunc = subscribeFunc
+	// Wait until the completer is started and ready to start processing
+	// batches. Alternatively, we could remove this and allow batches to start
+	// accumulating even if the service isn't started, but that could introduce
+	// some danger of the service never being started and therefore accumulating
+	// forever without completing jobs.
+	<-c.ready
```
bgentry, line 332: Can we instead make sure the completer is started & good to go before we even start working jobs? Seems better than having this channel receive on every single completion IMO. And the slight increase in startup time also seems worth it to be as sure as possible that jobs are ready to be worked successfully.

internal/jobcompleter/job_completer.go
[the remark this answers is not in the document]
brandur: Yep, actually was able to be removed completely.

internal/jobcompleter/job_completer.go
brandur: So actually I ended up putting in a different constant (min job threshold = 100) which the completer will continue to pool jobs to complete across timer ticks, and then signficantly decreasing the tick interval. So when many jobs are coming in quickly for a completion like on benchmarking, the completer notices sooner and starts inserting right away (it'll also continue inserting without waiting for a tick if we're still above the threshold on completing a batch), but when few jobs are coming in it's still able to batch reasonably well. Every five tick loops we force a completion of the batch, so the longest any job will have to wait is 250 ms, even in an environment where little is happening.

internal/jobcompleter/job_completer.go
brandur: True.

So this one is a bit of a story -- I modified `DebounceChan` to take a boolean in its initializer for `sendLeaading` for use in cases like this, and added a test.

_But_, then I modified the loop with the job threshold mentioned above, and since it's possible for the loop to skip completing a batch if not enough jobs have accumulated yet, the use of `DebounceChan` doesn't make sense anymore. Instead, it should just be a normal `time.Ticker`.

Changed it and it works fine. I left the `sendLeading` option in `DebounceChan` mentioned above since I'd already written a test case for it, but I guess this could be reverted.

internal/jobcompleter/job_completer.go
brandur: I ended up extracting this constant and putting it on the struct for testability, and added a few specific tests for the batch completer to exercise these edges.

internal/jobcompleter/job_completer.go
brandur: Renamed to `completeSubBatch`.

internal/jobcompleter/job_completer.go
brandur: So I ended up putting in a backpressure system, wherein the batch completer blocks are receiving more jobs if the backlog is too large, and added a batch completer test to try and exercise it specifically.

Hitting maximum backlog produces a warning log line, and I've found this to be very helpful for benchmarking purposes. Previously a backlogged completer was swept under the rug because infinite backlog was allowed, so a maximum backlog combined with backpressure forces the benchmark to pause until jobs can clear.

In terms of giving up after four attempts: I think unfortunately there's only so much we can do there. If the completer's failed four times, then likely the database is in a totally degenerate state or the client's lost its database connection, either of which are going to bubble up and create additional carnage in places other than completer. This is better now with the backpressure system though -- if the completer's having trouble getting a completion through, jobs accumulate until the backlog gets hit, and the whole client gets effectively paused until jobs can start completing again. I've observed this in my benchmarks from time to time as my Postgres degrades in odd ways.

internal/jobcompleter/job_completer.go
brandur: Yep, changed that. And actually, we only bother to fully wait for startup in the benchmark tests, where it's most appropriate to mandate a fully ready completer.

brandur:
Made a couple improvements and large changes, so here's a repost of some benchmark runs. One thing I found too is that I'm able to get more consistent results when operating the benchmark in "burn down" mode where we insert all jobs before starting.

* Before: ~10k jobs/sec
* After: ~46k jobs/sec

Run before changes:

``` sh
$ go build . && ./river bench --database-url $DATABASE_URL -n 1_000_000
bench: jobs worked [       9087 ], inserted [          0 ], job/sec [     4543.5 ] [2s]
bench: jobs worked [      10453 ], inserted [          0 ], job/sec [     5226.5 ] [2s]
bench: jobs worked [       8310 ], inserted [          0 ], job/sec [     4155.0 ] [2s]
bench: jobs worked [      11418 ], inserted [          0 ], job/sec [     5709.0 ] [2s]
bench: jobs worked [      10875 ], inserted [          0 ], job/sec [     5437.5 ] [2s]
bench: jobs worked [      10899 ], inserted [          0 ], job/sec [     5449.5 ] [2s]
bench: jobs worked [      10768 ], inserted [          0 ], job/sec [     5384.0 ] [2s]
bench: jobs worked [      11502 ], inserted [          0 ], job/sec [     5751.0 ] [2s]
bench: jobs worked [      11247 ], inserted [          0 ], job/sec [     5623.5 ] [2s]
bench: jobs worked [      10870 ], inserted [          0 ], job/sec [     5435.0 ] [2s]
bench: jobs worked [      11146 ], inserted [          0 ], job/sec [     5573.0 ] [2s]
bench: jobs worked [      10862 ], inserted [          0 ], job/sec [     5431.0 ] [2s]
bench: jobs worked [      11006 ], inserted [          0 ], job/sec [     5503.0 ] [2s]
bench: jobs worked [      10604 ], inserted [          0 ], job/sec [     5302.0 ] [2s]
bench: jobs worked [       8877 ], inserted [          0 ], job/sec [     4438.5 ] [2s]
bench: jobs worked [      10624 ], inserted [          0 ], job/sec [     5312.0 ] [2s]
bench: jobs worked [      10843 ], inserted [          0 ], job/sec [     5421.5 ] [2s]
bench: jobs worked [      16371 ], inserted [          0 ], job/sec [     8185.5 ] [2s]
bench: jobs worked [      18023 ], inserted [          0 ], job/sec [     9011.5 ] [2s]
bench: jobs worked [      21967 ], inserted [          0 ], job/sec [    10983.5 ] [2s]
bench: jobs worked [      29516 ], inserted [          0 ], job/sec [    14758.0 ] [2s]
bench: jobs worked [      27671 ], inserted [          0 ], job/sec [    13835.5 ] [2s]
bench: jobs worked [      28273 ], inserted [          0 ], job/sec [    14136.5 ] [2s]
bench: jobs worked [      31026 ], inserted [          0 ], job/sec [    15513.0 ] [2s]
bench: jobs worked [      30648 ], inserted [          0 ], job/sec [    15324.0 ] [2s]
bench: jobs worked [      31054 ], inserted [          0 ], job/sec [    15527.0 ] [2s]
bench: jobs worked [      28779 ], inserted [          0 ], job/sec [    14389.5 ] [2s]
bench: jobs worked [      27188 ], inserted [          0 ], job/sec [    13594.0 ] [2s]
bench: jobs worked [      24629 ], inserted [          0 ], job/sec [    12314.5 ] [2s]
bench: jobs worked [      19879 ], inserted [          0 ], job/sec [     9939.5 ] [2s]
bench: jobs worked [      25347 ], inserted [          0 ], job/sec [    12673.5 ] [2s]
bench: jobs worked [      27751 ], inserted [          0 ], job/sec [    13875.5 ] [2s]
bench: jobs worked [      28440 ], inserted [          0 ], job/sec [    14220.0 ] [2s]
bench: jobs worked [      29316 ], inserted [          0 ], job/sec [    14658.0 ] [2s]
bench: jobs worked [      28189 ], inserted [          0 ], job/sec [    14094.5 ] [2s]
bench: jobs worked [      29762 ], inserted [          0 ], job/sec [    14881.0 ] [2s]
bench: jobs worked [      27336 ], inserted [          0 ], job/sec [    13668.0 ] [2s]
bench: jobs worked [      26620 ], inserted [          0 ], job/sec [    13310.0 ] [2s]
bench: jobs worked [      27051 ], inserted [          0 ], job/sec [    13525.5 ] [2s]
bench: jobs worked [      26063 ], inserted [          0 ], job/sec [    13031.5 ] [2s]
bench: jobs worked [      25188 ], inserted [          0 ], job/sec [    12594.0 ] [2s]
bench: jobs worked [      27466 ], inserted [          0 ], job/sec [    13733.0 ] [2s]
bench: jobs worked [      27301 ], inserted [          0 ], job/sec [    13650.5 ] [2s]
bench: jobs worked [      26436 ], inserted [          0 ], job/sec [    13218.0 ] [2s]
bench: jobs worked [      25720 ], inserted [          0 ], job/sec [    12860.0 ] [2s]
bench: jobs worked [      25168 ], inserted [          0 ], job/sec [    12584.0 ] [2s]
bench: jobs worked [      26160 ], inserted [          0 ], job/sec [    13080.0 ] [2s]
bench: jobs worked [      23750 ], inserted [          0 ], job/sec [    11875.0 ] [2s]
bench: jobs worked [       2519 ], inserted [          0 ], job/sec [     1259.5 ] [2s]
bench: total jobs worked [    1000000 ], total jobs inserted [    1000000 ], overall job/sec [    10204.1 ], running 1m38.00023s
```

And two runs from after the changes, showing roughly similar results:

``` sh
$ go build . && ./river bench --database-url $DATABASE_URL -n 1_000_000
bench: jobs worked [      88058 ], inserted [          0 ], job/sec [    44029.0 ] [2s]
bench: jobs worked [      94047 ], inserted [          0 ], job/sec [    47023.5 ] [2s]
bench: jobs worked [      90256 ], inserted [          0 ], job/sec [    45128.0 ] [2s]
bench: jobs worked [      91427 ], inserted [          0 ], job/sec [    45713.5 ] [2s]
bench: jobs worked [     100908 ], inserted [          0 ], job/sec [    50454.0 ] [2s]
bench: jobs worked [      94504 ], inserted [          0 ], job/sec [    47252.0 ] [2s]
bench: jobs worked [      81588 ], inserted [          0 ], job/sec [    40794.0 ] [2s]
bench: jobs worked [      94824 ], inserted [          0 ], job/sec [    47412.0 ] [2s]
bench: jobs worked [      87340 ], inserted [          0 ], job/sec [    43670.0 ] [2s]
bench: jobs worked [      92049 ], inserted [          0 ], job/sec [    46024.5 ] [2s]
bench: jobs worked [      84999 ], inserted [          0 ], job/sec [    42499.5 ] [2s]
bench: total jobs worked [    1000000 ], total jobs inserted [    1000000 ], overall job/sec [    45851.7 ], running 21.809460917s
```

``` sh
$ go build . && ./river bench --database-url $DATABASE_URL -n 1_000_000
bench: jobs worked [      82657 ], inserted [          0 ], job/sec [    41328.5 ] [2s]
bench: jobs worked [      96057 ], inserted [          0 ], job/sec [    48028.5 ] [2s]
bench: jobs worked [      89829 ], inserted [          0 ], job/sec [    44914.5 ] [2s]
bench: jobs worked [      96847 ], inserted [          0 ], job/sec [    48423.5 ] [2s]
bench: jobs worked [      96042 ], inserted [          0 ], job/sec [    48021.0 ] [2s]
bench: jobs worked [      87198 ], inserted [          0 ], job/sec [    43599.0 ] [2s]
bench: jobs worked [      96474 ], inserted [          0 ], job/sec [    48237.0 ] [2s]
bench: jobs worked [      94126 ], inserted [          0 ], job/sec [    47063.0 ] [2s]
bench: jobs worked [      85323 ], inserted [          0 ], job/sec [    42661.5 ] [2s]
bench: jobs worked [      94043 ], inserted [          0 ], job/sec [    47021.5 ] [2s]
bench: jobs worked [      81387 ], inserted [          0 ], job/sec [    40693.5 ] [2s]
bench: total jobs worked [    1000000 ], total jobs inserted [    1000000 ], overall job/sec [    45753.1 ], running 21.856442959s
```

internal/baseservice/base_service.go
```diff
@@ -143,24 +143,27 @@ func (s *BaseService) CancellableSleepRandomBetweenC(ctx context.Context, sleepD
...
-}
-
-func (s *BaseService) exponentialBackoffSeconds(attempt, maxAttemptsBeforeReset int) float64 {
+// ExponentialBackoff returns a duration for a reasonable exponential backoff
+// interval for a service based on the given attempt number, which can then be
+// fed into CancellableSleep to perform the sleep. Uses a 2**N second algorithm,
+// +/- 10% random jitter. Sleep is cancelled if the given context is cancelled.
+//
+// Attempt should start at one for the first backoff/failure.
+func (s *BaseService) ExponentialBackoff(attempt, maxAttemptsBeforeReset int) time.Duration {
```
brandur, line 152: I retooled this a bit so it returns a duration instead of doing the sleep because it makes it easier to include the sleep duration in log statements which is pretty nice.

internal/riverinternaltest/riverinternaltest.go
```diff
@@ -58,7 +62,20 @@ func BaseServiceArchetype(tb testing.TB) *baseservice.Archetype {
...
-		Rand:       randutil.NewCryptoSeededConcurrentSafeRand(),
+		Rand:       rand,
+		TimeNowUTC: func() time.Time { return time.Now().UTC() },
+	}
+}
+
+// BaseServiceArchetypeDebug returns a new base service suitable for use in
+// tests with debug logging enabled. Useful for swapping into tests for
+// debugging, where additional context is often useful.
+func BaseServiceArchetypeDebug(tb testing.TB) *baseservice.Archetype {
```
brandur, line 73: This is kind of nice for dropping into test cases you're having trouble with for extra logging.

client.go
```diff
@@ -271,10 +271,10 @@ type Client[TTx any] struct {
 	driver    riverdriver.Driver[TTx]
 	elector   *leadership.Elector
 
-	// fetchNewWorkCancel cancels the context used for fetching new work. This
+	// fetchWorkCancel cancels the context used for fetching new work. This
 	// will be used to stop fetching new work whenever stop is initiated, or
 	// when the context provided to Run is itself cancelled.
-	fetchNewWorkCancel context.CancelCauseFunc
+	fetchWorkCancel context.CancelCauseFunc
```
brandur, line 277: This was alternatively named `fetchWork` versus `fetchNewWork` depending on the place, so here just standardized on the former since it's a little more succinct and the naming still describes it nicely.

cmd/river/go.mod
```diff
@@ -23,10 +24,9 @@ require (
 	github.com/jackc/pgpassfile v1.0.0 // indirect
 	github.com/jackc/pgservicefile v0.0.0-20221227161230-091c0ba34f0a // indirect
 	github.com/jackc/puddle/v2 v2.2.1 // indirect
-	github.com/oklog/ulid/v2 v2.1.0 // indirect
+	github.com/lmittmann/tint v1.0.4 // indirect
```
brandur, line 27: So ... I brought in this package for use with the benchmarker, which is a colorized log handler for slog. I was debugging some benchmark problems and this makes things _far_ easier to read because the default Go output is such absolute garbage.

The nice thing about Tint is that it itself has zero dependencies, and we only need it for the River CLI package without having to add it to the top level one, so I think it's okay.

Here's a screenshot on the benchmarker running in verbose (I also added a `--debug` that makes it even more verbose):

image

bgentry: Looks solid! I think I'm gonna pull this into my riverdemo app too.

bgentry (approved):
Sorry for the slow review, but this looks great. I had just a couple questions and a concern about leaking a pgx-specific puddle error outside the driver abstraction.

I would also like to talk about how we can extend this to support other job states. For example, snooze seems like a good candidate to make use of this optimization as well. We can do that out of band from this PR though!

client.go
```diff
@@ -612,19 +612,40 @@ func (c *Client[TTx]) Start(ctx context.Context) error {
...
+	// We use separate contexts for fetching and working to allow for a graceful
+	// stop. However, both inherit from the provided context so if it is
+	// cancelled a more aggressive stop will be initiated.
+	fetchCtx, fetchWorkCancel := context.WithCancelCause(ctx)
+	c.fetchWorkCancel = fetchWorkCancel
+	workCtx, workCancel := context.WithCancelCause(withClient[TTx](ctx, c))
+	c.workCancel = workCancel
 
+	for _, service := range c.services {
+		// TODO(brandur): Reevaluate the use of fetchNewWorkCtx here. It's
```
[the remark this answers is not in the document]
brandur: Ah good catch, thx.

cmd/river/riverbench/river_bench.go
```diff
@@ -94,17 +96,17 @@ func (b *Benchmarker[TTx]) Run(ctx context.Context) error {
 		// values against the wall, they perform quite well. Much better than
 		// the client's default values at any rate.
 		FetchCooldown:     2 * time.Millisecond,
-		FetchPollInterval: 5 * time.Millisecond,
+		FetchPollInterval: 20 * time.Millisecond,
```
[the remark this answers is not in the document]
brandur: Yeah ... I still wouldn't commit to any particular numbers because I can see fairly wild swings in performance between runs, but I played with these numbers more and I think these ones gave me slightly better results.

internal/jobcompleter/job_completer.go
```diff
@@ -56,170 +60,468 @@ type InlineJobCompleter struct {
...
+
+			for {
+				if err := c.handleBatch(ctx); err != nil {
+					c.Logger.Error(c.Name+": Error completing batch", "err", err)
+				}
+
+				// New jobs to complete may have come in while working the batch
+				// above. If enough have to bring us above the minimum complete
+				// threshold, do another again. Otherwise, break and listen for
+				// a new tick.
```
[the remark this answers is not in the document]
brandur: Fixed.

internal/jobcompleter/job_completer.go
```diff
@@ -56,170 +60,468 @@ type InlineJobCompleter struct {
...
+	c.waitOnBacklogChan = make(chan struct{})
+	c.waitOnBacklogWaiting = true
+	c.Logger.WarnContext(ctx, c.Name+": Hit maximum backlog; completions will wait until below threshold", "max_backlog", c.maxBacklog)
+}
 
-		ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
-		defer cancel()
+// As configued, total time from initial attempt is ~7 seconds (1 + 2 + 4) (not
+// including jitter).
+const numRetries = 3
```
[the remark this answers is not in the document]
brandur: Ugh, so this is really tricky to get right. It's 7 seconds, but if the loop is timing out every time (timeouts being a likely failure candidate), the total time is more like 37 seconds (7 seconds + 3 * 10 seconds), which is a really long time. I modified the comment above to be more explicit about that case.

This can be bad because the completer can still block a client shutdown, and blocking it for this length of time could be a serious problem. I've actually seen this happen to slightly smaller extents when a bad failure happens in the completer running a benchmark. I'll hit Ctrl+C to stop it, and it takes ~10 seconds to exit.

I think I'm going to punt on more changes on this in favor of a separate project thinking more about these retries. I'm not convinced that we're doing the right thing at all when it comes to this stuff. I think retrying once or _maybe_ twice makes a certain amount of sense, but given these are likely indicative of something very degenerate happening in the DB, continuing to beat a dead horse by retrying over and over again is likely to lead to more failures and more unnecessary database contention. Even the retry schedule is probably wrong — the first retry should probably be happening even sooner after an initial failure because in the event of a truly intermittent failure, even 1s is quite a long time.

internal/jobcompleter/job_completer.go
```diff
@@ -56,170 +60,468 @@ type InlineJobCompleter struct {
...
+
+		retVal, err := retryFunc(ctx) //nolint:contextcheck
 		if err != nil {
+			// A cancelled context will never succeed, return immediately.
+			if errors.Is(err, context.Canceled) {
+				return defaultVal, err
+			}
+
+			// A closed pool will never succeed, return immediately.
+			if errors.Is(err, puddle.ErrClosedPool) {
```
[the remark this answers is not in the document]
brandur: Good catch. I pushed this down a layer into the driver and added a test.

brandur:
> I would also like to talk about how we can extend this to support other job states. For example, snooze seems like a good candidate to make use of this optimization as well. We can do that out of band from this PR though!

Awesome. Yeah, I think adding support for other states would be great, and actually, probably reasonably doable even from a single query if we put our mind to it. This PR's touched files became more widespread than I would've liked though, so anxious to get it in. +1 looking into the other states out of band.

brandur:
Thanks!

PR #262, opened 2024-03-10
Make more client services into `startstop.Service`

brandur:
Here, we aim to move in a direction where the client can treat most of
its subservices generically by starting/stopping the notifier as a
generic service, along with converting three other smaller services over
to use the `startstop.Service` interface:

* Client monitor. (This ends up cleaning up its internal code
  significantly, and has a major improvement on its test coverage and
  robustness.)

* The client's statistics logging loop.

* The client's leadership change handling loop.

Both the latter two are very small functions, so I implement the new
function `startstop.StartStopFunc` which similar to `river.WorkFunc`,
provides an easy way to get a small service using only a function, which
keeps the code quite clean.

It's going to take a few more changes to get our full simplification
change, but you can already see the beginnings of it as many services
can be started and stopped in ~one line:

    func (c *Client[TTx]) signalStopComplete(ctx context.Context) {
        // Wait for producers and elector to exit:
        c.wg.Wait()

        // Stop all mainline services where stop order isn't important.
        startstop.StopAllParallel(c.services)

By the end of it, both the elector and producers will also become
services, and we'll be able to fully eliminate the wait group, along
with the stop channels and the roundabout stop functions like
`signalStopComplete`. The idea is that the client itself will also
become a start/stop service, and itself become protected against being
started multiple times, or potential races where `Start`/`Stop` may be
called by multiple goroutines.

### DIFF
--- a/client.go
+++ b/client.go
@@ -280,6 +280,7 @@ type Client[TTx any] struct {
 	notifier             *notifier.Notifier
 	producersByQueueName map[string]*producer
 	queueMaintainer      *maintenance.QueueMaintainer
+	services             []startstop.Service
 	subscriptions        map[int]*eventSubscription
 	subscriptionsMu      sync.Mutex
 	subscriptionsSeq     int // used for generating simple IDs
@@ -461,6 +462,8 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		instanceName := "default"
 
 		client.notifier = notifier.New(archetype, driver.GetListener(), client.monitor.SetNotifierStatus)
+		client.services = append(client.services, client.notifier)
+
 		var err error
 		client.elector, err = leadership.NewElector(archetype, driver.GetExecutor(), client.notifier, instanceName, client.ID(), 5*time.Second, 10*time.Second, logger)
 		if err != nil {
@@ -470,6 +473,12 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		if err := client.provisionProducers(); err != nil {
 			return nil, err
 		}
+
+		client.services = append(client.services,
+			startstop.StartStopFunc(client.logStatsLoop))
+
+		client.services = append(client.services,
+			startstop.StartStopFunc(client.handleLeadershipChangeLoop))
 
 		//
 		// Maintenance services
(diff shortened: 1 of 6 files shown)


client.go
```diff
@@ -598,45 +609,32 @@ func (c *Client[TTx]) Start(ctx context.Context) error {
 	// Monitor should be the first subprocess to start, and the last to stop.
 	// It's not part of the waitgroup because we need to wait for everything else
 	// to shut down prior to closing the monitor.
-	go c.monitor.Run()
+	//
+	// Unlike other services, it's given a background context so that it doesn't
+	// cancel on normal stops.
+	if err := c.monitor.Start(context.Background()); err != nil { //nolint:contextcheck
```
brandur, line 615: Won't return an error, but like other services, it has the option to do so.

client.go
```diff
@@ -598,45 +609,32 @@ func (c *Client[TTx]) Start(ctx context.Context) error {
...
-					c.handleLeadershipChange(ctx, notification)
-				}
-			}
-		}()
+	for _, service := range c.services {
+		if err := service.Start(fetchNewWorkCtx); err != nil {
+			// In case of error, stop any services that might have started. This
+			// is safe because even services that were never started will still
+			// tolerate being stopped.
+			startstop.StopAllParallel(c.services)
```
brandur, line 628: We now try and stop anything that started even in the event of a start error, which is kind of nice.

client.go
```diff
@@ -880,41 +882,80 @@ func (c *Client[TTx]) logStatsLoop(ctx context.Context) {
...
+				return
 
-		case <-ticker.C:
-			logStats()
+			case <-ticker.C:
+				logStats()
+			}
 		}
-	}
+	}()
```
brandur, line 906: This all just moves down from above where it was previously in a closure.

client_monitor.go
```diff
@@ -26,32 +26,31 @@ func newClientMonitor() *clientMonitor {
 		// so that the senders can avoid blocking and to account for some delivery delay.
 		snapshotBuffer:  make(chan snapshotAndSubscribers, 100),
 		currentSnapshot: componentstatus.ClientSnapshot{Producers: make(map[string]componentstatus.Status)},
-
-		shutdownOnce: &sync.Once{},
-		shutdownCh:   make(chan struct{}),
-		doneCh:       make(chan struct{}),
```
brandur, line 32: Lets us get rid of a whole bunch of vars, which is nice. We add a `StartStopStress` test case also, so we're now much more certain that this code works than before.

brandur:
@bgentry I'm sure you're just going to love this one lol, but I promise it's for the greater good.

bgentry:
> @bgentry I'm sure you're just going to love this one lol, but I promise it's for the greater good.

Actually wasn't too bad at all :)

brandur:
> Actually wasn't too bad at all :)

Hah, okay that's a relief :) Thanks!

PR #263, opened 2024-03-11
Refactor elector somewhat + test suite + poll-only mode

brandur:
Here, do a little refactoring in the elector. The overall code doesn't
change that much, but we try to tighten things up with little things
like improved logging and real exponential backoff. It becomes a
start/stop service so that it's a little more normalized with other code
and more robust on start/stop.

The major improvement is the addition of a test suite. Although there
was a nominal suite before that was added during the driver refactor to
test the `attemptElectOrReelect` function, the elector was only tested
indirectly otherwise through the client. The change comes with a variety
of tests that exercise the elector's various behaviors, including one
that pits multiple competing electors against each other to make sure
that works.

Lastly, the elector gains a "poll only" mode in which we check that it
can still function using polling only in case a listener isn't
available. This doesn't have any effect on River feature-wise yet, but
the idea is that after we've added a similar capability to the producer,
we'll be able to support systems where `LISTEN`/`NOTIFY` aren't
available like PgBouncer in transaction mode, or possibly even MySQL or
SQLite in the future. We send the poll only mode through the same
barrage of tests that we require it to pass when using a database pool.

### DIFF
--- a/internal/leadership/elector.go
+++ b/internal/leadership/elector.go
@@ -4,14 +4,21 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"log/slog"
 	"sync"
 	"time"
 
 	"github.com/riverqueue/river/internal/baseservice"
+	"github.com/riverqueue/river/internal/maintenance/startstop"
 	"github.com/riverqueue/river/internal/notifier"
+	"github.com/riverqueue/river/internal/rivercommon"
 	"github.com/riverqueue/river/internal/util/dbutil"
 	"github.com/riverqueue/river/riverdriver"
+)
+
+const (
+	electInterval           = 5 * time.Second
+	electInteralJitter      = 1 * time.Second
+	electIntervalTTLPadding = 10 * time.Second
 )
 
 type pgNotification struct {
@@ -43,16 +50,36 @@ func (s *Subscription) Unlisten() {
 	})
 }
 
+// Test-only properties.
+type electorTestSignals struct {
+	DeniedLeadership     rivercommon.TestSignal[struct{}] // notifies when elector fails to gain leadership
+	GainedLeadership     rivercommon.TestSignal[struct{}] // notifies when elector gains leadership
+	LostLeadership       rivercommon.TestSignal[struct{}] // notifies when an elected leader loses leadership
+	MaintainedLeadership rivercommon.TestSignal[struct{}] // notifies when elector maintains leadership
+	ResignedLeadership   rivercommon.TestSignal[struct{}] // notifies when elector resigns leadership
+}
+
+func (ts *electorTestSignals) Init() {
+	ts.DeniedLeadership.Init()
+	ts.GainedLeadership.Init()
+	ts.LostLeadership.Init()
+	ts.MaintainedLeadership.Init()
+	ts.ResignedLeadership.Init()
+}
+
 type Elector struct {
 	baseservice.BaseService
+	startstop.BaseStartStop
 
-	exec     riverdriver.Executor
-	id       string
-	interval time.Duration
-	logger   *slog.Logger
-	name     string
-	notifier *notifier.Notifier
-	ttl      time.Duration
+	clientID                   string
+	electInterval              time.Duration // period on which each elector attempts elect even without having received a resignation notification
+	electIntervalJitter        time.Duration
+	exec                       riverdriver.Executor
+	instanceName               string
+	leadershipNotificationChan chan struct{}
+	notifier                   *notifier.Notifier
+	testSignals                electorTestSignals
+	ttl                        time.Duration
 
 	mu            sync.Mutex
 	isLeader      bool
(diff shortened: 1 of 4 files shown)


internal/notifier/notifier.go
```diff
@@ -442,6 +442,8 @@ func (n *Notifier) Listen(ctx context.Context, topic NotificationTopic, notifyFu
 	}
 	n.subscriptions[topic] = append(existingSubs, sub)
 
+	n.Logger.InfoContext(ctx, n.Name+": Added subscription", "new_num_subscriptions", len(n.subscriptions[topic]), "topic", topic)
```
brandur, line 445: Found having a little more logging on subscriptions pretty handy for test debugging purposes.

internal/leadership/elector.go
```diff
@@ -62,206 +89,294 @@ type Elector struct {
...
-	sub, err := notifier.ListenRetryLoop(ctx, &e.BaseService, e.notifier, notifier.NotificationTopicLeadership, handleNotification)
-	if err != nil { //nolint:staticcheck
-		// TODO(brandur): Propagate this after refactor.
-	}
-	if sub != nil {
-		defer sub.Unlisten(ctx)
-	}
+			if notification.Action != "resigned" || notification.Name != e.instanceName {
+				// We only care about resignations on because we use them to preempt the
+				// election attempt backoff. And we only care about our own key name.
```
bgentry, line 139: typo here:

> We only care about resignations on because

internal/leadership/elector.go
```diff
@@ -62,206 +89,294 @@ type Elector struct {
...
 		}
-		if success {
-			return true
+		if elected {
+			return nil
 		}
 
+		numErrors = 0
+
+		e.Logger.Info(e.Name+": Leadership bid was unsuccessful (not an error)", "client_id", e.clientID)
```
bgentry, line 227: `Info` level might be a bit noisy for this running every 5 seconds on every worker and it being a completely benign/expected scenario, maybe `Debug`?
```suggestion
		e.Logger.Debug(e.Name+": Leadership bid was unsuccessful (not an error)", "client_id", e.clientID)
```

internal/leadership/elector.go
```diff
@@ -62,206 +89,294 @@ type Elector struct {
...
+
 		select {
-		case <-ctx.Done():
-			return false
-		case <-time.After(e.interval):
-			// TODO: This could potentially leak memory / timers if we're seeing a ton
-			// of resignations. May want to make this reusable & cancel it when retrying?
-		case <-leadershipNotificationChan:
-			// Somebody just resigned, try to win the next election immediately.
+		case <-e.CancellableSleepRandomBetweenC(ctx, e.electInterval, e.electInterval+e.electIntervalJitter):
```
bgentry, line 231: there's a potential leaking of timers here, but only if resignations are happening frequently

internal/leadership/elector.go
```diff
@@ -62,206 +89,294 @@ type Elector struct {
...
+			e.testSignals.GainedLeadership.Signal(struct{}{})
+
+			err := e.keepLeadershipLoop(ctx)
+			if err != nil {
+				if errors.Is(err, context.Canceled) {
+					return
+				}
+
+				if !errors.Is(err, errLostLeadership) {
+					e.Logger.Error(e.Name+": Error keeping leadership", "client_id", e.clientID, "err", err)
```
bgentry, line 191: Is this inverted? The message seems to align with when the node _lost_ its leadership

internal/notifier/notifier.go
bgentry: This one also feels like it should maybe be a `Debug` level given it's indicative of healthy operation and pretty low level, thoughts?

bgentry (approved):
Couple of minor comments, but this looks really fantastic. Great work 🎉

internal/leadership/elector.go
brandur: Thx.

internal/leadership/elector.go
brandur: I'll say that because we probably don't expect _that_ many different River clients operating in one installation, so it's not like these will be getting endlessly spewed.

That said, now that the tests are all fixed up, I can live with debug, so changed.

Also, I changed everything to `InfoContext`/`DebugContext` which I'd meant to do but forgot.

internal/leadership/elector.go
brandur: Yeah, good point. I just added the comment back that was previously there.

I think it might be kind of neat if we could do a `timeutil` utility exactly like `Ticker` but which would tick within a random range. I'll put it on my TODO list.

internal/leadership/elector.go
brandur: A little hard to read, but the idea is to log any error that's not a reelection bid loss.

Changed the code so that:

* Inverted the condition and put in a `continue`, with the log statement falling to below the conditional.
* Renamed the error `errLostLeadershipReelection` for better clarity.

internal/notifier/notifier.go
brandur: K, made those two debug too.

brandur:
Thanks!

PR #266, opened 2024-03-12
General clean up of `testdbman` program + drop Cobra dependency

brandur:
This one's mainly in pursuit of dropping the Cobra dependency in the
main River package. It's still nice to have Cobra for the River CLI, but
the only reason that top-level River needs it is that `testdbman` (the
command used to drop and raise test databases needs it). Since
`testdbman`'s CLI API is so simple anyway, here we move it off of Cobra
and over to a small internal CLI framework, maintaining an identical
interface as before (with the small change that `--help` is now `-help`
since we're using Go's internal flag module).

This would normally be a dumb thing to do, but I don't expect the
`testdbman` interface to have to get anymore complicated (it hasn't
changed at all since River's inception), the internal framework isn't a
huge amount of code, and I put some test coverage on it to make sure
that everything works as expected.

In addition to that, we do a little housekeeping on the internal code
for `testdbman`:

* Reduce visual noise by stripping extra lines and functions that were
  doing very little, and removing some specific contexts in favor of one
  top-level one for the program.

* Move the core functions and command definitions into one file so that
  quick file find works a little better and it's easier to get a
  holistic view of what everything is doing. Especially after trimming
  code, there's so little code that it all fits together without feeling
  crowded, and previously some of the files like `reset.go` were doing
  practically nothing.

* A single unified error handling schema. Previously some code was
  printing to stderr and aborting the program and some code was
  returning errors. Now everything returns an error with a single
  top-level error handler.

* Make output and error formatting more consistent across commands.

* Tighten up some documentation and comments that'd fallen out of date
  since the program was originally written.

* Remove some helper functions like `dbURL` that were doing more to
  obscure than to help (it had a special path for the management
  database URL and then did a string concatenation otherwise, but it's
  much more clear to just use a constant for the management database URL
  when we know that we want it). Add helpers where appropriate like
  `generateTestDBNames` that let us put a little (very modest amount for
  the time being) test coverage onto the program.


brandur:
@bgentry This one low priority, but just as a general nicety.

internal/cmd/testdbman/main.go
```diff
@@ -1,26 +1,403 @@
...
+func generateTestDBNames(numDBs int) []string {
+	dbNames := []string{
+		"river_testdb",
+		"river_testdb_example",
+	}
+
+	// This is the same default as pgxpool's maximum number of connections
+	// when not specified -- either 4 or the number of CPUs, whichever is
+	// greater. If changing this number, also change the similar value in
+	// `riverinternaltest` where it's duplicated.
```
bgentry, line 158: This comment is probably better placed above line 138 where the `numDBs` is calculated

bgentry (approved):

internal/cmd/testdbman/main.go
brandur: Ah whoops. Good call. Moved.

PR #271, opened 2024-03-13
Debug bootstrap for intermittent test in periodic job scheduler

brandur:
This one's a special CI and logging set up aimed to help me debug #215,
which I'm finding practically impossible to reproduce locally, but which
happens with reasonable frequency in CI. Not intended for merge.


brandur:
Opened #274 with a fix.

PR #272, opened 2024-03-16
Make client a start/stop service

brandur:
Here, continue service refactoring to its ultimate conclusion, and make
the client itself a start/stop service, having a couple advantages:

* (IMO) considerably simplifies the start and stop code, putting it all
  in one place, and largely even one function. Fewer helpers to follow
  around to understand what's going on.

* Makes the client behave gracefully on double starts/stops.

* Allows the client to easily handle start/stop under duress. Any number
  of goroutines can be starting or stopping it simultaneously and it's
  guaranteed free from races.

Because the client's stop functions are different from the signature of
other start/stop services (taking a context and returning an error), I
had to modify the base start/stop service somewhat and allow for an
additional `StopInit` helper (like `StartInit`) that lets stop behavior
be customized while still providing race-free operations.

### DIFF
--- a/client.go
+++ b/client.go
@@ -23,6 +23,7 @@ import (
 	"github.com/riverqueue/river/internal/maintenance/startstop"
 	"github.com/riverqueue/river/internal/notifier"
 	"github.com/riverqueue/river/internal/rivercommon"
+	"github.com/riverqueue/river/internal/util/maputil"
 	"github.com/riverqueue/river/internal/util/randutil"
 	"github.com/riverqueue/river/internal/util/sliceutil"
 	"github.com/riverqueue/river/internal/util/valutil"
@@ -262,19 +263,15 @@ type QueueConfig struct {
 // multiple instances operating on different databases or Postgres schemas
 // within a single database.
 type Client[TTx any] struct {
-	// BaseService can't be embedded like on other services because its
-	// properties would leak to the external API.
-	baseService baseservice.BaseService
+	// BaseService and BaseStartStop can't be embedded like on other services
+	// because their properties would leak to the external API.
+	baseService   baseservice.BaseService
+	baseStartStop startstop.BaseStartStop
 
 	completer jobcompleter.JobCompleter
 	config    *Config
 	driver    riverdriver.Driver[TTx]
 	elector   *leadership.Elector
-
-	// fetchWorkCancel cancels the context used for fetching new work. This
-	// will be used to stop fetching new work whenever stop is initiated, or
-	// when the context provided to Run is itself cancelled.
-	fetchWorkCancel context.CancelCauseFunc
 
 	monitor              *clientMonitor
 	notifier             *notifier.Notifier
@@ -284,10 +281,10 @@ type Client[TTx any] struct {
 	subscriptions        map[int]*eventSubscription
 	subscriptionsMu      sync.Mutex
 	subscriptionsSeq     int // used for generating simple IDs
-	stopComplete         chan struct{}
 	statsAggregate       jobstats.JobStatistics
 	statsMu              sync.Mutex
 	statsNumJobs         int
+	stopped              chan struct{}
 	testSignals          clientTestSignals
 	uniqueInserter       *dbunique.UniqueInserter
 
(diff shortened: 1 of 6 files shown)


brandur:
This one a lot more readable without whitespace changes:

https://github.com/riverqueue/river/pull/272/files?w=1

client_test.go
```diff
@@ -515,17 +536,6 @@ func Test_Client_Stop(t *testing.T) {
...
-	t.Run("not started", func(t *testing.T) {
-		t.Parallel()
-
-		dbPool := riverinternaltest.TestDB(ctx, t)
-		client := newTestClient(t, dbPool, newTestConfig(t, nil))
-
-		err := client.Stop(ctx)
-		require.Error(t, err)
-		require.Equal(t, "client not started", err.Error())
-	})
```
brandur, line 527: I originally left this error in, but to guarantee non-raciness, I wouldn't had to an extra mutex to protect it, and it didn't seem worth it since the error's not particularly valuable. (The error wasn't race protected before either, but now we're actually testing it so the problem was revealed.)

client.go
```diff
@@ -575,146 +571,154 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
...
+			return rivercommon.ErrShutdown
+		}
+		return err
+	}
 
-	// As of now, the Adapter doesn't have any async behavior, so we don't need
-	// to wait for it to stop.  Once all executors and completers are done, we
-	// know that nothing else is happening that's from us.
+	go func() {
+		defer close(stopped)
```
bgentry, line 679: It might be more readable and easily debuggable if some of these closures were split into internal client methods. Thinking in particular of the potential for a trace to have anonymous goroutines rather than named methods.

Not a huge concern but it might be worth trying to shuffle some of this around to improve the readability of `Start()`? Particularly here and in the large startup closure above.

client_test.go
```diff
@@ -2564,8 +2574,8 @@ func Test_Client_InsertTriggersImmediateWork(t *testing.T) {
 	ctx := context.Background()
 	require := require.New(t)
 
-	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
-	t.Cleanup(cancel)
+	// ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	// t.Cleanup(cancel)
```
bgentry, line 2578: Mistakenly commented?

internal/maintenance/startstop/start_stop.go
```diff
@@ -79,27 +100,63 @@ func (s *BaseStartStop) StartInit(ctx context.Context) (context.Context, bool, c
...
+	if !shouldStop {
+		return
+	}
+
+	<-stopped
+	finalizeStop(true)
+}
+
+// StopInit provides a way to build a more customized Stop implementation. It
+// should be avoided unless there'a an exception reason not to because Stop
```
bgentry, line 121:
```suggestion
// should be avoided unless there's an exceptional reason not to because Stop
```

bgentry (approved):

client_test.go
brandur: Oops yep. Actually we should talk about this, but the reason I had it commented is that in case your program is actually stuck, the canceled context supersedes `go test`'s `-timeout` option, so instead of `-timeout` stopping the test of a full stack trace of every goroutine that was still running, you end up with a canceled test with no debugging information.

So a long way of saying that tighter `-timeout` parameters are a better option than context timeouts when it comes to use in tests, but in the interest of not getting into that now, I just uncommented these lines.

internal/maintenance/startstop/start_stop.go
brandur: Good catch, thanks.

client.go
brandur: Took a look at this, and not against it, but I think I'm going to punt for the time being. I kind of like having all the logic `Start` (and the majority of the stop) logic in one place because you can scan it all at once without having to jump around. Also lets us reduce the number of struct variables and helpers in favor of definitions that can't leak into other contexts, and are defined more visibly, closer to home.

brandur:
ty.

PR #273, opened 2024-03-16
Migration CLI: Add `migrate-get` command and `--dry-run`/`--show-sql` options

### LINKED ISSUE #209: River Schema Migration Improvement
I realize that everyone has their favorite coolaid but applying the migrations with the current pattern makes it a bit difficult to implement.  I was wondering if there was a way to run the river CLI tool to generate a SQL patch of the given changes?

Most people have already chosen a tool they like and are actively using it, so having yet another way to modify the SQL schema is a bit distasteful.  Now we have two different tools that have write access to a DB that need to be both installed, managed and executed as part of a given deployment.

Given the wide range of options in the ecosystem. (See below), I was wondering if we could just create a SQL 

- https://github.com/ariga/atlas
- https://github.com/amacneil/dbmate
- https://github.com/golang-migrate/migrate
- https://github.com/pressly/goose
- https://github.com/rubenv/sql-migrate

What I'd like to see is something like.

river migrate-up --dry ## or pick your favorite syntax to achieve the goal.

SQL:
```sql
-- migrate-up
ALTER TABLE add COLUMN.... 

--migrate-down
ALTER TABLE DROP COLUMN....
```

How you choose to integrate the SQL into your migration tool is up to you, but at least give me the SQL that will be executed.

brandur:
This one's in pursuit of resolving #209, which I've decided to tackle
now because we're going to have to cut a CLI release across River
package versions for #258 anyway, so this'll reuse some work.

A new `river migrate-get` command becomes available, whose only job is
to dump SQL from River migrations so that it can easily be plugged into
other migration frameworks. Its use looks like:

    river migrate-get --version 3 --down > version3.down.sql
    river migrate-get --version 3 --up > version3.up.sql

It can also take multiple versions:

    river migrate-get --version 3,2,1 --down > river.down.sql
    river migrate-get --version 1,2.3 --up > river.up.sql

It can also dump _all_ migrations, which will be useful in cases where
users want to avoid River's internal migration framework completely, and
use their own:

    river migrate-get --all --exclude-version 1 --up > river_all.up.sql
    river migrate-get --all --exclude-version 1 --down > river_all.down.sql

Along with that, `migrate-down` and `migrate-up` get a few new useful
options:

* `--dry-run`: Prints information on migrations that would be run, but
  doesn't modify the database in any way.

* `--show-sql`: Prints SQL for each migration step that was applied.

This gives users an easy way to, after a River upgrade, run the CLI to
see what commands would be run were they to migrate, but without
actually performing the migration, likely a step that most production
users would perform to be cautious:

    river migrate-up --dry-run --show-sql

I've also done a little cleanup around the River CLI's `main.go`. The
`--verbose` and `--debug` commands added in #258 are now promoted to
persistent flag configuration so they're available for all commands, and
we now have one standardized way of initializing an appropriate logger.

Fixes #209.


cmd/river/main.go
```diff
@@ -69,66 +90,104 @@ before starting the client, and works until all jobs are finished.
...
+migration SQL to be dumped for use elsewhere.
+
+Specify a single version with --version, and one of --down or --up.
+	`),
+			Run: func(cmd *cobra.Command, args []string) {
+				execHandlingError(func() (bool, error) { return migrateGet(ctx, makeLogger(), os.Stdout, &opts) })
+			},
+		}
+		cmd.Flags().BoolVar(&opts.Down, "down", false, "print down migration")
+		cmd.Flags().BoolVar(&opts.Up, "up", false, "print up migration")
```
brandur, line 161: It's a little odd that this command takes `--down` and `--up` options, when the migration commands are `migrate-down` and `migrate-up`, but I tried to consider all the alternatives, and I think this is the least bad. Having two full commands for `migrate-get-down` and `migrate-get-up` feels too heavy handed and unwieldy.

cmd/river/main_test.go
```diff
@@ -0,0 +1,16 @@
...
+
+	"github.com/riverqueue/river/rivermigrate"
+)
+
+func TestMigrationComment(t *testing.T) {
+	t.Parallel()
+
+	require.Equal(t, "-- River migration 001 [down]", migrationComment(1, rivermigrate.DirectionDown))
+	require.Equal(t, "-- River migration 002 [up]", migrationComment(2, rivermigrate.DirectionUp))
+}
```
brandur, line 16: Ugh, I wish I could write more tests for the CLI, but I'm getting hard blocked by the fact that none of our internal test utilities are available to this module. Something to look into separately, but punting on it for now.

bgentry (approved):

brandur:
thx!

PR #275, opened 2024-03-17
Make more River logging debug level

brandur:
Here, in an attempt to reduce the level of noise of River logging
somewhat, move more logging that's not as useful from into to debug
level. Most of the changed log statements are the start/stop message of
each subservice (so the River client will still emit start/stop at info
level, but each subservice moves theirs to debug).

Also remove a few log statements that are quite duplicative of an almost
identical one happening elsehwere (e.g. "client started" in benchmark,
which is almost the same as "client started" in the client), and a few
that I'm pretty sure I'd left in by accident while debugging.

The CLI now takes a `--debug` parameter, and `riverinternaltest` now
provides `BaseServiceArchetypeDebug` to easily swap into a failing test,
so access to the additional logging is fairly easy to get at in case
it's needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- The level of some of River's common log statements has changed, most often demoting `info` statements to `debug` so that `info`-level logging is overall less verbose. [PR #275](https://github.com/riverqueue/river/pull/275).
+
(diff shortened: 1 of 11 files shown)


brandur:
@bgentry Really small one aimed at reducing default log verbosity. I figure striking the perfect balance here is going to take a little trial and error, but hopefully this gets the ball rolling.

bgentry (approved):

brandur:
Thx. I also added a short changelog entry about this. It's a small change, but one that's somewhat observable from a user's perspective.

PR #276, opened 2024-03-17
Tee up version `v0.1.0` including new CLI release

brandur:
Tees up version `v0.1.0`, containing a number of changes like `river
bench`, `river migrate-get`, and a new, faster, completer.

We could pool a few more changes before release, but I kind of want to
get this out the door so I don't forget to the cut a CLI release and
remove the `replace` directives in its `go.mod`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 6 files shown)


brandur:
Unfortunately this build is going to crash and burn because I took those `replace` directives out of the CLI's `go.mod`, but it should be fixed as cut the final release.

bgentry (approved):

brandur:
Thanks!

PR #277, opened 2024-03-17
Bump River CLI to `v0.1.0`

brandur:
Follows up #276. Now that the main River package is tagged with
`v0.1.0`, we now bump that version in the River CLI module. After this
is merged, we'll tag it as `v0.1.0` as well.


brandur:
@bgentry Actually, just going to bypass branch protections so I can finish this really. The change is super super tiny and its requirement is implied by #276.

PR #279, opened 2024-03-19
Fix release instructions for River CLI + make accidental copy/paste harder

brandur:
I noticed yesterday while releasing `v0.1.0` that the CLI's release
instructions are still not quite right. In particular, the main block of
`git tag` includes `cmd/river/$VERSION`, so it's a really easy to
mistakenly tag the CLI with the wrong commit when you're really trying
to release it separately.

In this change:

* Move the CLI tag to a separate command block with clear instructions
  to skip it if the CLI needs a special release.

* Expand the section on releasing the CLI to have specific `go get`
  commands for each required mainline package containing `$VERSION`.

* Indicate that a second pull request will be required for the CLI
  release, and add a step explicitly indicating that `replace`
  directives in the CLI's `go.mod` should be removed (it's better for
  the build's health if they're removed in the second phase rather than
  the first).

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -27,30 +27,39 @@ queries. After changing an sqlc `.sql` file, generate Go with:
 
 1. Fetch changes to the repo and any new tags. Export `VERSION` by incrementing the last tag. Execute `update-submodule-versions` to add it the project's `go.mod` files:
 
-```shell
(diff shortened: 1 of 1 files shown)


bgentry (approved):

brandur:
thx.

PR #280, opened 2024-03-19
Fix hot loop in reindexer service + no more disabling sleep in base services

brandur:
Fix a bug in the reindexer which was causing it to enter a hot loop on
shutdown. It wasn't doing any work since reindexing is currently
disabled, but it was emitting an inordinate amount of logging output.

The problem stemmed from trying to use a `context.Deadline` as a timer:

    for {
        timerCancel := context.WithDeadline(ctx, nextRunAt)

        select {
        case <-ctx.Done():
            timerCancel()
            return
        case <-timerCtx.Done():
        }

        ...
    }

The bug is easy to miss on the first read, but involves context
inheritance. A service's context is cancelled when its being stopped,
and in this case `timerCtx` inherits from `ctx`, so on stop, _both_
`ctx` and `timerCtx` are cancelled, making the `select` statement's
precedence indeterminate, and causing the loop to fall through and
continue an arbitrary number of times. (I would've expected the branch
taken to be a 50/50, but there's some other implementation detail at
work here because I was seeing `timerCtx` preferred in a statistically
improbable way on most runs, and it'd often fire 4-5 times before
`ctx.Done()` was found a test case would end.)

I replaced the use of a context with a `CancellableSleep` with a context
check following it.

This then let to a bit of a yakshake because the cancellable sleep was
being respect _either_, which turned out to be because all base service
sleep is disabled in the queue maintainers in order to avoid their
staggered sleep that occurs on start up.

I changed my mind a while ago that being able to broadly disable sleep
like this is actually not a very good idea, but hadn't gotten around to
refactoring it out yet, so I finally did so here. The staggered sleep is
moved to a new base function with a disable flag specifically for this
use rather than all sleeps everywhere:

    func (s *queueMaintainerServiceBase) StaggerStart(ctx context.Context) {
        if s.staggerStartupDisabled {
            return
        }

        const (
            jitterMin = 0 * time.Second
            jitterMax = 1 * time.Second
        )

        s.CancellableSleepRandomBetween(ctx, jitterMin, jitterMax)
    }

The client then may disable start on all the queue maintainer services
as it initializes them based on a non-exported flag from its `Config`.
Unlike disabling sleep, disabling start up stagger is _always_
appropriate in client tests, and won't lead to unintended side effects
like accidentally disabling sleep on exponential backoffs.

Over all, despite a little more work for services that want to disable
their sleep for tests, I think this produces a far more obvious model
that's going to lead to fewer test bugs.

### DIFF
--- a/internal/baseservice/base_service.go
+++ b/internal/baseservice/base_service.go
@@ -22,13 +22,6 @@ import (
 // embedded in BaseService, so these properties are available on services
 // directly.
 type Archetype struct {
-	// DisableSleep disables sleep in services as long as they're sleeping via
-	// `BaseService`'s `CancellableSleep` functions. This is meant to provide a
-	// convenient way to disable sleep in one place and which will automatically
-	// propagate from a parent into its subcomponents as they initialize
-	// themselves from its archetype.
-	DisableSleep bool
-
 	// Logger is a structured logger.
 	Logger *slog.Logger
 
@@ -46,14 +39,6 @@ type Archetype struct {
 	// injection. Services should try to use this function instead of the
 	// vanilla ones from the `time` package for testing purposes.
 	TimeNowUTC func() time.Time
-}
-
-// WithSleepDisabled disables sleep in services that are using `BaseService`'s
-// `CancellableSleep` functions and returns the archetype for convenience. Use
-// of this is only appropriate in tests.
-func (a *Archetype) WithSleepDisabled() *Archetype {
-	a.DisableSleep = true
-	return a
 }
 
 // BaseService is a struct that's meant to be embedded on "service-like" objects
@@ -75,13 +60,8 @@ type BaseService struct {
 }
 
 // CancellableSleep sleeps for the given duration, but returns early if context
-// has been cancelled. If the service's `DisableSleep` option is set (usually
-// appropriate in tests), sleep is skipped automatically.
+// has been cancelled.
 func (s *BaseService) CancellableSleep(ctx context.Context, sleepDuration time.Duration) {
-	if s.DisableSleep {
-		return
-	}
-
 	timer := time.NewTimer(sleepDuration)
 
 	select {
@@ -94,8 +74,7 @@ func (s *BaseService) CancellableSleep(ctx context.Context, sleepDuration time.D
 }
 
 // CancellableSleep sleeps for the given duration, but returns early if context
-// has been cancelled. If the service's `DisableSleep` option is set (usually
-// appropriate in tests), sleep is skipped automatically.
+// has been cancelled.
 //
 // This variant returns a channel that should be waited on and which will be
 // closed when either the sleep or context is done.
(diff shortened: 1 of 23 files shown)


internal/baseservice/base_service.go
```diff
@@ -48,14 +41,6 @@ type Archetype struct {
 	TimeNowUTC func() time.Time
 }
 
-// WithSleepDisabled disables sleep in services that are using `BaseService`'s
-// `CancellableSleep` functions and returns the archetype for convenience. Use
-// of this is only appropriate in tests.
-func (a *Archetype) WithSleepDisabled() *Archetype {
```
brandur, line 54: There are quite a few LOCs changed in this diff, but most of them are just removing this `WithSleepDisabled` feature from the base service. Having run into trouble with it multiple times now, I'd say it's just too dangerous/unintuitive. Services that need to disable sleep can add their own local flags for it.

internal/maintenance/reindexer_test.go
```diff
@@ -44,12 +44,24 @@ func TestReindexer(t *testing.T) {
...
 			ScheduleFunc: fromNow(500 * time.Millisecond),
 		}, bundle.exec)
+		svc.StaggerStartupDisable(true)
 		svc.TestSignals.Init()
 		t.Cleanup(svc.Stop)
 
 		return svc, bundle
 	}
 
+	runImmediatelyThanOnceAnHour := func() func(time.Time) time.Time {
```
bgentry, line 54:
```suggestion
	runImmediatelyThenOnceAnHour := func() func(time.Time) time.Time {
```

internal/maintenance/reindexer_test.go
```diff
@@ -44,12 +44,24 @@ func TestReindexer(t *testing.T) {
...
+		alreadyRan := false
+		return func(t time.Time) time.Time {
+			if alreadyRan {
+				return t.Add(time.Hour)
+			}
+			alreadyRan = true
+			return t.Add(time.Millisecond)
+		}
```
bgentry, line 62: does `TickerWithInitialTick` not work here?

internal/baseservice/base_service.go
bgentry: It does seem like a bit of a step back to spread out the problematic behavior into more places, but whatever you think is the best option here.

bgentry (approved):

internal/maintenance/reindexer_test.go
brandur: Whoops, thx.

internal/maintenance/reindexer_test.go
brandur: I thinkkk they're kind of orthogonal. The main purpose of this function is that it explicitly returns something that's got the right signature to be a `ScheduleFunc`.

internal/baseservice/base_service.go
brandur: Yeah ... I  think being able to disable sleep is good, but being able to _broadly_ disable it is just too dangerous/easy to miss. I'm totally fine trying to come up with a different/better generalized approach if we can find one, but more precise sleep disables for the relatively few test cases that need it seems like a reasonable compromise for now.

brandur:
Thx! Addressed feedback and added minor changelog entry.

PR #281, opened 2024-03-19
Poll only mode for environments without listen/notify

brandur:
Here, add a new River client option in `Config.PollOnly` which causes it
to start in "poll only" mode where a notifier isn't initialized, and no
`LISTEN` statements are issued. This has the downside of events like a
leadership resignation or a new job being available not noticed as
quickly, but the advantage of making River operable on systems where
listen/notify aren't supported, like PgBouncer in transaction mode, or
maybe even eventually MySQL or SQLite.

A slightly unrelated change that I also made here was to give the
elector a more traditional `Config` struct (like other services take)
that encompasses more of its configuration instead of raw constructor
parameters. I thought I needed to do this originally to better expose a
custom elect interval for the client's `PollOnly` test, but it turned
out that my slow test was actually due to a different problem and the
extra configuration wasn't actually necessary. It seemed to me to clean
things up somewhat though, so I left the refactor in.

I also remove the listen retry loop on client start up found in the
elector and producer. These are probably a bad idea because they may
hide a real database problem as they enter their retry loops, and cause
the client to require a very lengthy amount of time for its `Start` to
fail because of the built-in sleep backoffs. But we'd added them over
concerns that the notifier's previous implementation (prior to #253) of
discarding all listen/notify problems may have papered over errors that
would've otherwise been returned for people trying to use a River client
against say a PgBouncer operating in transaction pooling mode.

So instead, we now fail fast if there's a problem with listen/notify in
their database, and for those wanting to use PgBouncer in transaction
pooling mode, we can now recommend using the much cleaner approach of
activating `PollOnly` instead. Users can implement their own retry loop
on client `Start` if they'd like to protect against intermittent
problems on listen/notify start up, but personally I wouldn't expect
this to ever be desirable. (The alternative is to allow the program to
fail fast and have its supervisor resurrect it, like most programs would
do for any other start up hiccup.) The program will already fail fast
under most circumstances due to the `SELECT` it performs on start.

### DIFF
--- a/client.go
+++ b/client.go
@@ -135,6 +135,21 @@ type Config struct {
 	// in the client.
 	PeriodicJobs []*PeriodicJob
 
+	// PollOnly starts the client in "poll only" mode, which avoids issuing
+	// `LISTEN` statements to wait for events like a leadership resignation or
+	// new job available. The program instead polls periodically to look for
+	// changes (checking for new jobs on the period in FetchPollInterval).
+	//
+	// The downside of this mode of operation is that events will usually be
+	// noticed less quickly. A new job in the queue may have to wait up to
+	// FetchPollInterval to be locked for work. When a leader resigns, it will
+	// be up to five seconds before a new one elects itself.
+	//
+	// The upside is that it makes River compatible with systems where
+	// listen/notify isn't available. For example, PgBouncer in transaction
+	// pooling mode.
+	PollOnly bool
+
 	// Queues is a list of queue names for this client to operate on along with
 	// configuration for the queue like the maximum number of workers to run for
 	// each queue.
@@ -270,13 +285,12 @@ type Client[TTx any] struct {
 	baseService   baseservice.BaseService
 	baseStartStop startstop.BaseStartStop
 
-	completer jobcompleter.JobCompleter
-	config    *Config
-	driver    riverdriver.Driver[TTx]
-	elector   *leadership.Elector
-
+	completer            jobcompleter.JobCompleter
+	config               *Config
+	driver               riverdriver.Driver[TTx]
+	elector              *leadership.Elector
 	monitor              *clientMonitor
-	notifier             *notifier.Notifier
+	notifier             *notifier.Notifier // may be nil in poll-only mode
 	periodicJobs         *PeriodicJobBundle
 	producersByQueueName map[string]*producer
 	queueMaintainer      *maintenance.QueueMaintainer
@@ -408,6 +422,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		JobTimeout:                  valutil.ValOrDefault(config.JobTimeout, JobTimeoutDefault),
 		Logger:                      logger,
 		PeriodicJobs:                config.PeriodicJobs,
+		PollOnly:                    config.PollOnly,
 		Queues:                      config.Queues,
 		ReindexerSchedule:           config.ReindexerSchedule,
 		RescueStuckJobsAfter:        valutil.ValOrDefault(config.RescueStuckJobsAfter, rescueAfter),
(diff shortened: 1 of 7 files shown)


client.go
```diff
@@ -448,22 +463,39 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
...
+				JobTimeout:        config.JobTimeout,
+				MaxWorkers:        queueConfig.MaxWorkers,
+				Notifier:          client.notifier,
+				Queue:             queue,
+				RetryPolicy:       config.RetryPolicy,
+				SchedulerInterval: config.schedulerInterval,
+				StatusFunc:        client.monitor.SetProducerStatus,
+				Workers:           config.Workers,
+			})
+			client.monitor.InitializeProducerStatus(queue)
```
brandur, line 498: Moved initializing the producers up into `New` here. The helper function was so small, and this makes seeing their dependencies super clear as they're initialized alongside all the other services.

internal/leadership/elector.go
```diff
@@ -13,13 +13,19 @@ import (
...
-	electInteralJitter      = 1 * time.Second
-	electIntervalTTLPadding = 10 * time.Second
+	electIntervalDefault           = 5 * time.Second
+	electIntervalJitterDefault     = 1 * time.Second
+	electIntervalTTLPaddingDefault = 10 * time.Second
+
+	// TODO: for now we only support a single instance per database/schema.
+	// If we want to provide isolation within a single database/schema,
+	// we'll need to add a client config for this.
+	instanceNameDefault = "default"
```
brandur, line 28: Moved this down from client so it's defined closer to where it's used. Totally fine to move it back up as necessary, but there's no plans AFAIK to make use of it anytime soon and this keeps things clearer for now.

brandur:
@bgentry Going to quiet down on changes after this, but we were 95% of the way to this feature and it required fairly minimal changes, so decided to take it through. Thoughts?

producer.go
```diff
@@ -210,7 +210,7 @@ func (p *producer) StartWorkContext(fetchCtx, workCtx context.Context) error {
 			fetchLimiter.Call()
 		}
 		// TODO(brandur): Get rid of this retry loop after refactor.
-		insertSub, err = notifier.ListenRetryLoop(fetchCtx, &p.BaseService, p.config.Notifier, notifier.NotificationTopicInsert, handleInsertNotification)
+		insertSub, err = p.config.Notifier.Listen(fetchCtx, notifier.NotificationTopicInsert, handleInsertNotification)
```
bgentry, line 213: comment above still relevant?

producer.go
```diff
@@ -241,7 +241,7 @@ func (p *producer) StartWorkContext(fetchCtx, workCtx context.Context) error {
 			)
 		}
 		// TODO(brandur): Get rid of this retry loop after refactor.
-		jobControlSub, err = notifier.ListenRetryLoop(fetchCtx, &p.BaseService, p.config.Notifier, notifier.NotificationTopicJobControl, handleJobControlNotification)
+		jobControlSub, err = p.config.Notifier.Listen(fetchCtx, notifier.NotificationTopicJobControl, handleJobControlNotification)
```
bgentry, line 244: comment above still relevant?

bgentry (approved):

client_test.go
```diff
@@ -180,34 +180,47 @@ func runNewTestClient(ctx context.Context, t *testing.T, config *Config) *Client
 	return client
 }
 
-func Test_Client(t *testing.T) {
+func Test_Client_Standard(t *testing.T) {
```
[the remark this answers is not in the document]
brandur: Ugh, I think I was trying to rename this to prevent overzealous test matching when trying to run a single test with `-run`. This is a long issue to explain, so reverting this and will start a separate discussion about it if necessary.

internal/leadership/elector.go
brandur: kk, hopefully not too big of a problem. Thx.

producer.go
brandur: Oops, totally thought I killed these. Thanks!

brandur:
Ty.

PR #282, opened 2024-03-19
Use timer for reindexer instead of cancellable sleep

brandur:
Quick follow up to #280 in which we rebuild the reindexer somewhat to
look more like the periodic job enqueuer. The benefits in readability of
each is debatable, but the periodic job enqueuer's timer model has the
minor advantage of reusing the same timer over and over again, and also
that it makes the difference maintenance services look more like each
other. It also allowed the opportunity to simplify the code somewhat and
removed one of the time variables which is no longer needed.

Also, change a line in the periodic job enqueuer so we use a `return`
statement instead of a break with a named loop.


internal/maintenance/reindexer.go
```diff
@@ -105,31 +105,48 @@ func (s *Reindexer) Start(ctx context.Context) error {
 		s.Logger.DebugContext(ctx, s.Name+logPrefixRunLoopStarted)
 		defer s.Logger.DebugContext(ctx, s.Name+logPrefixRunLoopStopped)
 
-		// On each run, we calculate the new schedule based on the previous run's
-		// start time. This ensures that we don't accidentally skip a run as time
-		// elapses during the run.
```
brandur, line 110: Moved this comment down below because I think it's more salient down there where the second+ next run ats are being calculated.

internal/maintenance/reindexer.go
```diff
@@ -105,31 +105,48 @@ func (s *Reindexer) Start(ctx context.Context) error {
...
-		lastRunAt := time.Now().UTC()
+		nextRunAt := s.Config.ScheduleFunc(time.Now().UTC())
 
-		for {
-			nextRunAt := s.Config.ScheduleFunc(lastRunAt)
+		s.Logger.InfoContext(ctx, s.Name+": Scheduling first run", slog.Time("next_run_at", nextRunAt))
```
brandur, line 110: Added this log line so it's easier to observe the right thing happening from the test suite.

brandur:
@bgentry Gonna pull this one in too since it's really a very minor cosmetic change that just has the effect of making the reindexer a little more alike other maintenance services that are behaving in very similar ways. Quite uncontroversial I think.

Issue #283, opened 2024-03-21
Periodic Jobs are emitted twice than expected


RidaLiu:
Hi, I'm learning how to integrate river with current project. The use case is I have 2 scheduled tasks, task1 is running every minute, task2 is running every 2 minutes. Also they are set as unique in the queue level. 
the demo code is 
```go

func (CronJobArgs2) Kind() string { return "CronTask2" }

type CronTaskWorker2 struct {
	river.WorkerDefaults[CronJobArgs2]
	ctc    *conf.CronJob
	logger *zap.Logger
}

func NewCronTaskWorker2(ctc *conf.CronJob, logger *zap.Logger) *CronTaskWorker2 {
	return &CronTaskWorker2{ctc: ctc, logger: logger}
}

// biz logic
func (w *CronTaskWorker2) Work(ctx context.Context, job *river.Job[CronJobArgs2]) error {
	//biz logic, call repo and other biz modules
	w.logger.Info("Task 2")
	return nil
}

func (w *CronTaskWorker2) AddToRiverWorkers(workers *river.Workers) {
	w.logger.Info("Add worker2 to workers")
	if err := river.AddWorkerSafely(workers, w); err != nil {
		w.logger.Error("Add worker2 to workers", zap.String(lg.ErrorKey, err.Error()))
	}
}

// add cron exp
func (w *CronTaskWorker2) NewPeriodicJob() *river.PeriodicJob {
	schedule, err := cron.ParseStandard("*/2 * * * *")
	if err != nil {
		return nil
	}
	return river.NewPeriodicJob(schedule,
		func() (river.JobArgs, *river.InsertOpts) {
			return CronJobArgs2{}, nil
		},
		nil,
	)
}
func (CronTaskWorker2) InsertOpts() river.InsertOpts {
	return river.InsertOpts{
		UniqueOpts: river.UniqueOpts{
			ByQueue: true,
		},
	}
}

```

But when I create the river client and start, I fount that sometime the tasks are emitted twice near the schedule time point. I just query the data from river_job. 
like the screenshot below, task1 and task2 are emitted twice at the same schedule timestamp. I don't have much clue to debug it. 
![Snip20240321_19](https://github.com/riverqueue/river/assets/122007139/45fec6ce-aea2-47af-a86b-7749415f38e7)

brandur:
Just before anything else, could you confirm you're on version >= 0.0.17? There was a bug specifically around the interaction of periodic and unique jobs that was fixed there.

[...]

PR #284, opened 2024-03-23
Periodic job enqueuer: Use original next run time when calculating next schedule

brandur:
This one came up while I was looking into #283. Although I don't think
it'll fully resolve what's being reported there, this implementation
feels a little more correct compared to what we had originally.

Currently, when scheduling the next run of a job, we use the current
time. However, the current time may be somewhat staggered from when the
job was supposed to run at because we build in a 100 ms "ready margin"
when looking for jobs to run. Although a minor problem, this makes it
conceivably possible for a periodic job to drift 100 ms at a time until
it's deviated from its theoretically ideal run schedule by a non-trivial
margin (well, debatably, your definition of trivial depends a lot on
schedule tolerance).

Here, make a minor change that uses the job's original run target time
when calculating the next one. This will have been within 100 ms of the
current time anyway, but is a little more accurate for what was intended.

I started writing a test case for this, but anything I put in is so
intrusive that it feels like the noise added to code is bad enough that
it outweighs any additional marginal benefit from the test, so I just
opted not to have a test. The existing ones don't differentiate a
schedule based on current time versus original run time, but they do vet
that the change still works.

### DIFF
--- a/internal/maintenance/periodic_job_enqueuer.go
+++ b/internal/maintenance/periodic_job_enqueuer.go
@@ -168,7 +168,11 @@ func (s *PeriodicJobEnqueuer) Start(ctx context.Context) error {
 						continue
 					}
 
-					periodicJob.nextRunAt = periodicJob.ScheduleFunc(now)
+					// Although we may have inserted a new job a little
+					// preemptively due to the margin applied above, try to stay
+					// as true as possible to the original schedule by using the
+					// original run time when calculating the next one.
+					periodicJob.nextRunAt = periodicJob.ScheduleFunc(periodicJob.nextRunAt)
 
 					if insertParams, uniqueOpts, ok := s.insertParamsFromConstructor(ctx, periodicJob.ConstructorFunc); ok {
 						if !uniqueOpts.IsEmpty() {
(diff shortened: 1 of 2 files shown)


brandur:
@bgentry Wrote a fair bit of text here, but quite a small change. Seem okay?

internal/maintenance/periodic_job_enqueuer.go
```diff
@@ -168,7 +168,11 @@ func (s *PeriodicJobEnqueuer) Start(ctx context.Context) error {
 						continue
 					}
 
-					periodicJob.nextRunAt = periodicJob.ScheduleFunc(now)
+					// Although we may have inserted a new job a little
+					// assertively due to the margin applied above, try to stay
```
bgentry, line 172:
```suggestion
					// preemptively due to the margin applied above, try to stay
```

bgentry (approved):
Absolutely seems more correct! I suppose we could construct a test case for this using a synthetic test clock? Might be tricky though.

brandur:
> Absolutely seems more correct! I suppose we could construct a test case for this using a synthetic test clock? Might be tricky though.

Yeah, I actually wrote one that worked, but then the trouble was that I was now accessing interval state of the periodic enqueuer, which means that I could potentially race with the enqueuer accessing its own internals. So then I would've had to add a mutex that was only needed because of the test case, and it kind of snowballed like that until I was like "this just isn't worth it".

PR #288, opened 2024-03-27
Add capability to schedule periodic jobs dynamically

### LINKED ISSUE #242: Add periodic jobs dynamically
Based on the current documentation, it seems like it's only possible to add periodic jobs during client creation. It would be nice to be able add periodic jobs after the client has been created, similar to what you can do with scheduled jobs.

Why?

It might be a niche use case, but I have this requirement to react to events in a message driven system, and based on those events I want to create or delete periodic jobs dynamically. river would be a perfect use case for this if this feature was available.

brandur:
Here, attempt to resolve #242 by providing a way for new periodic jobs
to be added dynamically after a client's already started (i.e. not only
by passing them to the client's initial constructor).

The new API puts all functions on a separate `PeriodicJobs` bundle to
avoid further polluting `Client`'s namespace, and to make all related
functions easy to find.

    riverClient.PeriodicJobs().Add(
        river.NewPeriodicJob(
            river.PeriodicInterval(15*time.Minute),
            func() (river.JobArgs, *river.InsertOpts) {
                return PeriodicJobArgs{}, nil
            },
            nil,
        ),
    )

Additions return a periodic job "handle" which can be used to remove an
added job:

    periodicJobHandle := riverClient.PeriodicJobs().Add(...)

    riverClient.PeriodicJobs().Remove(periodicJobHandle)

I used a handle because the `river.PeriodicJob` construct isn't
particularly pretty to keep around, and contains a number of fields that
are aren't comparable so its unsuitable for use in equality checks or as
a map key.

Adding a new periodic job bumps the enqueuer's run loop so that it
enqueues the job immediately if it's configured with `RunOnStart`, and
schedules its initial target run time. In other words, adding or
removing periodic jobs should take effect ~instantly.

Fixes #242.


periodic_job.go
```diff
@@ -0,0 +1,189 @@
+package river
```
brandur, line 1: Renamed this from `periodic.go` to `periodic_job.go` to be a bit more descriptive.

periodic_job.go
```diff
@@ -0,0 +1,189 @@
...
+// Remove removes a periodic job, cancelling all scheduled runs.
+//
+// Requires the use of the periodic job handle that was returned when the job
+// was added.
+//
+// Adding or removing periodic jobs has no effect unless this client is elected
+// leader because only the leader enqueues periodic jobs. To make sure that a
+// new periodic job is fully enabled or disabled, it should be added or removed
+// from _every_ active River client across all processes.
+func (b *PeriodicJobBundle) Remove(periodicJobHandle rivertype.PeriodicJobHandle) {
```
brandur, line 152: Debated on whether this should return (1) nothing, (2) an error, (3) an OK boolean, or even (4) panic when a job handle isn't found.

Decided to go with no return value because I figure an unknown job will be an edge most people don't have to handle, and it mirrors the Go API of `delete(map, key)`.

internal/maintenance/periodic_job_enqueuer_test.go
```diff
@@ -264,39 +259,226 @@ func TestPeriodicJobEnqueuer(t *testing.T) {
...
+		wg.Wait()
+	})
+
 	t.Run("NoJobsConfigured", func(t *testing.T) {
 		t.Parallel()
 
 		svc, _ := setup(t)
 
-		svc.periodicJobs = []*PeriodicJob{}
+		// TODO: How about just don't configure the jobs in the first place.
```
bgentry, line 477: missed TODO?

bgentry (approved):
Looks great! Only minor comments ✨

periodic_job.go
```diff
@@ -0,0 +1,189 @@
...
+// Remove removes many periodic jobs, cancelling all scheduled runs.
+//
+// Requires the use of the periodic job handles that were returned when the jobs
+// were added.
+//
+// Adding or removing periodic jobs has no effect unless this client is elected
+// leader because only the leader enqueues periodic jobs. To make sure that a
+// new periodic job is fully enabled or disabled, it should be added or removed
+// from _every_ active River client across all processes.
+func (b *PeriodicJobBundle) RemoveManyAll(periodicJobHandles []rivertype.PeriodicJobHandle) {
```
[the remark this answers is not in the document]
brandur: Damn, good catch. Totally a refactor problem where I renamed something, but apparently left an outdated artifact in there. That was close ...

internal/maintenance/periodic_job_enqueuer_test.go
brandur: Yep, oops. I ended up changing the setup function to not add default jobs, but forgot to fix this.

brandur:
Thanks!

PR #293, opened 2024-03-28
Change return value of `InsertMany` from `int64` to more conventional `int` (breaking)

brandur:
Here, change the type of `Client.InsertMany` functions from `int64`s to
a more conventional `int`. This isn't really super necessary, but the
`int64` was an artifact of an internal value returned by sqlc, and using
`int` instead (which is also an int64 on the vast majority of all
system) is more conventional.

Because callers almost always know the size of the slice they passed to
`InsertMany`, or can ascertain it easily, I expect this change to break
very few integrations in practice. As we can see from our own test
suite, little had to change because `InsertMany`'s return value is
rarely needed.


brandur:
To ship out alongside https://github.com/riverqueue/river/pull/236 and https://github.com/riverqueue/river/pull/292 in a single new version containing multiple minor breaking changes.

bgentry (approved):
Needs changelog but LGTM :v:

brandur:
Great! Changelog added, but will hold until we've got all the breaking changes ready.

PR #297, opened 2024-04-14
Add `rivertype.JobStates` with full list of job states

### LINKED ISSUE #296: Add `func JobStates() []JobState` in `rivertype` package
Motivation: we just wrote a translation layer from `JobState` to an internal type, and in writing unit tests for that translation, wanted a way to ensure we pass in all possible inputs.

We're happy to send a PR to add this if you guys are OK with adding it to the exported surface of this package.

brandur:
Requested in #296, add a new `JobStates()` function that contains a
complete list of all `JobState*` constant values, which may be useful in
places like testing.

Also add a test that does a basic parsing of the corresponding Go file,
look for all `JobState*` values, and makes sure that `JobStates()`
contains every possible value to prevent future regressions.

Fixes #296.

### DIFF
--- a/rivertype/river_type_test.go
+++ b/rivertype/river_type_test.go
@@ -1,0 +1,81 @@
+package rivertype_test
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/riverqueue/river/rivertype"
+)
+
+func TestJobStates(t *testing.T) {
+	t.Parallel()
+
+	jobStates := rivertype.JobStates()
+
+	// One easy check that doesn't require the source file reading below.
+	require.Contains(t, jobStates, rivertype.JobStateAvailable)
+
+	// Get all job state names from the corresponding source file and make sure
+	// they're included in JobStates. Helps check that we didn't add a new value
+	// but forgot to add it to the full list of constant values.
+	for _, nameAndValue := range allValuesForStringConstantType(t, "river_type.go", "JobState") {
+		t.Logf("Checking for job state: %s / %s", nameAndValue.Name, nameAndValue.Value)
+		require.Contains(t, jobStates, rivertype.JobState(nameAndValue.Value))
+	}
+}
+
+// stringConstantNameAndValue is a name and value for a string constant like
+// `JobStateAvailable` + `available`.
+type stringConstantNameAndValue struct{ Name, Value string }
+
+// allValuesForStringConstantType reads a Go source file and looks for all
+// values for the named string constant.
+func allValuesForStringConstantType(t *testing.T, srcFile, typeName string) []stringConstantNameAndValue {
+	t.Helper()
+
+	fset := token.NewFileSet()
+
+	src, err := os.ReadFile(srcFile)
+	require.NoError(t, err)
+
+	f, err := parser.ParseFile(fset, srcFile, src, parser.ParseComments)
+	require.NoError(t, err)
+
+	var valueNames []stringConstantNameAndValue
+
+	for _, decl := range f.Decls {
+		if gen, ok := decl.(*ast.GenDecl); ok && gen.Tok == token.CONST {
+			for _, spec := range gen.Specs {
+				// Always ast.ValueSpec for token.CONST.
+				valueSpec := spec.(*ast.ValueSpec) //nolint:forcetypeassert
+
+				typeIdent, ok := valueSpec.Type.(*ast.Ident)
+				if !ok || typeIdent.Name != typeName {
+					continue
+				}
+
+				for i, nameIdent := range valueSpec.Names {
+					// Force type assert because we expect one of our constants
+					// to be defined as a basic type literal like this.
+					basicLitExpr := valueSpec.Values[i].(*ast.BasicLit) //nolint:forcetypeassert
+
+					valueNames = append(valueNames, stringConstantNameAndValue{
+						Name:  nameIdent.Name,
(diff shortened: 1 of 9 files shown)


rivertype/river_type_test.go
```diff
@@ -0,0 +1,81 @@
...
+	jobStates := rivertype.JobStates()
+
+	// One easy check that doesn't require the source file reading below.
+	require.Contains(t, jobStates, rivertype.JobStateAvailable)
+
+	// Get all job state names from the corresponding source file and make sure
+	// they're included in JobStates. Helps check that we didn't add a new value
+	// but forgot to add it to the full list of constant values.
+	for _, nameAndValue := range allValuesForStringConstantType(t, "river_type.go", "JobState") {
+		t.Logf("Checking for job state: %s / %s", nameAndValue.Name, nameAndValue.Value)
```
brandur, line 27:
```
=== RUN   TestJobStates
=== PAUSE TestJobStates
=== CONT  TestJobStates
    river_type_test.go:27: Checking for job state: JobStateAvailable / available
    river_type_test.go:27: Checking for job state: JobStateCancelled / cancelled
    river_type_test.go:27: Checking for job state: JobStateCompleted / completed
    river_type_test.go:27: Checking for job state: JobStateDiscarded / discarded
    river_type_test.go:27: Checking for job state: JobStateRetryable / retryable
    river_type_test.go:27: Checking for job state: JobStateRunning / running
    river_type_test.go:27: Checking for job state: JobStateScheduled / scheduled
--- PASS: TestJobStates (0.00s)
```

bgentry:
Note that #301 adds a new `pending` state, so this function and its tests will need to be updated to include that new state. I've added a TODO to that PR for this.

rivertype/river_type_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	jobStates := rivertype.JobStates()
+
+	// One easy check that doesn't require the source file reading below.
+	require.Contains(t, jobStates, rivertype.JobStateAvailable)
+
+	// Get all job state names from the corresponding source file and make sure
+	// they're included in JobStates. Helps check that we didn't add a new value
+	// but forgot to add it to the full list of constant values.
+	for _, nameAndValue := range allValuesForStringConstantType(t, "river_type.go", "JobState") {
```
bgentry, line 26: Great idea!

rivertype/river_type_test.go
```diff
@@ -0,0 +1,81 @@
...
+						Name:  nameIdent.Name,
+						Value: basicLitExpr.Value[1 : len(basicLitExpr.Value)-1], // strip quote on either side
+					})
+				}
+			}
+		}
+	}
+
+	if len(valueNames) < 1 {
+		require.FailNow(t, "Not values found", "No values found for source file and constant type: %s / %s", srcFile, typeName)
```
bgentry, line 77: typo?
```suggestion
		require.FailNow(t, "No values found", "No values found for source file and constant type: %s / %s", srcFile, typeName)
```

bgentry (approved):

brandur:
> Note that #301 adds a new `pending` state, so this function and its tests will need to be updated to include that new state. I've added a TODO to that PR for this.

Ah good to know. Welp, tests should fail, so hopefully we'll be forced to get this right.

rivertype/river_type_test.go
brandur: Yep, or probably Vim-o as I was trying to do something and accidentally added/replaced a character. Thanks!

brandur:
Thx.

dhermes:
Thanks for doing this @brandur! I was thinking you'd end up implementing this as something like a check against the output of `\dT+ river_job_state`. E.g. you could wire this query up with `sqlc` (thanks `psql --echo-hidden`!!):

```sql
SELECT
  n.nspname AS schema,
  e.enumlabel AS enum_value
FROM
  pg_catalog.pg_type AS t
  LEFT JOIN pg_catalog.pg_namespace AS n ON n.oid = t.typnamespace
  LEFT JOIN pg_catalog.pg_enum AS e ON e.enumtypid = t.oid
WHERE
  t.typname = 'river_job_state'
ORDER BY e.enumsortorder;
```

In our setup, where we put you guys in `riverqueue` schema, this produces

```
+------------+------------+
| schema     | enum_value |
|------------+------------|
| riverqueue | available  |
| riverqueue | cancelled  |
| riverqueue | completed  |
| riverqueue | discarded  |
| riverqueue | retryable  |
| riverqueue | running    |
| riverqueue | scheduled  |
+------------+------------+
```

bgentry:
@dhermes this would be fine for a unit test, though I’m not sure it’s worth the effort. I wouldn’t want to base the actual API on that query though (not sure if this is what you were suggesting).

While I don’t expect to add more states beyond this, #301 does add a new `pending` state. If you were to deploy the new code but forget to run the migration, River’s API should still tell you what all the actual states are that it knows about IMO.

dhermes:
> I wouldn't want to base the actual API on that query though (not sure if this is what you were suggesting).

Haha sorry I was **not** trying to suggest that. Definitely not.

> this would be fine for a unit test, though I'm not sure it's worth the effort. 

I've got some existing unit tests like this in our codebase that leverage `sqlc` so will be pretty low effort.

For me, it's just about trying to get the best of both worlds of Go and PostgreSQL. I.e. using the right native SQL / PostgreSQL capabilities for enums, and the right Go capabilities for enums. And then just making sure we have a guarantee at dev time that the native Go and native PostgreSQL are agreeing with each other.

If we're not aligned on this, all good. I'm not trying to step on any toes, it's your project! (And we love this project, it's awesome!)

PR #298, opened 2024-04-15
Add concurrent safe `StubTime` test helper

brandur:
This one's not strictly necessary, but it came up in another project so
I thought I may as well port a similar feature here. There's currently a
test helper for setting a stub time in tests, but it's not very useful,
and not useful in case a test case is setting a new test time while a
service is simultaneously trying to access it.

Here, replace `StubTime` with a new test helper that returns a
`(getTime, setTime)` function pair in which each invocation is protected
by a mutex so the concurrency problem described above is solved.

As seen here, this helper isn't used all that much and could plausibly
be removed or refactored out. However, there's no particularly good
reason to do that since it's not causing any trouble so I think
augmenting it is just as good of a potential path.


brandur:
This one's a really minor test-only change, so in the interest of getting a lot of PRs done, going to time out and merge it.

PR #300, opened 2024-04-15
Remove `river.JobState*` aliases in favor of using `rivertype` directly

brandur:
One of the accidental API misdesigns I forgot to address on the initial
version of River was that type aliases for `JobState*` had been left in
the top-level `river` namespace.

Having them in `river` may make use slightly more convenient, but makes
"jump to definition" worse because it requires an additional redirect,
and the decision to use `river` or `rivertype` ambiguous in that there's
no canonically correct answer. The aliases also make adding a new state
a little more risky because the alias may be forgotten.

I wouldn't normally make this change post-ship, but given we're doing a
couple breaking changes all at once, it's our last, best opportunity to
fix the (mild) problem.


bgentry (approved):
I wish we didn't end up having to put these in a sub-package to begin with, but since we did, this is the right call.

Do you think there are any website or other doc fixes required for this?

brandur:
> Do you think there are any website or other doc fixes required for this?

Nah, I think we're okay there (also no example tests needed to be updated).

```
$ cd /Users/brandur/Documents/projects/river-homepage
$ ag 'river\.JobState'
$
```

brandur:
With #304 in, going to start merging these other changes prior to release.

PR #301, opened 2024-04-15
v4 migration: App level notifications, queue status tracking, pause/resume

bgentry:
There's a lot in this PR, all of it somewhat related and dependent on the new v4 database migration.

## v4 migration summary

* Make `river_job.args` non-null (there should be no null values in here anyway)
* Make `river_job.metadata` non-null (there should be no null values in here anyway)
* Drop job insert trigger notifications so this can be done at the application level
* Stricter `finalized_at` constraint
* Add `pending` job state. Currently unused, but will be used to build higher level functionality by staging jobs that are not yet ready to run (for some reason other than their scheduled time being in the future).
* Add a `river_queue` table for queue state tracking, pause/resume.

I believe all of these are totally safe to perform on an actively running River installation with no interruptions. We should carefully consider this though. The removal of insert triggers does increase new job execution latency until a subsequent deploy of updated River clients with app-level notification logic, but this is a minor and temporary degradation.

## Application level insert notifications

The initial design for River utilized a trigger on job insert that issued notifications (`NOTIFY`) so that listening clients could quickly pick up the work if they were idle. While this is good for lowering latency, it does have the side effect of emitting a large amount of notifications any time there are lots of jobs being inserted. This adds overhead, particularly to high-throughput installations.

To improve this situation and reduce overhead in high-throughput installations, the notifications have been refactored to be emitted at the application level. A client-level debouncer ensures that these notifications are not emitted more often than they could be useful. If a queue is due for an insert notification (on a particular Postgres schema), the notification is piggy-backed onto the insert query within the transaction. While this has the impact of increasing insert latency for a certain percentage of cases, the effect should be small.

Additionally, the initial release of River did not properly scope notification topics within the global `LISTEN/NOTIFY` namespace. If two River installations were operating on the same Postgres database but within different schemas (search paths), their notifications would be emitted on a shared topic name. This is no longer the case and all notifications are prefixed with a `{schema_name}.` string.

## Queue status tracking, pause and resume

A useful operational lever is the ability to pause and resume a queue without shutting down clients. In addition to pause/resume being a feature request from #54, as part of my work on River's UI I've encountered the need to list out the active queues so that they can be displayed and manipulated.

A new `river_queue` table is introduced in the migration for this purpose. Upon startup, every producer in each River `Client` will make an `UPSERT` query to the database to either register the queue as being active, or if it already exists it will instead bump the timestamp to keep it active. This query will be run periodically as long as the `Client` is alive, even if the queue is paused. A separate query will delete/purge any queues which have not been active in awhile (currently fixed to 24 hours). This currently happens independently in each producer, but it really doesn't need to (see TODO below).

This table provides a place to pause and resume a single queue by name, or _all_ queues using the special `*` value. Each producer will watch for notifications on the relevant `LISTEN/NOTIFY` topic unless operating in poll-only mode, in which case they will periodically poll for changes to this record in the database.

## TODOs

* [x] Make queue purging a standalone maintenance process not connected to the producers.
* [x] Changelog entry
* [x] update `rivertype.JobStates` to include `pending` state

Closes #54.


bgentry:
Nice, I made some race conditions 😆  Well actually they appear to be just in the tests as a result of the `SharedTx` thing we were just discussing, not necessarily a problem with the actual code.

internal/maintenance/queue_cleaner.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/riverqueue/river/internal/maintenance/startstop"
+	"github.com/riverqueue/river/internal/rivercommon"
+	"github.com/riverqueue/river/internal/util/timeutil"
+	"github.com/riverqueue/river/internal/util/valutil"
+	"github.com/riverqueue/river/riverdriver"
+)
+
+const (
+	QueueCleanerIntervalDefault = time.Hour
+	QueueRetentionPeriodDefault = 24 * time.Hour
```
brandur, line 21: Add "cleaner" to the name of this one, and for the time being, maybe unexport both constants. Neither are needed outside the packages and IMO no reason to ever really make these configurable unless there's a really amazing reason to do so.

internal/maintenance/queue_cleaner.go
```diff
@@ -0,0 +1,162 @@
...
+func (ts *QueueCleanerTestSignals) Init() {
+	ts.DeletedBatch.Init()
+}
+
+type QueueCleanerConfig struct {
+	// Interval is the amount of time to wait between runs of the cleaner.
+	Interval time.Duration
+	// QueueRetentionPeriod is the amount of time to keep queues around before
+	// they're removed.
+	QueueRetentionPeriod time.Duration
```
brandur, line 38: Maybe drop this one down to just `RetentionPeriod`. The additional "Queue" stutters with what's implied by the struct's name.

bgentry:
@brandur alright, I think your fix in #311 did the trick on all the `SharedTx` issues, so this is good for review.

Some notes: we did get a single flaky failure on [this run](https://github.com/riverqueue/river/actions/runs/8774845966/job/24076490957) on the `Client` start/stop stress test. And there are some other issues to work through in [this latest run](https://github.com/riverqueue/river/actions/runs/8774870619/job/24076543522?pr=301).

I have to break for food but will pick back up on those shortly. Feel free to review meanwhile if you're available.

internal/riverinternaltest/riverdrivertest/riverdrivertest.go
```diff
@@ -792,6 +795,88 @@ func ExerciseExecutorFull[TTx any](ctx context.Context, t *testing.T, driver riv
 			require.Equal(t, rivertype.JobStateCompleted, job.State)
 			require.Equal(t, []string{"tag"}, job.Tags)
 		})
+
+		// TODO(bgentry): these are probably in the wrong file or location within
+		// the file and should go somewhere else?
+		t.Run("JobFinalizedAtConstraint", func(t *testing.T) {
```
brandur, line 801: RE TODO: If you feel it may still be wrong, should probably try to find a new convention for where to put it. Should probably find a way to resolve this TODO before merge since it's fairly trivial.

internal/riverinternaltest/testfactory/test_factory.go
```diff
@@ -36,6 +36,14 @@ type JobOpts struct {
 func Job(ctx context.Context, tb testing.TB, exec riverdriver.Executor, opts *JobOpts) *rivertype.JobRow {
 	tb.Helper()
 
+	job, err := exec.JobInsertFull(ctx, JobBuild(tb, opts))
+	require.NoError(tb, err)
+	return job
+}
+
+func JobBuild(tb testing.TB, opts *JobOpts) *riverdriver.JobInsertFullParams {
```
brandur, line 44: I was thinking about doing something like this before, but I think we need to establish a different convention, because as is, the naming suggests that this is a factory function for a "job build" entity.

Maybe something like `Job_Build` or `Job_InsertParams`?

internal/riverinternaltest/testfactory/test_factory.go
```diff
@@ -111,3 +117,26 @@ var seq int64 = 1 //nolint:gochecknoglobals
...
+	tb.Helper()
+
+	metadata := opts.Metadata
+	if opts.Metadata == nil {
+		metadata = []byte("{}")
+	}
+
+	queue, err := exec.QueueCreateOrSetUpdatedAt(ctx, &riverdriver.QueueCreateOrSetUpdatedAtParams{
+		Metadata:  metadata,
+		Name:      ptrutil.ValOrDefault(opts.Name, "default"),
```
brandur, line 137: Could you change this to?

``` go
		Name:      ptrutil.ValOrDefaultFunc(opts.Name, func() string { return fmt.Sprintf("queue_%02d", nextSeq()) }),
```

The basic idea with these factory functions is you can call them many times with no (or very few) opts and still get a new valid object back. See migrations above for example, which _could_ always returning migration version 1, but generates a sequence instead for caller convenience.

This would also allow you to simplify your tests quite a bit, many of which have to override `Name` to work around this default.

brandur, line 137: Could you change this to?

``` go
		Name:      ptrutil.ValOrDefaultFunc(opts.Name, func() string { return fmt.Sprintf("queue_%02d", nextSeq()) }),
```

The basic idea with these factory functions is you can call them many times with no (or very few) opts and still get a new valid object back. See migrations above for example, which _could_ always returning migration version 1, but generates a sequence instead for caller convenience.

This would also allow you to simplify your tests quite a bit, many of which have to override `Name` to work around this default.

producer.go
```diff
@@ -295,6 +332,51 @@ type insertPayload struct {
 	Queue string `json:"queue"`
 }
 
+func (p *producer) handleJobControlNotification(workCtx context.Context) func(notifier.NotificationTopic, string) {
```
brandur, line 335: Okay, so if I understand correctly, the "job control" topic is shared between (1) job cancel notifications, and (2) queue pause/resume notifications.

It's not super clear to me what these have in common, except that they're both topics used by the producer. But job inserts are also topics shared by the producer, so by that logic, shouldn't they also share the same topic?

producer.go
```diff
@@ -283,6 +318,8 @@ type jobControlAction string
 
 const (
 	jobControlActionCancel jobControlAction = "cancel"
+	jobControlActionPause  jobControlAction = "pause"
+	jobControlActionResume jobControlAction = "resume"
```
brandur, line 322: If all these keep sharing, we definitely want a rename of these at least, e.g.:

`Cancel` -> `JobCancel`
`Pause` -> `QueuePause`
`Resume` -> `QueueResume`

producer.go
```diff
@@ -470,6 +577,84 @@ func (p *producer) handleWorkerDone(job *rivertype.JobRow) {
...
+
+func (p *producer) fetchQueueSettings(ctx context.Context) (*rivertype.Queue, error) {
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+
+	return p.exec.QueueGet(ctx, p.config.Queue)
+}
+
+func (p *producer) reportQueueStatusLoop(ctx context.Context) {
+	// TODO(bgentry): initial randmized sleep?
```
brandur, line 627: Can this be resolved? Maybe just use one of the cancellable sleep helpers with a small jitter even if we're not totally married to the exact timing. Feels better than leaving the TODO in.

producer_test.go
```diff
@@ -410,4 +438,126 @@ func testProducer(t *testing.T, makeProducer func(ctx context.Context, t *testin
...
 		startstoptest.Stress(ctx, t, producer)
 	})
+
+	t.Run("QueuePausedBeforeStart", func(t *testing.T) {
+		t.Parallel()
+
+		producer, bundle := setup(t)
+		AddWorker(bundle.workers, &noOpWorker{})
+
+		// TODO: maybe add a separate QueueInsertFull to driver to clean this up?
```
brandur, line 448: IMO: while there's some argument that jobs should have a fast path because we want them highly performant for benchmarks and the like, one extra insert parameter on queues is going to have ~zero practical performance impact. Maybe just add any desired ones to the normal insert path as optional parameters.

queue_pause_opts.go
```diff
@@ -0,0 +1,4 @@
+package river
+
+// QueuePauseOpts are optional settings for pausing or unpausing a queue.
```
brandur, line 3: I've been struggling with this myself just writing comments here, but for purposes of docs we should try to consolidate on one term. Either "resume" or "unpause", but try to stay consistent.

riverdriver/riverpgxv5/internal/dbsqlc/pg_misc.sql
```diff
@@ -1,5 +1,15 @@
...
-SELECT pg_notify(@topic, @payload);
\ No newline at end of file
+-- name: PGNotifyMany :exec
+WITH topic_to_notify AS (
+    SELECT
+        concat(current_schema(), '.', @topic::text) AS topic,
+        unnest(@payload::text[]) AS payload
+)
+
+SELECT pg_notify(
```
brandur, line 11: Minor, but any chance we could try to by convention omit the newlines between CTE cases? Every time I see a `SELECT` like this with a bare line above it, I read it as a standalone expression before realizing there's CTE(s) up there and I have to read upwards to get the whole context.

riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
```diff
@@ -290,11 +297,10 @@ river_job_scheduled AS (
...
     RETURNING *
 )
-SELECT count(*)
-FROM (
-    SELECT pg_notify(@insert_topic, json_build_object('queue', queue)::text)
-    FROM river_job_scheduled
-) AS notifications_sent;
+SELECT
+    queue,
+    scheduled_at
```
brandur, line 302: What do you think about having this return just the full job row instead? There's a minor performance implication, but keeps things more consistent and requires fewer bespoke types in the driver machinery.

internal/riverinternaltest/riverdrivertest/riverdrivertest.go
```diff
@@ -1646,6 +1729,256 @@ func ExerciseExecutorFull[TTx any](ctx context.Context, t *testing.T, driver riv
...
+				t.Parallel()
+
+				exec, _ := setupExecutor(ctx, t, driver, beginTx)
+
+				queue1 := testfactory.Queue(ctx, t, exec, &testfactory.QueueOpts{Name: ptrutil.Ptr("queue1")})
+				require.Nil(t, queue1.PausedAt)
+				queue2 := testfactory.Queue(ctx, t, exec, &testfactory.QueueOpts{Name: ptrutil.Ptr("queue2")})
+				require.Nil(t, queue2.PausedAt)
+
+				require.NoError(t, exec.QueuePause(ctx, "*"))
```
brandur, line 1916: Could we use some kind of global constant for this, even if it's just for use inside River's internal code?

I was trying to look for how wildcard pause/resume worked, and found this very awkward to search for. It's very short for one, there's no constant so you can't "find references", and "*" is a regex meaningful character of course, so searching for it is a little extra painful as you need to put in the right escapes.

internal/maintenance/queue_cleaner.go
```diff
@@ -0,0 +1,162 @@
...
+	for {
+		// Wrapped in a function so that defers run as expected.
+		queuesDeleted, err := func() ([]string, error) {
+			ctx, cancelFunc := context.WithTimeout(ctx, 30*time.Second)
+			defer cancelFunc()
+
+			queuesDeleted, err := s.exec.QueueDeleteExpired(ctx, &riverdriver.QueueDeleteExpiredParams{
+				Max:              s.batchSize,
+				UpdatedAtHorizon: time.Now().Add(-s.Config.QueueRetentionPeriod),
+			})
```
brandur, line 135: Am I understanding correctly that when you use the global pause/resume of "*", you can never pause all queues for more than 24 hours?

Unlike a normal queue, River clients won't report on the global queue "*", so it won't have it's `updated_at` field bumped unless the user pauses/resumes again. So after 24 hours it gets reaped here even if the user never resumed an initial pause.

brandur (commented):
Great!!! Lots of useful new stuff in here. River's going to be looking pretty feature complete with this in ...

client.go
```diff
@@ -44,6 +45,9 @@ const (
 	PriorityDefault    = rivercommon.PriorityDefault
 	QueueDefault       = rivercommon.QueueDefault
 	QueueNumWorkersMax = 10_000
+
+	queueSettingsPollIntervalDefault   = 2 * time.Second
+	queueSettingsReportIntervalDefault = 10 * time.Minute
```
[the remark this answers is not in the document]
bgentry: Sure, I pulled these up into the client initially expecting to expose them but talked myself out of that. We can push them back down for now until it becomes clear they need to be exposed.

internal/maintenance/queue_cleaner.go
bgentry: That is certainly not the intention, no. The intention is that as long as `Client` is configured to work a queue and is actively started, the `updated_at` timestamp should be regularly bumped. The queue should never be deleted by the cleaner process unless no started Clients have been configured to use it within the past 24 hours.

The queries for pausing and resuming use a `CASE` to conditionally filter the `river_queue` rows by `name` only when `name != '*'`, so if `*` is the provided value, _all_ registered job queues are paused. However if you were to then start a new `Client` with a brand new queue name registered, it would not be paused because it was not in the table at the time you paused everything.

Can you do me a favor and specifically look at whether this is clear to you when you now look at the code? And if not, where might we improve that or better document it?

client.go
```diff
@@ -1309,10 +1335,23 @@ func (c *Client[TTx]) insert(ctx context.Context, exec riverdriver.Executor, arg
 		return nil, err
 	}
 
-	jobInsertRes, err := c.uniqueInserter.JobInsert(ctx, exec, params, uniqueOpts)
+	execTx, err := exec.Begin(ctx)
```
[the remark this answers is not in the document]
bgentry: I'm fine with that. We might have some scattered remnants of the name `execTx` (don't think I came up w/ that name myself) so I can fix the ones I introduced in this PR and then add a separate commit to fix preexisting ones.

client.go
```diff
@@ -1423,6 +1489,93 @@ func (c *Client[TTx]) insertManyParams(params []InsertManyParams) ([]*riverdrive
...
+func (c *Client[TTx]) maybeNotifyInsert(ctx context.Context, execTx riverdriver.ExecutorTx, state rivertype.JobState, queue string) error {
+	if state != rivertype.JobStateAvailable {
+		return nil
+	}
+	return c.maybeNotifyInsertForQueues(ctx, execTx, []string{queue})
+}
+
+func (c *Client[TTx]) maybeNotifyInsertMany(ctx context.Context, execTx riverdriver.ExecutorTx, insertParams []*riverdriver.JobInsertFastParams) error {
+	queues := sliceutil.Map(insertParams, func(params *riverdriver.JobInsertFastParams) string { return params.Queue })
+	return c.maybeNotifyInsertForQueues(ctx, execTx, queues)
```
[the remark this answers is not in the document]
bgentry: Excellent catch, thank you. I added a TODO for the moment because I want to make sure I get test coverage on this, will for sure fix that before merging.

client.go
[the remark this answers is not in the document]
bgentry: Actually there are quite a few `execTx` references, mind if we fix all the unrelated ones in a subsequent PR?

client.go
```diff
@@ -1528,6 +1681,83 @@ func (c *Client[TTx]) JobListTx(ctx context.Context, tx TTx, params *JobListPara
...
+func (c *Client[TTx]) QueueGet(ctx context.Context, name string) (*rivertype.Queue, error) {
+	return c.driver.GetExecutor().QueueGet(ctx, name)
+}
+
+// QueueList returns a list of all queues that are currently active or were
+// recently active. Limit and offset can be used to paginate the results.
+//
+// The provided context is used for the underlying Postgres query and can be
+// used to cancel the operation or apply a timeout.
+func (c *Client[TTx]) QueueList(ctx context.Context, limit, offset int) ([]*rivertype.Queue, error) {
```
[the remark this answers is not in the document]
bgentry: I definitely didn't want to take on something like cursor based pagination here. I think the compromise you describe is reasonable, however do you really think we shouldn't even have a param for the limit that we want returned? Even with a cursor based pagination story you'll still have that, right?

internal/maintenance/queue_cleaner.go
bgentry: `QueueCleanerRetentionPeriodDefault` (née `QueueRetentionPeriodDefault`) is actually referenced from related client tests in two places, and in the `Client` maintenance initialization. Would you prefer that this config not be exposed and that the `QueueCleaner` apply its own default without exposing that value to the main package?

internal/riverinternaltest/riverdrivertest/riverdrivertest.go
bgentry: Yes, would like to resolve this and other TODOs before merging. Left this note bc I didn't want to lose focus/flow on what I was doing at the time.

What do you think? Does it seem ok for this to live within `JobInsertFull`? The challenge is that we may not be able to control the params we need to using any other insert mechanism, so maybe it is in fact the most appropriate place?

producer.go
bgentry: The topic naming is not ideal for this. We could definitely keep the actual _job_ control topic separate from these messages which are more of _queue_ control; the only downsides to that are another constant and another subscription that must be initialized when each producer starts.

Another option is to pick a new dual purpose name.

internal/riverinternaltest/testfactory/test_factory.go
bgentry: Great, thank you, ended up removing basically all explicitly specified queue names in tests.

producer.go
bgentry: Yes, didn't mean to leave this in, but did mean to ask you about the best way to solve this. I'm realizing now that `producer` is just a service, I can use `s.CancellableSleepRandomBetween()`, right?

producer_test.go
bgentry: Yep great call, fixed and cleaned up a few places that were separately re-pausing (except those which were explicitly testing the act of pausing).

client_test.go
```diff
@@ -461,6 +458,117 @@ func Test_Client(t *testing.T) {
...
+		startedJobID = riverinternaltest.WaitOrTimeout(t, jobStartedChan)
+		require.Equal(t, insertResAlternate2.Job.ID, startedJobID)
+
+		// Resume all queues:
+		require.NoError(t, client.QueueResume(ctx, "*", nil))
+		client.producersByQueueName[QueueDefault].testSignals.Resumed.WaitOrTimeout()
+
+		startedJobID = riverinternaltest.WaitOrTimeout(t, jobStartedChan)
+		require.Equal(t, insertRes2.Job.ID, startedJobID)
+	})
```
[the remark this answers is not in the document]
bgentry: Added a simpler single-queue pause/resume test, lmk if that's what you were envisioning.

internal/riverinternaltest/testfactory/test_factory.go
bgentry: renamed to `Job_Build`

internal/riverinternaltest/riverdrivertest/riverdrivertest.go
bgentry: Refactored all instances of this to use `riverdriver.AllQueuesString`, lmk if you have a better idea. It's used across the driver and the main `river` package so the options were limited without making it an official "public" API.

client.go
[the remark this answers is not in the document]
bgentry: Added test coverage for notifying on `Available` only, both on `Insert` and `InsertMany` 🙌  Definitely a gap prior to this. Lmk if it looks alright!

client.go
```diff
@@ -1423,6 +1486,101 @@ func (c *Client[TTx]) insertManyParams(params []InsertManyParams) ([]*riverdrive
...
 
+func (c *Client[TTx]) maybeNotifyInsert(ctx context.Context, execTx riverdriver.ExecutorTx, state rivertype.JobState, queue string) error {
+	if state != rivertype.JobStateAvailable {
+		return nil
+	}
+	return c.maybeNotifyInsertForQueues(ctx, execTx, []string{queue})
+}
+
+func (c *Client[TTx]) maybeNotifyInsertMany(ctx context.Context, tx riverdriver.ExecutorTx, insertParams []*riverdriver.JobInsertFastParams) error {
+	queues := make([]string, 0, len(insertParams))
```
brandur, line 1497: Small one, but thoughts on using a map instead of slice here? It probably doesn't matter for small numbers of insert many params, but for large ones (e.g. 1k or 10k like the benchmark does), you're allocating quite a large slice here unnecessarily since the cardinality of the queues in use is likely to be a tiny number like 1 or 2.

client.go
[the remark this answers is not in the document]
brandur: Yeah, seems fine.

client.go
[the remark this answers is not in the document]
brandur: Awesome. Left one more comment about possible use of a map, but looks good.

client.go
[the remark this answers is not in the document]
brandur: Oh yeah a limit is totally fine, but what do you think about making it the same as job listing, so like `QueueListParams.First`?

client_test.go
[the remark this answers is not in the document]
brandur: Yep, exactly. Thanks.

producer.go
brandur: Yep, exactly.

client.go
bgentry: Hmm, I'd have to then iterate through the map again afterward to get its keys, right? Maybe I should just scale back the allocation and allow it to reallocate as needed for larger slices? 🤔

client.go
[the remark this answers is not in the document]
bgentry: Reworked to use a similar pagination setup to `JobList`, thanks! Left only a `First(n)` option in there for now with no other sort or filtering options.

internal/riverinternaltest/riverdrivertest/riverdrivertest.go
bgentry: Moved to `rivercommon` :v:

internal/maintenance/queue_cleaner.go
bgentry: Another way of stating this: calling `QueuePause` with `*` does not persist a queue record with the `*` value in it. Instead, it triggers the `CASE` in the pause query to avoid filtering queue rows by name and instead selects all unpaused queues for update. Those queues will continue to have their `updated_at` bumped periodically by any producer that is actively using them, so they will remain paused for as long as they are in use (plus the cleanup horizon).

client.go
```diff
@@ -1423,6 +1489,93 @@ func (c *Client[TTx]) insertManyParams(params []InsertManyParams) ([]*riverdrive
 	return insertParams, nil
 }
 
+func (c *Client[TTx]) maybeNotifyInsert(ctx context.Context, execTx riverdriver.ExecutorTx, state rivertype.JobState, queue string) error {
+	if state != rivertype.JobStateAvailable {
+		return nil
+	}
+	return c.maybeNotifyInsertForQueues(ctx, execTx, []string{queue})
```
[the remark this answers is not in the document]
bgentry: I looked at this for a bit and decided to go a slightly different direction by extracting more of the shared `InsertMany` / `InsertManyTx` logic. In doing so I lost the dedicated `maybeNotifyInsertForQueues` helper.

client.go
bgentry: Opted to start with a default capacity of 10. This is probably going to be way more than enough for almost all River installs—how often are people really going to insert jobs on more than 10 different queues at one time?

bgentry:
There are just a few items left here, mainly the name of the pubsub topic(s) we'll use for job and queue controls. Please lmk your thoughts on all the unresolved items! 🙏

producer.go
bgentry: I removed the "job" bit from the naming and it's now just "control" throughout. Let me know if you want to paint the shed another color before merge 🎨 🖌️

client.go
brandur: Yeah, should have relatively few total queues.

Looks good.

internal/maintenance/queue_cleaner.go
brandur: Ah, I see! Okay, my mistake — I misunderstood how that piece worked.

client.go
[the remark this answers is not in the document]
brandur: Perfect, thanks.

producer.go
brandur: Just a bump on this one. The topic rename looks great, but given we're still sharing, I think it makes sense to be a bit more explicit about each action by giving it a noun. Also, we can probably strip "job" from the beginning of everything now that the topic's been renamed.

What do you think of this?

``` go
const (
	controlActionCancel controlActionJobCancel   = "job_cancel"
	controlActionPause  controlActionQueuePause  = "queue_pause"
	controlActionResume controlActionQueueResume = "queue_resume"
)
```

producer.go
brandur: Heh heh. Can't have enough layers of paint! LG.

internal/riverinternaltest/riverdrivertest/riverdrivertest.go
brandur: Yeah, we may want to consider a special "DB constraints" section at some point, but this seems fine for now.

Let's just drop the TODO and reorg later if it comes up.

CHANGELOG.md
```diff
@@ -7,6 +7,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ## [Unreleased]
 
+⚠️ Version 0.5.0 contains a new database migration, version 4. This migration is backward compatible with any River installation running the v3 migration. Be sure to run the v4 migration prior to deploying the code from this release.
+
+### Added
+
+- Add `pending` job state. This is currently unused, but will be used to build higher level functionality for staging jobs that are not yet ready to run (for some reason other than their scheduled time being in the future). Pending jobs will never be run or deleted and must first be moved to another state by external code. [PR #301](https://github.com/riverqueue/river/pull/301).
+- Queue status tracking, pause and resume. [PR #301](https://github.com/riverqueue/river/pull/301).
+
+  A useful operational lever is the ability to pause and resume a queue without shutting down clients. In addition to pause/resume being a feature request from #54, as part of the work on River's UI it's been useful to list out the active queues so that they can be displayed and manipulated.
```
brandur, line 17: Sorry for the super nit, but mind linking up 54 since it won't be in the rendered changelog.
```suggestion
  A useful operational lever is the ability to pause and resume a queue without shutting down clients. In addition to pause/resume being a feature request from [#54](https://github.com/riverqueue/river/pull/54), as part of the work on River's UI it's been useful to list out the active queues so that they can be displayed and manipulated.
```

CHANGELOG.md
```diff
@@ -7,6 +7,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+⚠️ Version 0.5.0 contains a new database migration, version 4. This migration is backward compatible with any River installation running the v3 migration. Be sure to run the v4 migration prior to deploying the code from this release.
```
brandur, line 10: We may want to put together a little upgrade guide on the website because people's familiarity with the last time they migrated may have become long atrophied at this point (i.e. how to use the CLI, the fact that the CLI will need to be updated separately to know about the new migration, etc.).

brandur (approved):
> There are just a few items left here, mainly the name of the pubsub topic(s) we'll use for job and queue controls. Please lmk your thoughts on all the unresolved items! 🙏

Thanks for getting through all that! I added a couple more tiny comments, but looks solid.

Changelog looks great.

Let's get this shipped.

PR #303, opened 2024-04-16
Prepare release `v0.3.0`

brandur:
Prep release of `v0.3.0`. This is a relatively small one containing
mainly #281 and #297, but I think it's not a bad idea to cut a release
since we haven't done one in a while, and to get these additive changes
out ahead of the number of breaking changes that'll probably come with
the next release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.3.0] - 2024-04-15
+
 ### Added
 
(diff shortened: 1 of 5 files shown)


bgentry (approved):
Need some docs for poll only?

brandur:
> Need some docs for poll only?

Yeah, trying to think about how to do this best. At the very least some of the PgBouncer docs probably need to be rewritten, but that said, I'd probably want to try and test things a little better this time.

I don't think getting this feature out there is super high priority, so I may keep it on the backburner for a bit. I was thinking about initially suggesting it to people who seem like they might be in a position where it'd be particularly useful for them.

PR #304, opened 2024-04-19
Change JobList ergnomics - add filtering by states, return cursor from JobList function

bgentry:
A tweaked version of https://github.com/riverqueue/river/pull/236, removing queries which cannot be efficient on a large `river_job` table due to the lack of indexes.

Users are of course free to add their own indexes and write their own query functions, but we're a bit uncomfortable offering official APIs which will scale poorly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,8 @@ Although it comes with a number of improvements, there's nothing particularly no
 - River uses a new job completer that batches up completion work so that large numbers of them can be performed more efficiently. In a purely synthetic (i.e. mostly unrealistic) benchmark, River's job throughput increases ~4.5x. [PR #258](https://github.com/riverqueue/river/pull/258).
 - Changed default client IDs to be a combination of hostname and the time which the client started. This can still be changed by specifying `Config.ID`. [PR #255](https://github.com/riverqueue/river/pull/255).
 - Notifier refactored for better robustness and testability. [PR #253](https://github.com/riverqueue/river/pull/253).
+- JobList/JobListTx now support querying Jobs by a list of Job Kinds and States (breaking change). Also allows for filtering by specific timestamp values. [PR #236](https://github.com/riverqueue/river/pull/236).
+
 
 ## [0.0.25] - 2024-03-01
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- JobList/JobListTx now support querying Jobs by a list of Job Kinds and States (breaking change). Also allows for filtering by specific timestamp values. Thank you Jos Kraaijeveld (@thatjos)! 🙏🏻 [PR #236](https://github.com/riverqueue/river/pull/236).
```
brandur, line 12: Been using this convention in the other PRs to indicate the breaking change:

```suggestion
- **Breaking change:** JobList/JobListTx now support querying Jobs by a list of Job Kinds and States. Also allows for filtering by specific timestamp values. Thank you Jos Kraaijeveld (@thatjos)! 🙏🏻 [PR #236](https://github.com/riverqueue/river/pull/236).
```

job_list_params.go
```diff
@@ -193,6 +190,23 @@ func (p *JobListParams) toDBParams() (*dblist.JobListParams, error) {
...
+			switch state {
+			case JobStateCancelled, JobStateCompleted, JobStateDiscarded:
+			default:
+				currentNonFinalizedStates = append(currentNonFinalizedStates, state)
+			}
+		}
+		// This indicates the user overrode the States list with only non-finalized
+		// states prior to then requesting FinalizedAt ordering.
+		if len(currentNonFinalizedStates) == 0 {
+			return nil, errors.New("cannot order by finalized_at with non-finalized state filters")
```
brandur, line 206: Could we put the non-finalized states into the error message with a `%+v` or something? Just helps make the error quicker to resolve.

brandur (approved):
Great! IMO let's get this in. As noted in chat I wouldn't mind trying to take a stab at a couple more small tweaks, but let's do that in a follow up.

thatjos:
Thanks so much! ❤️

PR #306, opened 2024-04-20
Document job states in `rivertype`

brandur:
I happened to notice the other day that some types that have survived to
this day without being document are those in `JobState*`. Since the
purpose of each of these states isn't necessarily obvious based on their
name alone, document them and provide some useful details that may save
users from having to jump to the website or read source code.

### DIFF
--- a/rivertype/river_type.go
+++ b/rivertype/river_type.go
@@ -111,17 +111,63 @@ type JobRow struct {
 	Tags []string
 }
 
-// JobState is the state of a job. Jobs start as `available` or `scheduled`, and
-// if all goes well eventually transition to `completed` as they're worked.
+// JobState is the state of a job. Jobs start their lifecycle as either
+// JobStateAvailable or JobStateScheduled, and if all goes well, transition to
+// JobStateCompleted after they're worked.
 type JobState string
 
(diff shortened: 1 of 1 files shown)


rivertype/river_type.go
```diff
@@ -111,17 +111,62 @@ type JobRow struct {
...
+	//
+	// Jobs that will be retried very soon in the future may be changed to
+	// JobStateAvailable immediately instead of JobStateRetryable so that they
+	// don't have to wait for the job scheduler to run.
 	JobStateRetryable JobState = "retryable"
-	JobStateRunning   JobState = "running"
+
+	// JobStateRunning are jobs which are actively running.
+	//
+	// If a worker dies unexpectedly while jobs were being worked, jobs will be
```
bgentry, line 159: Might be worth clarifying the behavior here as this is specifically true when the client doesn’t have an opportunity to log the failure (program crash, hardware failure, or job that never returns from work).

bgentry (approved):

rivertype/river_type.go
brandur: Sure, good idea. Updated!

PR #308, opened 2024-04-21
Prepare release `v0.4.0`

brandur:
Prepare `v0.4.0` for release. As usual, update all version in `go.mod`
files. Also, add some text in the changelog with a little more
information on breaking changes and the rationale behind putting many of
them in a single release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.4.0] - 2024-04-20
+
 ### Changed
+
+⚠️ Version 0.4.0 has a number of small breaking changes which we've decided to release all as part of a single version. More breaking changes in one release is inconvenient, but we've tried to coordinate them in hopes that any future breaking changes will be non-existent or very rare. All changes will get picked up by the Go compiler, and each one should be quite easy to fix. The changes don't apply to any of the most common core APIs, and likely many projects won't have to change any code.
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,12 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.4.0] - 2024-04-20
+
 ### Changed
 
+Version 0.4.0 comes with a number of small breaking changes which we've decided to release all as part of a single version. More breaking changes in one release is inconvenient, but we've tried to coordinate them in hopes that any future breaking changes will be non-existent or very rare. All changes will get picked up the Go compiler, and each one should be quite easy to fix. The changes don't apply to any of the most common core APIs, and likely many projects won't have to change any code.
```
bgentry, line 14: Should we put a ⚠️ emoji or anything with this? It’s a rare scenario and we want it to be extra visible.

bgentry, line 14:
```suggestion
Version 0.4.0 comes with a number of small breaking changes which we've decided to release all as part of a single version. More breaking changes in one release is inconvenient, but we've tried to coordinate them in hopes that any future breaking changes will be non-existent or very rare. All changes will get picked up by the Go compiler, and each one should be quite easy to fix. The changes don't apply to any of the most common core APIs, and likely many projects won't have to change any code.
```

bgentry (approved):

CHANGELOG.md
brandur: Done.

PR #311, opened 2024-04-21
Fix bug in `SharedTx` when `Query` returns on error condition

brandur:
Fix a hard-to-spot but long running bug in `SharedTx` in which when
`Query` was invoked, the standard success case unlocked the `SharedTx`
as expected, but if returning under error condition, the parent
`SharedTx` would not be unlocked.

This wasn't firing most of the time, but could occur in some stress
tests where maintenance services would occasionally successfully make it
into their work loop before their context was cancelled. The context
cancellation would error `Query`, not unlock `SharedTx`, and cause an
error if another service then later on entered its work loop and tried
to use `SharedTx`.


brandur:
@bgentry Was running your test suite from #284, and I'm pretty sure this is what was causing the problem.

bgentry (approved):
Ah yep makes sense! Tricky mutexes when the design doesn’t allow for `defer`.

brandur:
> Ah yep makes sense! Tricky mutexes when the design doesn’t allow for `defer`.

Yeah, definitely kind of a Go weakness that's hard to design around.

PR #315, opened 2024-04-22
Fix `river.JobStateAvailable` reference in `cmd/river/`

dhermes:
- Should be `rivertype`
- I discovered this while trying to pull in `0.4.0` release this morning (thanks @dependabot!); `cmd/river/` won't build in the release (didn't realize these aliases were still until #300)

### DIFF
--- a/client.go
+++ b/client.go
@@ -1465,7 +1465,7 @@ type JobListResult struct {
 // provided context is used for the underlying Postgres query and can be used to
 // cancel the operation or apply a timeout.
 //
-//	params := river.NewJobListParams().WithLimit(10).State(river.JobStateCompleted)
+//	params := river.NewJobListParams().WithLimit(10).State(rivertype.JobStateCompleted)
 //	jobRows, err := client.JobList(ctx, params)
 //	if err != nil {
 //		// handle error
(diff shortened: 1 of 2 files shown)


bgentry (approved):
Thank you for submitting a fix so quickly! 🙏🏻

dhermes:
@bgentry You've got to approve CI to run here

dhermes:
Oh shoot, I can't merge it either, sorry for the ping @bgentry!

Screenshot 2024-04-22 at 1 59 58 PM

brandur:
Thanks @dhermes.

Out of curiosity, you said you actually ran into an error because of this? I would not have expected this to happen because unlike other internal Go modules, `./cmd/river` is affixed to a particular version of other River packages (still 0.1.0 currently). That affixed version didn't change, so the removal of the `river.JobState*` aliases from the most recent version was safe and should have had no affect on it. I tried `go install ...` and it still seems to work happily.

Do you have something exotic going on in your build, or am I missing something?

dhermes:
@brandur We have `cmd/river/` in our `go.mod` + `vendor/`, and there isn't a mix-and-match of dependencies in a `go.mod` (so if `riverqueue/river` gets updated, it's global).

We vendor in `cmd/river/` so that in CI we don't end up having to do separate out-of-band Go build steps for our "check if `river` migrations are valid" (separate from building our source tree).

I suppose the build cost could be paid upfront and placed in a GitHub Action (e.g. via something like https://full-stack.blend.com/how-we-write-github-actions-in-go.html), just haven't realized there was a need until (maybe) right now.

brandur:
@dhermes Ah gotcha! Okay that makes sense. Thanks for explaining.

PR #324, opened 2024-04-30
add example for queue pause and resume

bgentry:
Adding this so I can link to it from docs.

### DIFF
--- a/example_queue_pause_test.go
+++ b/example_queue_pause_test.go
@@ -1,0 +1,127 @@
+package river_test
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"time"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+
+	"github.com/riverqueue/river"
+	"github.com/riverqueue/river/internal/riverinternaltest"
+	"github.com/riverqueue/river/internal/util/slogutil"
+	"github.com/riverqueue/river/riverdriver/riverpgxv5"
+)
+
+type ReportingArgs struct{}
+
+func (args ReportingArgs) Kind() string { return "Reporting" }
+
+type ReportingWorker struct {
+	river.WorkerDefaults[ReportingArgs]
+	jobWorkedCh chan<- string
+}
+
+func (w *ReportingWorker) Work(ctx context.Context, job *river.Job[ReportingArgs]) error {
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case w.jobWorkedCh <- job.Queue:
+		return nil
+	}
+}
+
+// Example_queuePause demonstrates how to pause queues to prevent them from
+// working new jobs, and later resume them.
+func Example_queuePause() {
+	ctx := context.Background()
+
+	dbPool, err := pgxpool.NewWithConfig(ctx, riverinternaltest.DatabaseConfig("river_testdb_example"))
+	if err != nil {
+		panic(err)
+	}
+	defer dbPool.Close()
+
+	// Required for the purpose of this test, but not necessary in real usage.
+	if err := riverinternaltest.TruncateRiverTables(ctx, dbPool); err != nil {
+		panic(err)
+	}
+
+	const (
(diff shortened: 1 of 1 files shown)


example_queue_pause_test.go
```diff
@@ -0,0 +1,137 @@
...
+	fmt.Printf("Job worked on %s queue\n", receivedQueue)
+	if receivedQueue != reliableQueue {
+		panic("expected reliable queue, got " + receivedQueue)
+	}
+
+	fmt.Printf("Inserting job into %s queue\n", unreliableQueue)
+	if _, err = riverClient.Insert(ctx, ReportingArgs{}, &river.InsertOpts{Queue: unreliableQueue}); err != nil {
+		panic(err)
+	}
+	waitExpectTimeout(jobWorkedCh)
```
brandur, line 96: Okay, so one thing with the example tests is that unlike the rest of the suite, they always run sequentially. So by adding this statement, we add exactly +1 second to the runtime of `go test .`, which is already extremely slow.

Instead of this, what do you think about something like:

* Pause the unreliable queue.
* Insert the unreliable job first, then the  the reliable job.
* Wait for the reliable job to be executed. Because it's got a larger ID than the unreliable job, we would have expected that the unreliable one would get worked first if it was going to be worked.
* Unpause the unreliable queue.
* Wait for one more completion.

We then depend partly on the test output to verify that everything happened in the order we expected.

bgentry: My concern is this flow makes us more susceptible to a race condition because there is no public API to determine when the producer has actually been paused (received the notification about the pause). My example here is actually vulnerable to this as well, though in practice it's less likely because of the latency between the pause and the subsequent insert + work.

Thoughts on how to avoid that race altogether?

Best idea I can come up with is to make `QueuePause()` ~synchronous within a given `Client`. It actually wouldn't be too tough to do that, but here are some considerations to highlight with doing that:

* We already have `producersByQueueName` and it's not even mutated after the `Client` starts as of today so it wouldn't require a mutex. Though we will probably eventually have more dynamic producers
* A small drawback would be that the client would still receive and handle the `NOTIFY` _after_ pausing ~synchronously. Even in poll-only mode, the poll routine keeps track of what it thinks the last pause state was and so it would also emit a duplicate event.
* These events would get sent on a producer's `queueControlCh` which is buffered, so they would send immediately. However the actual pausing will still take a small amount of time and would still result in some raciness in the operation. Making it _completely_ synchronous would require feedback or a blocking call to wait for the pause to be handled, so that's not great.

Alternatively we introduce an additional kind of subscription event that gets published when a queue has _actually_ paused or resumed.

brandur: Talked about this offline, but some event-based solutions are in progress now.

bgentry: Rebased on #327 and updated to use the flow you described above, made reliable by the new queue subscription events.

example_queue_pause_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// Resume the unreliable queue so it can work the job:
+	fmt.Printf("Resuming %s queue\n", unreliableQueue)
+	if err := riverClient.QueueResume(ctx, unreliableQueue, nil); err != nil {
+		panic(err)
+	}
+	receivedQueue = waitOrTimeout(jobWorkedCh)
+	fmt.Printf("Job worked on %s queue\n", receivedQueue)
+	if receivedQueue != unreliableQueue {
+		panic("expected unreliable queue, got " + receivedQueue)
```
brandur, line 111: What do you think about changing this to a `fmt.Printf` and having the output of the example assert its correctness instead?

When it comes to these examples, definitely the fewer LOCs helps to sharpen and clarify.

brandur (approved):

example_queue_pause_test.go
bgentry: Actually I can just straight remove these assertions, the printf above already covers it. Good call!

PR #325, opened 2024-05-02
Remove all in-code use of leadership "instance name" + constrain column to single value

brandur:
With the addition of #301, and more specifically the schema-based
namespacing that it brings in around listen/notify, we're fully moving
towards a world where the recommendation for running multiple Rivers in
a single database is very definitive: isolate them by schema.

The elector's had a long-standing parameter called "instance name"
that's stored into the `river_leader` table, and which was prospectively
going to be used for River namespacing, but something we never made use
of it. In a world of schema isolation, each schema will have its own
`river_leader` table, and no kind of additional namespacing is needed
within the table.

I was originally going to try and approach herein we drop `name` out of
`river_leader` completely, but looking more closely, it's the table's
primary key. We could add a new column that'd act as a primary key
instead (e.g. imagine a boolean primary key column with a check
constraint that makes sure it's always true), but nothing we'd add would
be that much better. Instead, I elected to give `name` a default value
of `default` (matching the previous default instance name), and add a
check constraint verifying that it's always `default`, making it
effectively a single row table. The nice part about this approach is
that we can put these changes into the V4 migration in #301, and we
won't require any additional changes in any future migration.

With `name` now constrained to a single value, we can simplify all the
`river_leader`-based queries by removing their name parameters, then
remove instance name completely from elector code and drivers, giving us
a thorough overall cleanup.

### DIFF
--- a/internal/leadership/elector.go
+++ b/internal/leadership/elector.go
@@ -21,17 +21,11 @@ const (
 	electIntervalDefault           = 5 * time.Second
 	electIntervalJitterDefault     = 1 * time.Second
 	electIntervalTTLPaddingDefault = 10 * time.Second
-
-	// TODO: for now we only support a single instance per database/schema.
-	// If we want to provide isolation within a single database/schema,
-	// we'll need to add a client config for this.
-	instanceNameDefault = "default"
 )
 
 type dbLeadershipNotification struct {
-	Name     string `json:"name"`
-	LeaderID string `json:"leader_id"`
 	Action   string `json:"action"`
+	LeaderID string `json:"leader_id"`
 }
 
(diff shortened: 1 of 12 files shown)


brandur:
@bgentry Thoughts on this approach? See PR description, but I realized that `name` was acting as the primary key for `river_leader`. We could change the key to something else, but it'd involve adding some new dummy column to act as one, so I opted instead to default `name `to `default`, and add a check constraint to verify that only the value `default` is allowed. An advantage of this compared to what you and I had discussed before is that `name` will stay around, meaning no future migration will be required.

bgentry (approved):
Works for me!

PR #326, opened 2024-05-02
Wrap queues returned from list in result type like we do for rows

brandur:
Something I noticed while rereading some of the queue code is that
although we added a `QueueListParams` to make queue listing symmetric
with job listing (and leaving space for future filters and a possible
cursor), we were still returning `[]*Queue`, which has the downside of
not leaving any space for a last cursor property should we add one in
the future.

Here, in the interest of keeping job/queue listing looking like each
other, have `QueueList` return a `QueueListResult`. So far this struct
contains only a `Queues` property, but a cursor could be added to it
should this functionality be implemented later.

~~Also, rename `rivertype.Queue` to `rivertype.QueueRow`. This makes it a
little bit uglier, but I figure it'd not the worst idea to try and keep
it matched to the convention seen in `rivertype.JobRow`.~~

### DIFF
--- a/client.go
+++ b/client.go
@@ -1611,7 +1611,10 @@ func validateQueueName(queueName string) error {
 // JobListResult is the result of a job list operation. It contains a list of
 // jobs and a cursor for fetching the next page of results.
 type JobListResult struct {
-	Jobs       []*rivertype.JobRow
+	// Jobs is a slice of job returned as part of the list operation.
+	Jobs []*rivertype.JobRow
+
+	// LastCursor is a cursor that can be used to list the next page of jobs.
 	LastCursor *JobListCursor
 }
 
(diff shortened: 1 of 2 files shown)


rivertype/river_type.go
```diff
@@ -214,9 +214,9 @@ type AttemptError struct {
 // subsequently remove the periodic job with `Remove()`.
 type PeriodicJobHandle int
 
-// Queue is a configuration for a queue that is currently (or recently was) in
+// QueueRow is a configuration for a queue that is currently (or recently was) in
 // use by a client.
-type Queue struct {
+type QueueRow struct {
```
bgentry, line 219: I'm hesitant to do this. We had to call the job one `JobRow` because we wanted the generic typed variant to be `Job[T]`. However we have no such issue with the queue type and I doubt we will, so it seems unnecessary to clutter this one up with a `*Row` suffix.

bgentry (commented):
Hmm, don't think I'm a fan of this one but let's chat if you feel strongly :v:

rivertype/river_type.go
brandur: Fair enough — maybe our convention can just be that every model gets its normal noun as a name, and `JobRow` is the single exceptional case. I reverted back to `Queue` and updated the commit message.

bgentry (approved):

PR #327, opened 2024-05-02
emit subscription events on pause + resume

bgentry:
Following up on https://github.com/riverqueue/river/pull/324#discussion_r1586541810, this adds `Client` subscription events for queue pause and resume. This allows us to make tests and examples free of race conditions, even without access to internal test signals.

Unlike for job events coming from the completer, here it is not straightforward to emit the entire representation of the `Queue` record. That's because we don't always have a fresh version of that record available when handling a pause or resume event. We considered emitting that as part of the pause or resume event, but that doesn't work when operating on all queues via `*`. We also considered doing this at the time the event is emitted by the producer, but that is pretty messy because we'd have to reload the row before emitting the event, leaving an odd gap and race condition before the event emission.

Instead, this approach uses a mostly-unpopulated `Queue` struct with only the `Name` and all other fields left with their default value. It's not ideal but it seems like the best pragmatic compromise at this time.

### DIFF
--- a/client.go
+++ b/client.go
@@ -499,6 +499,9 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 				MaxWorkers:        queueConfig.MaxWorkers,
 				Notifier:          client.notifier,
 				Queue:             queue,
+				QueueEventCallback: func(event *Event) {
+					client.distributeQueueEvent(event)
+				},
 				RetryPolicy:       config.RetryPolicy,
 				SchedulerInterval: config.schedulerInterval,
 				StatusFunc:        client.monitor.SetProducerStatus,
@@ -909,8 +912,11 @@ func (c *Client[TTx]) SubscribeConfig(config *SubscribeConfig) (<-chan *Event, f
 	return subChan, cancel
 }
 
-// Distribute a single job into any listening subscriber channels.
-func (c *Client[TTx]) distributeJob(job *rivertype.JobRow, stats *JobStatistics) {
+// Distribute a single event into any listening subscriber channels.
+//
+// Job events should specify the job and stats, while queue events should only specify
+// the queue.
+func (c *Client[TTx]) distributeJobEvent(job *rivertype.JobRow, stats *JobStatistics) {
 	c.subscriptionsMu.Lock()
 	defer c.subscriptionsMu.Unlock()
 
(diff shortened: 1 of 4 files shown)


client.go
```diff
@@ -920,20 +926,30 @@ func (c *Client[TTx]) distributeJob(job *rivertype.JobRow, stats *JobStatistics)
...
+		default:
+			// linter exhaustive rule prevents this from being reached
+			panic("unreachable state to distribute, river bug")
+		}
+	} else {
+		switch queue.PausedAt {
+		case nil:
+			event = &Event{Kind: EventKindQueueResumed, Queue: queue}
+		default:
+			event = &Event{Kind: EventKindQueuePaused, Queue: queue}
```
brandur, line 951: What do you think about just having the producer generate the event rather than having a separate field that's then reassembled into an event post-hoc? See:

https://github.com/riverqueue/river/pull/328/files

Feels a little bit cleaner, and it may help for assembling a full queue object.

client_test.go
```diff
@@ -511,74 +509,73 @@ func Test_Client(t *testing.T) {
...
-
-		insertRes1, err := client.Insert(ctx, &JobArgs{}, nil)
+		insertRes1, err := client.Insert(ctx, &noOpArgs{}, nil)
 		require.NoError(t, err)
 
-		startedJobID := riverinternaltest.WaitOrTimeout(t, jobStartedChan)
-		require.Equal(t, insertRes1.Job.ID, startedJobID)
+		event := riverinternaltest.WaitOrTimeout(t, subscribeChan)
+		require.Equal(t, EventKindJobCompleted, event.Kind)
+		require.Equal(t, insertRes1.Job.ID, event.Job.ID)
```
brandur, line 520: Just for the sake of keeping these tests shorter, what do you think of only subscribing to queue-based events here, then we can skip the checks on job events?

client_test.go
```diff
@@ -511,74 +509,73 @@ func Test_Client(t *testing.T) {
...
 
 		// Pause only the default queue:
 		require.NoError(t, client.QueuePause(ctx, QueueDefault, nil))
-		client.producersByQueueName[QueueDefault].testSignals.Paused.WaitOrTimeout()
+		event = riverinternaltest.WaitOrTimeout(t, subscribeChan)
+		require.Equal(t, EventKindQueuePaused, event.Kind)
+		require.Equal(t, QueueDefault, event.Queue.Name)
```
brandur, line 526: This might depend on whether we get the full queue object in the end, but IMO, feels cleaner to have a single assertion on the full event object. More readable, and produces a better error in case of failure too:

```
		pauseEvent := riverinternaltest.WaitOrTimeout(t, subscribeChan)
		require.Equal(t, &Event{Kind: EventKindQueuePaused, Queue: QueueDefault}, pauseEvent)
```

bgentry: I agree but I think this breaks apart when using the full queue row unfortunately due to the timestamps. Please take a look at the new version when posted though because you may still have ideas to improve.

client_test.go
bgentry: With the rebased version here, I don't think subscribing to only queue events would give us coverage around when the jobs are handled (i.e. not while paused).

client_test.go
bgentry: Ended up making this change given the updated approach, thanks!

client_test.go
brandur: Cool. Yeah I guess it reduces some of the other boiler plate, so seems good.

brandur (approved):
Great! :shipit:

PR #328, opened 2024-05-02
Add subscription events that are emitted when a queue is paused or resumed

brandur:
Here, largely as a potential way to resolve #324, add new events
`EventKindQueuePaused` and `EventKindQueueResumed` which are emitted
whenever a queue is paused or resumed. Tests will be able to use these
to wait on the completion of these actions even if calls to `QueuePause`
or `QueueResume` are asynchronous (as they are currently), or if the
pause/resume was initiated by a different client. I expect them to also
be useful for general telemetry/logging purposes.

Recall that when we first designed the subscription API, we explicitly
_did not_ add a way to subscribe to all possible event types. This means
that although the new queue events fill `Event.Queue` instead of
`Event.Job` and `Event.JobStats`, the API is still forward compatible
because nobody could be listening to these yet.

### DIFF
--- a/client.go
+++ b/client.go
@@ -488,21 +488,33 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		})
 		client.services = append(client.services, client.elector)
 
+		distributedQueueEvents := func(event *Event) {
+			for _, sub := range client.subscriptions {
+				if sub.ListensFor(event.Kind) {
+					select {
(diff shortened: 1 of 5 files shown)


brandur:
@bgentry Wrote this on the airplane before seeing you'd done it already. Opening PR for posterity.

bgentry:
Should be superseded by the updated #327.

PR #337, opened 2024-05-04
Drop `lib/pq` use in job listing

brandur:
I missed this before, but while working on something else I noticed
randomly that apparently we use `lib/pq` in the job list module to
encode an array of job states.

I think it was an oversight to have brought this in as a hard
dependency, so here remove the use of `pg.Array` in favor of just using
a normal Go slice instead.


bgentry (approved):
How the heck did this sneak in?? My bad I think …

brandur:
Hah, yeah I was surprised to see it too!

Thanks.

Issue #340, opened 2024-05-06
Periodic river job `scheduled_at` does not match timestamp returned from schedule func


dhermes:
For example, using a `cron.ParseStandard("CRON_TZ=UTC 39 * * * *")` (this is `github.com/robfig/cron/v3`, see code below) my schedule func returned `2024-05-06T18:39:00.000Z` but the `scheduled_at` on the job is `2024-05-06T18:39:00.000609Z`.

## Minimal viable repro

<details>
  <summary>Example Go code</summary>

```go
package main

import (
	"context"
	"fmt"
	"os"
	"time"

	"github.com/jackc/pgx/v5/pgxpool"
	"github.com/riverqueue/river"
	"github.com/riverqueue/river/riverdriver/riverpgxv5"
	cron "github.com/robfig/cron/v3"
)

// HelloJobArgs is a job args struct for a `HelloJob`.
type HelloJobArgs struct{}

// Kind satisfies `river.JobArgs`.
func (HelloJobArgs) Kind() string {
	return "hello"
}

// HelloJob is a worker for printing "Hello, world!". It satisfies
// `river.Worker[HelloJobArgs]`.
type HelloJob struct {
	river.WorkerDefaults[HelloJobArgs]
}

// Work works a `HelloJob`.
func (*HelloJob) Work(_ context.Context, job *river.Job[HelloJobArgs]) error {
	fmt.Printf("Hello, world! Job ID: %d (scheduled at: %s)\n", job.ID, job.ScheduledAt)
	return nil
}

// NewPeriodicJob returns a `river.PeriodicJob` for running a `HelloJob` on
// a `crontab` schedule `s`.
func NewPeriodicJob(s river.PeriodicSchedule) *river.PeriodicJob {
	return river.NewPeriodicJob(
		s,
		func() (river.JobArgs, *river.InsertOpts) {
			now := time.Now().UTC()
			fmt.Printf("Producing job args: %s\n", now)
			return HelloJobArgs{}, nil
		},
		&river.PeriodicJobOpts{RunOnStart: false},
	)
}

type wrappedCron struct {
	Cron cron.Schedule
}

func (w *wrappedCron) Next(current time.Time) time.Time {
	next := w.Cron.Next(current)
	fmt.Printf("Next(%s) -> %s\n", current, next)
	return next
}

func getPool(ctx context.Context) (*pgxpool.Pool, error) {
	config, err := pgxpool.ParseConfig("postgres://...")
	if err != nil {
		return nil, err
[...]

PR #341, opened 2024-05-07
Periodic job enqueuer: Set `scheduled_at` according to expected next run time

brandur:
Currently, a periodic job that's inserted gets a value in `scheduled_at`
that defaults to the database's `now()`. Here, we alter this with a
minor nicety requested in #340 wherein we set `scheduled_at` to the
precise time we expected a periodic job to be scheduled. This value is
will be very close to `now()`, but likely a little sooner because
there's some lag in the insert loop between when job insert params are
generated and when the insert happens.

Jobs are inserted with state `available` (rather than `scheduled`), so
this should have very little functional effect. In a busy system it may
cause the periodic jobs to be favored slightly because when locking
available jobs, `scheduled_at` is one of the fields we order on for
priority.


bgentry (approved):

dhermes:
Awesome stuff, thank you!

brandur:
Thanks guys. I unset this as resolving #340 since there's still a little discussion going over there.

PR #344, opened 2024-05-08
Prepare version `v0.6.0`

brandur:
This release isn't strictly required, but it's been about a week since
the last one, and we've accumulated a few changes since (e.g. test
helpers like `RequireNotInserted`, dropping `lib/pq` dependency, and
more accurate `scheduled_at` for periodic job inserts). Let's cut a
release to keep our cadence up in that department.


bgentry (approved):

brandur:
Thanks!

PR #348, opened 2024-05-09
also log job kind 

elee1766:
it would be rather useful to also log the kind here, so that one doesn't need to look in the db to figure out which job is failing.


brandur (approved):
Works for me!

Issue #347, opened 2024-05-09
can't configure RescueStuckJobsAfter for an individual job


elee1766:
we have some jobs that take 1-2 minutes, and other jobs that takes at most 5 seconds, and so we set worker timeouts accordingly

however, we can't configure RescueStuckJobsAfter for an individual job. this means that when a worker running the 5 seconds task dies, we need to wait for the global RescueStuckJobsAfter, instead of what should be 5-10 seconds. this means that updates that should be happening around once a second pause for a few minutes instead of a few seconds. 

we schedule with a unique constraint because this task runs on every event, but sometimes events can be up to 10 per second, and we are okay with not scheduling the task if it is already running since it is an aggregate and will just be picked up on the next unique task. if we schedule too many jobs, we will not only schedule more than we can process, but be doing extra work, and need to worry about race conditions where tasks that started earlier finish later. 

one possible solution is to read the `jobs` table during InsertTx, and check using `scheduled_time` if the job is stuck ourselves, and manually fix it, but i'm not sure if that's allowed. for now i am checking the scheduled_time in a loop in my go application to manually rescue the task as this one specific task is critical to the movement of the rest of the applications. 

a `deadline` feature in `InsertOpts` i think would make me really happy, as it would also resolve other issues im having to work around. workers should discard jobs that they pick up that are past their deadline, and a maintenance job which "discards" tasks that are after their deadline could run on a regular interval. 

the problem is adjacent to something that @bgentry mentioned in a previous reply to me in https://github.com/riverqueue/river/issues/336#issuecomment-2093895895 . This is actually a different summary job in a different app, we deal with a lot of data that needs to be updated in this manner.

> It also sounds like you’re looking to ensure that only one of these summary jobs is running at a given moment. This is also something there’s no official support for, though you can hack it in with your own locking mechanism. However it’s something we have thought about and hope to implement. 

we have had success using some hacky locking mechanisms for our other tasks, however in this case we are trying using the job=running as our locking mechanism, and the issue is that it takes too long for a dying worker to "abandon" their lock.  

for reference, the system we are migrating from used a user defined hash as an optional uniqueness key with support for a job timeout and deadline on the job level. the key would be locked when scheduled, and unlocked at deadline or job completion.  we could bring that idea back and use it along with river.  

this calls back to what i was told to do here: https://github.com/riverqueue/river/discussions/346

and also very much relates to this problem: https://github.com/riverqueue/river/issues/165

[...]

PR #350, opened 2024-05-11
Consider per-worker timeout overrides when rescuing jobs

brandur:
This one came up when I was thinking about the job specific rescue
threshold floated in [1].

I was going to suggest the possible workaround of setting an aggressive
rescue threshold combined with a low job timeout globally, and then
override the timeout on any specific job workers that needed to run
longer than the new low global job timeout. But then I realized this
wouldn't work because the job rescuer doesn't account for job-specific
timeouts -- it just rescues or discards everything it finds beyond the
run's rescue threshold.

Here, add new logic to address that problem. Luckily we were already
pulling worker information to procure what might be a possible custom
retry schedule, so we just have to piggyback onto that to also examine
a possible custom work timeout.

[1] https://github.com/riverqueue/river/issues/347

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- River now considers per-worker timeout overrides when rescuing jobs so that jobs with a long custom timeout won't be rescued prematurely. [PR #350](https://github.com/riverqueue/river/pull/350).
+
 ## [0.6.0] - 2024-05-08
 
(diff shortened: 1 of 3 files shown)


bgentry (approved):
Good idea and LGTM. Think this can also use a doc change on https://riverqueue.com/docs/maintenance-services#rescuer ?

Noticing two other things about that doc section:

1. It doesn't mention hardware faults or software crashes as a potential cause of a "stuck" job, which we should probably add.
2. "transaction job completion" should probably be "transactional job completion"

brandur:
Cool, done in https://github.com/riverqueue/river-homepage/pull/93.

Issue #302, opened 2024-04-15
Is there a way to use transactions without pgx?


mrkaspa:
I have a project with bun sql and I would like to insert a job as part of a transaction, but I can only use pgx.Tx transactions. Also I can use sql.Tx but the driver does not support this

brandur:
Not currently — I have a branch where this _almost_ works, but it's going to take a little more work still. I think I'll get this sooner than later.

mrkaspa:
@brandur great to hear that, hope to use that feature soon

hypnoglow:
We also use [bun](https://bun.uptrace.dev/) and would love to see support for `bun.Tx` and/or `sql.Tx`. Otherwise it seems impossible to integrate with existing transactions.

brandur:
Okay, good to know. Going to try and get to this pretty soon.

brandur:
Hey guys, could you tell me what underlying driver you're using to power `database/sql` in general? Do you use `lib/pq` or is `database/sql` + Pgx, or something else like Bun's Postgres driver?

hypnoglow:
Hey @brandur 

Currently, we consider only `pgx` to be a viable driver. We switch old apps from `lib/pq`. And `bun` native driver has some problems that we don't have using `pgx`.

brandur:
Excellent, that's helpful. Thanks Igor.

dhermes:
Chiming in here! We are currently using `database/sql` with `github.com/jackc/pgx/v5/stdlib` as the actual driver.

brandur:
Opened #351. We'll need some additional docs, but it make use of River with Bun possible.

ewhauser:
We use `lib/pq` and would love to see River working with it. Happy to do some legwork

ewhauser:
Actually, ignore this request. Sounds like after some research it's better that we just migrate over to `pgx` anyways.

brandur:
This one's pretty close in that there's a working driver in #351, but yes, you should definitely switch to pgx anyway.

bgentry:
If anybody out there is interested in giving this a try, `database/sql` functionality was merged in #351. We still haven't cut an official release with it (should come in v0.10.0) nor have we finished documentation on it, but the actual support has been merged.

hypnoglow:
Hey @bgentry 

I found that docs about drivers are now outdated, without database/sql not mentioned. https://riverqueue.com/docs/database-drivers

Issue #58, opened 2023-11-21
Gorm integration


mgdigital:
Hi, great project! 

I'm looking at using this to replace [Asynq](https://github.com/hibiken/asynq) in my application and remove the Redis dependency. I'm wondering how I can integrate this with [Gorm](https://gorm.io/) and insert jobs as part of a Gorm transaction - they are both using the PGX driver so it should be possible but Gorm is wrapping transactions in layers of abstraction. A dirty workaround might be to get Gorm to execute the insert query here: https://github.com/riverqueue/river/blob/master/internal/dbsqlc/river_job.sql.go#L450 but that doesn't seem ideal. 

Gorm is a popular ORM so I expect easy integration would improve adoption of River...

bgentry:
Hi @mgdigital, there are a couple of challenges to work through here:

1. Although River has a concept of "drivers", this was a last-minute addition and you can probably see signs of this throughout the interfaces. There are a lot of places with hard dependencies on `pgx.Tx` for example which might be hard to cleanly abstract with database-agnostic interfaces. Not necessarily impossible but it will take some careful thought and a lot of work.
2. There's a distinction between Clients where you're inserting jobs, and those where you're working. The worker Clients are doing a lot more database-specific stuff beyond just simple queries including LISTEN/NOTIFY which is tightly tied to `pgx.Conn` as of today and would be pretty difficult to abstract away.

In short this isn't something that's likely to happen soon, though I do hope we can get there eventually. It's tough that the Go ecosystem has such a fragmented database driver landscape 😕

brandur:
Opened #351. We'll need some additional docs, but it make use of River with GORM possible.

arp242:
> The worker Clients are doing a lot more database-specific stuff beyond just simple queries including LISTEN/NOTIFY which is tightly tied to pgx.Conn as of today and would be pretty difficult to abstract away.

If you use `sql.Open("pgx", connect)`, then `github.com/jackc/pgx/v5/stdlib` basically just wraps `pgx.Conn` a bit to be compatible with stdlib.

The difficult scenario now is situations where you do use pgx, but not "directly" but via gorm, or some other abstraction.

`riverpgxv5.New(nil)` doesn't really work because this expects `pgx.Tx`, but I only have stdlib `*sql.Tx` because of this abstraction.

The bit that's still missing is:

    river.NewClient(riverpgxv5.NewStdlib(nil), ...)
    // Returns: *river.Client[*sql.Tx]

Or something...

[...]

PR #351, opened 2024-05-12
Fully functional `database/sql` driver

brandur:
Here, implement the rest of driver functionality on `riverdatabasesql`,
the existing driver for Go's built-in `database/sql` package. Previously
it only supported a minimal interface allowing it to run migrations, but
nothing more sophisticated like inserting jobs.

The benefit of a fully functional driver is that it will allow River to
be integrated with with other Go database packages that aren't built
around Pgx like Bun (requested in #302) and GORM (requested in #58).
I'll need to write up some documentation, but this change should make
both of those integrations possible immediately.

It also lays the groundwork for future non-Postgres drivers. It's going
to be a little more still, but I want to take a stab at SQLite, and this
change will get us a lot of the way there.

There's no way with `database/sql` to support listen/notify, so here we
introduce the idea of a poll only driver. River's client checks whether
a driver can support listen/notify on initialization, and if not, it
enters poll only mode the same way as if configured with `PollOnly`.

An intuitive idiosyncrasy of this set up is that even when using the
`database/sql` driver bundled here, regardless of whether they're
working with Bun, GORM, or whatever,  users will generally still be
using Pgx under the hood since it's the only maintained and fully
functional Postgres driver in the Go ecosystem. With that said, the
driver still has to bundle in `lib/pq` for various constructs like
`pq.Array` because we're using sqlc, and sqlc's `database/sql` driver
always uses `lib/pq`. I tried to find a way around this, but came out
fairly convinced that there is none. To rid ourselves of `lib/pq`
completely we'd need sqlc to ship an alternative Pgx driver that used
Pgx internally, but exposed a `database/sql` interface using `*sql.Tx`
instead of `pgx.Tx`.


client.go
```diff
@@ -1264,6 +1268,18 @@ func insertParamsFromArgsAndOptions(args JobArgs, insertOpts *InsertOpts) (*rive
...
+		for _, tag := range tags {
+			if len(tag) > 255 {
+				return nil, nil, errors.New("tags should be a maximum of 255 characters long")
+			}
+			// Restricted commas because we need those for batch inserts with
+			// the riverdatabasesql driver. We may want to restrict other
+			// special characters as well.
+			if strings.Contains(tag, ",") {
+				return nil, nil, errors.New("tags should not contain commas")
+			}
```
bgentry, line 1281: maybe tags should be restricted to `/a-z_\-/i`? _possibly_ include single spaces in that but I don't think it would harm any use cases to leave that out.

riverdriver/riverdatabasesql/internal/dbsqlc/sqlc.yaml
```diff
@@ -22,10 +24,19 @@ sql:
...
           ttl: "TTL"
 
         overrides:
+          # `database/sql` really does not play nicely with json/jsonb. If it's
+          # left as `[]byte` or `json.RawMessage`, `database/sql` will try to
+          # encode it as binary (with a \x) which Postgres won't accept as
+          # json/jsonb at all. Using a custom struct crashed and burned, even
+          # with a custom scanner implementation. This is the only way I could
+          # get it to work: strings are compatible with our use of bytes slices,
+          # but Postgrs will also accept them as json/jsonb.
```
bgentry, line 36:
```suggestion
          # but Postgres will also accept them as json/jsonb.
```

bgentry:
> There's no way with `database/sql` to support listen/notify, so here we
> introduce the idea of a poll only driver. River's client checks whether
> a driver can support listen/notify on initialization, and if not, it
> enters poll only mode the same way as if configured with `PollOnly`.

I think we've discussed this possibility before, but wanted to raise it again as it's relevant here. If users are actually using `pgx` under the hood of `database/sql` or Bun, it shouldn't be too difficult for us to use a raw pgx connection for the listener—provided the driver's constructor gives us a way to create one.

We can't rely on a single conn never failing, so we'd need to at least be able to recreate a provided conn from its [`ConnConfig`](https://pkg.go.dev/github.com/jackc/pgx#ConnConfig), OR we could accept a listener-specific `pgxpool.Pool` as part of the constructor so that we can steal conns from it as needed. We also talked about having different constructors within a single driver that give access to different functionality, so even if we don't add this now it doesn't preclude us doing so in the future. I think a change/addition like this would bring this `database/sql` driver pretty much to feature parity though!

> With that said, the
> driver still has to bundle in `lib/pq` for various constructs like
> `pq.Array` because we're using sqlc, and sqlc's `database/sql` driver
> always uses `lib/pq`. I tried to find a way around this, but came out
> fairly convinced that there is none. To rid ourselves of `lib/pq`
> completely we'd need sqlc to ship an alternative Pgx driver that used
> Pgx internally, but exposed a `database/sql` interface using `*sql.Tx`
> instead of `pgx.Tx`.

Dang, I wonder how tough of a lift this would be? Especially since iirc pgx has complete support for most of its types within database/sql. I tried searching for issues related to this and only found https://github.com/sqlc-dev/sqlc/issues/2960 that was relevant.

Regarding the rest of this, I'm not gonna get through a full review this evening but should be able to get through it in the morning.

riverdriver/riverdatabasesql/internal/dbsqlc/sqlc.yaml
brandur: Fixed, thanks.

client.go
brandur: Alright, I added this variant instead:

``` go
var tagRE = regexp.MustCompile(`\A[\w][\w\-]+[\w]\z`)
```

Test matrix:

``` go
func TestTagRE(t *testing.T) {
	t.Parallel()

	require.Regexp(t, tagRE, "aaa")
	require.Regexp(t, tagRE, "_aaa")
	require.Regexp(t, tagRE, "aaa_")
	require.Regexp(t, tagRE, "777")
	require.Regexp(t, tagRE, "my-tag")
	require.Regexp(t, tagRE, "my_tag")
	require.Regexp(t, tagRE, "my_longer_tag")
	require.Regexp(t, tagRE, "My_Capitalized_Tag")
	require.Regexp(t, tagRE, "ALL_CAPS")
	require.Regexp(t, tagRE, "1_2_3")

	require.NotRegexp(t, tagRE, "a")
	require.NotRegexp(t, tagRE, "aa")
	require.NotRegexp(t, tagRE, "-aaa")
	require.NotRegexp(t, tagRE, "aaa-")
	require.NotRegexp(t, tagRE, "special@characters$banned")
	require.NotRegexp(t, tagRE, "commas,never,allowed")
}
```

brandur:
> I think we've discussed this possibility before, but wanted to raise it again as it's relevant here. If users are actually using `pgx` under the hood of `database/sql` or Bun, it shouldn't be too difficult for us to use a raw pgx connection for the listener—provided the driver's constructor gives us a way to create one.

I don't love this — IMO, this sort of internal magic would be quite surprising and there'd be no intuitive way to predict that it might happen. It also puts a Pgx dependency back into `riverdatabasesql` — not the worst thing, but not great.

There's a much better way to do this which is that if you're using Pgx but still need `database/sql` in some places (say you're using Bun): (1) initialize one work client with `riverpgxv5`, and (2) initialize a second insert client that you use with Bun. Check out the latest blog post teed up in for the home page — there's a code sample in there of what it'd look like.

> Dang, I wonder how tough of a lift this would be? Especially since iirc pgx has complete support for most of its types within database/sql. I tried searching for issues related to this and only found https://github.com/sqlc-dev/sqlc/issues/2960 that was relevant.

I don't think it'd be _that_ hard as far as adding drivers to Sqlc goes (overall, it doesn't seem to use _that much_ lib/pq, just a few functions here and there), but any new Sqlc driver looks like kind of a lot of work :/ This is one that I'm kind of hoping that if we wait long enough someone will eventually do it, but certainly more than I want to commit to right now.

> Regarding the rest of this, I'm not gonna get through a full review this evening but should be able to get through it in the morning.

Awesome, thx.

client.go
```diff
@@ -1264,6 +1268,15 @@ func insertParamsFromArgsAndOptions(args JobArgs, insertOpts *InsertOpts) (*rive
...
 		tags = []string{}
+	} else {
+		for _, tag := range tags {
+			if len(tag) > 255 {
+				return nil, nil, errors.New("tags should be a maximum of 255 characters long")
+			}
+			if !tagRE.MatchString(tag) {
+				return nil, nil, errors.New("tags should match regex " + tagRE.String())
+			}
+		}
```
bgentry, line 1279: This probably requires its own changelog entry.

riverdriver/riverdatabasesql/river_database_sql.go
bgentry, line 1: Definitely a lot of copy-pasta type boilerplate in here, huh? Not sure there's much to be done about it though 😕

riverdriver/riverdatabasesql/river_database_sql.go
```diff
@@ -70,107 +73,392 @@ func (e *Executor) Exec(ctx context.Context, sql string) (struct{}, error) {
...
+	// for named parameters, but neither Pgx or lib/pq implement it, so just use
+	// dumb string replacement given we're only injecting a very basic value
+	// anyway.
+	for name, value := range namedArgs {
+		newQuery := strings.Replace(query, "@"+name, fmt.Sprintf("%v", value), 1)
+		if newQuery == query {
+			return nil, fmt.Errorf("named query parameter @%s not found in query", name)
+		}
+		query = newQuery
+	}
```
bgentry, line 269: 😵‍💫

bgentry (approved):
Ok, awesome work! Also after proofing out #352 I'm feeling a lot better about shipping this PR, although I would still like to talk about whether we can ship that additional one before we blog about and promote this new `database/sql` support—I want as many people as possible to get the full River experience ⚡️

riverdriver/riverdatabasesql/river_database_sql.go
brandur: Yeah it's a little unfortunate :( I think we could improve some of it by having some shared helpers somewhere, but it's a little tricky since they're in different packages. Also, a lot of the code looks like perfect copy/pasta, but believe it or not isn't quite — there's lots of tiny differences that are required to make things work with `database/sql`.

client.go
brandur: Yep, good call. Added.

brandur:
@bgentry I need to base another PR on this one so gonna bring it for now, although won't cut a release right away.

PR #352, opened 2024-05-14
Allow using database/sql driver with functioning pgx listener

bgentry:
Building on #351, this PR allows users to take advantage of the new `riverdatabasesql` driver while also keeping full support for `LISTEN` and avoiding poll-only mode.

This is accomplished through a `riverdatabasesql.NewWithListener()` constructor that allows the `database/sql` driver to be used with a functioning listener implementation. Also adds a `riverpgxv5.NewListener()` constructor to allow creating a listener with a raw pgx pool.

These can be combined to allow full listener support as long as the underlying database driver supports it, even when it's used within an abstraction like `database/sql` or Bun.

Needs tests if we want to proceed with this. I'm pleased with how little code it took to achieve though!

You could imagine this being used in the future with a non-Postgres listener implementation such as Redis, though for that to work we may need to loosen notifications to be emitted outside of a transaction (and may not be able to emit them automatically at all for the `*Tx` insert variants bc we cannot tell when a transaction completes).


leaanthony-sc:
Any news on this? It would be incredible to get LISTEN/NOTIFY using the standard driver!

dragondgold:
Indeed, this would be nice to have. Is this being worked on?

PR #363, opened 2024-05-19
Exit with status 1 on Cobra problems like a missing required flag

### LINKED ISSUE #362: cmd/river: Exit code 0 on flag error
When I run:
```
$ go run github.com/riverqueue/river/cmd/river@latest migrate-up
```
I get:
```
Error: required flag(s) "database-url" not set
Usage:
[...]
```
But the exit code is 0, which means success:
```
$ echo $?
0
```
The command should exit with a non-zero code on error.

brandur:
Here, exit with status 1 from the River CLI when a problem intrinsic to
Cobra occurred like the user entered an unknown command, or didn't pass
required flag.

Cobra is a little odd in its behavior in that it already prints such a
problem to the terminal, but also returns an error with the same
information. This change looks for the presence of an error and exits
with status 1 if one occurred, but doesn't double print any additional
information about it.

Fixes #362.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -168,6 +168,13 @@ jobs:
           fi
         working-directory: ./cmd/river
 
+      - name: river unknown command (expect failure)
+        run: |
+          if ./river not-a-command; then
+            echo "expected non-zero exit code" && exit 1
+          fi
+        working-directory: ./cmd/river
+
   golangci:
     name: lint
     runs-on: ubuntu-latest
--- a/cmd/river/main.go
+++ b/cmd/river/main.go
@@ -234,10 +234,12 @@ migrations that need to be run, but without running them.
 		rootCmd.AddCommand(cmd)
 	}
 
-	// Cobra will already print an error on an uknown command, and there aren't
-	// really any other important top-level error cases to worry about as far as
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -168,6 +168,13 @@ jobs:
           fi
         working-directory: ./cmd/river
 
+      - name: river unknown command (expect failure)
+        run: |
+          if ./river not-a-command; then
+            echo "expected non-zero exit code" && exit 1
+          fi
+        working-directory: ./cmd/river
```
brandur, line 176: Tested locally as well:

``` sh
$ go run . migrate-up
Error: required flag(s) "database-url" not set
Usage:
  river migrate-up [flags]

Flags:
      --database-url postgres://...   URL of the database to migrate (should look like postgres://...
      --dry-run                       print information on migrations, but don't apply them
  -h, --help                          help for migrate-up
      --max-steps int                 maximum number of steps to migrate
      --show-sql                      show SQL of each migration
      --target-version int            target version to migrate to (final state includes this version, but none after it)

Global Flags:
      --debug     output maximum logging verbosity (debug level)
  -v, --verbose   output additional logging verbosity (info level)

exit status 1

$ echo $?
1
```

brandur:
Thanks!

PR #375, opened 2024-05-29
Improve documentation for insert-only clients

### LINKED ISSUE #371: Docs should make insert-only Client mode clearer
When creating a new River client, the client forces you to supply queues and workers. Specifying this configuration does things like 1) run leader election 2) start the reindexer 3) start polling for jobs 4) etc. There are a couple reason the client should support insert only mode:

1) There are many instances where a service/job will just be inserting jobs into the queue vs. also running workers
2) If you are testing code that is only doing inserts

brandur:
Try to address #371 in that it's currently too hard to discover that an
insert-only client is possible, and how to initialize one. There is some
documentation on the `Queues` variable, but it's not that easy to find.

Here, add docs in a few other places under the principle that some docs
repetition isn't the worse thing in that it helps make these concepts
more obvious to find. We put docs in the README and package docs (which
mirror each other), along with some on the `river.Config` struct.

Fixes #371.

### DIFF
--- a/client.go
+++ b/client.go
@@ -48,6 +48,12 @@ const (
 )
 
 // Config is the configuration for a Client.
+//
+// Both Queues and Workers are required for a client to work jobs, but an
+// insert-only client can be initialized by omitting Queues, and not calling
+// Start for the client. Workers can also be omitted, but it's better to include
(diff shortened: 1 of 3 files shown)


bgentry (approved):
Thanks!

PR #379, opened 2024-05-31
extract Client subscriptions into service

bgentry:
This change extracts the Client subscriptions logic into a separate `startstop.Service` which can be started and stopped along with the other services. The important change that enables this is switching from a _callback_ for job events to a _channel_ for job events. The channel is passed to the completer during init, and the completer then owns it as the sole sender. When the completer is stopped, it must close the channel to indicate that there are no more job completion events to be processed.

This moves us closer to having all the key client services be able to be managed as a single pool of services, and they can all have their shutdown initiated in parallel (even though some must still wait for others to shut down first). Importantly, this paves the way for additional services to be added (even by external libraries) without needing to deal with more complex startup & shutdown ordering scenarios. I leveraged the `startstop.StartStopBase` type in all the completers to make this work cleanly.

Additionally, I updated the payload type for completer callbacks/notifications. Rather than sending a single job each time, they can now send slices of events. This is great for the batch completer, because it doesn't have to deal with mutexes or channel sends for every individual job it completed—instead it can send the whole batch through at once.

Currently based on #377.

### DIFF
--- a/subscription_manager.go
+++ b/subscription_manager.go
@@ -1,0 +1,242 @@
+package river
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/riverqueue/river/internal/baseservice"
+	"github.com/riverqueue/river/internal/jobcompleter"
+	"github.com/riverqueue/river/internal/jobstats"
+	"github.com/riverqueue/river/internal/maintenance/startstop"
+	"github.com/riverqueue/river/internal/util/sliceutil"
+	"github.com/riverqueue/river/rivertype"
+)
+
+type subscriptionManager struct {
+	baseservice.BaseService
+	startstop.BaseStartStop
+
+	subscribeCh <-chan []jobcompleter.CompleterJobUpdated
+
+	statsMu        sync.Mutex // protects stats fields
+	statsAggregate jobstats.JobStatistics
+	statsNumJobs   int
+
+	mu               sync.Mutex // protects subscription fields
+	subscriptions    map[int]*eventSubscription
+	subscriptionsSeq int // used for generating simple IDs
+}
+
+func newSubscriptionManager(archetype *baseservice.Archetype, subscribeCh <-chan []jobcompleter.CompleterJobUpdated) *subscriptionManager {
+	return baseservice.Init(archetype, &subscriptionManager{
+		subscribeCh:   subscribeCh,
+		subscriptions: make(map[int]*eventSubscription),
+	})
+}
+
+// ResetSubscribeChan is used to change the channel that the subscription
+// manager listens on. It must only be called when the subscription manager is
+// stopped.
+func (sm *subscriptionManager) ResetSubscribeChan(subscribeCh <-chan []jobcompleter.CompleterJobUpdated) {
+	sm.subscribeCh = subscribeCh
+}
+
+func (sm *subscriptionManager) Start(ctx context.Context) error {
+	_, shouldStart, stopped := sm.StartInit(ctx)
+	if !shouldStart {
+		return nil
+	}
+
+	go func() {
+		// This defer should come first so that it's last out, thereby avoiding
+		// races.
+		defer close(stopped)
+
+		for updates := range sm.subscribeCh {
+			sm.distributeJobUpdates(updates)
+		}
+	}()
+
+	return nil
+}
+
+func (sm *subscriptionManager) Stop() {
+	shouldStop, stopped, finalizeStop := sm.StopInit()
+	if !shouldStop {
+		return
+	}
+
+	<-stopped
+
+	// Remove all subscriptions and close corresponding channels.
+	func() {
+		sm.mu.Lock()
+		defer sm.mu.Unlock()
+
+		for subID, sub := range sm.subscriptions {
+			close(sub.Chan)
+			delete(sm.subscriptions, subID)
+		}
+	}()
+
+	finalizeStop(true)
+}
+
+func (sm *subscriptionManager) logStats(ctx context.Context, svcName string) {
+	sm.statsMu.Lock()
+	defer sm.statsMu.Unlock()
+
+	sm.Logger.InfoContext(ctx, svcName+": Job stats (since last stats line)",
+		"num_jobs_run", sm.statsNumJobs,
+		"average_complete_duration", sm.safeDurationAverage(sm.statsAggregate.CompleteDuration, sm.statsNumJobs),
+		"average_queue_wait_duration", sm.safeDurationAverage(sm.statsAggregate.QueueWaitDuration, sm.statsNumJobs),
+		"average_run_duration", sm.safeDurationAverage(sm.statsAggregate.RunDuration, sm.statsNumJobs))
+
+	sm.statsAggregate = jobstats.JobStatistics{}
+	sm.statsNumJobs = 0
+}
+
+// Handles a potential divide by zero.
+func (sm *subscriptionManager) safeDurationAverage(d time.Duration, n int) time.Duration {
+	if n == 0 {
+		return 0
+	}
+	return d / time.Duration(n)
+}
+
+// Receives updates from the completer and prompts the client to update
+// statistics and distribute jobs into any listening subscriber channels.
+// (Subscriber channels are non-blocking so this should be quite fast.)
+func (sm *subscriptionManager) distributeJobUpdates(updates []jobcompleter.CompleterJobUpdated) {
+	func() {
+		sm.statsMu.Lock()
+		defer sm.statsMu.Unlock()
+
+		for _, update := range updates {
+			stats := update.JobStats
+			sm.statsAggregate.CompleteDuration += stats.CompleteDuration
+			sm.statsAggregate.QueueWaitDuration += stats.QueueWaitDuration
+			sm.statsAggregate.RunDuration += stats.RunDuration
+			sm.statsNumJobs++
+		}
+	}()
+
+	sm.mu.Lock()
+	defer sm.mu.Unlock()
+
+	// Quick path so we don't need to allocate anything if no one is listening.
+	if len(sm.subscriptions) < 1 {
+		return
+	}
+
+	for _, update := range updates {
+		sm.distributeJobEvent(update.Job, jobStatisticsFromInternal(update.JobStats))
+	}
+}
+
+// Distribute a single event into any listening subscriber channels.
+//
+// Job events should specify the job and stats, while queue events should only specify
+// the queue.
+//
+// MUST be called with sm.mu already held.
+func (sm *subscriptionManager) distributeJobEvent(job *rivertype.JobRow, stats *JobStatistics) {
+	var event *Event
+	switch job.State {
+	case rivertype.JobStateCancelled:
+		event = &Event{Kind: EventKindJobCancelled, Job: job, JobStats: stats}
+	case rivertype.JobStateCompleted:
+		event = &Event{Kind: EventKindJobCompleted, Job: job, JobStats: stats}
+	case rivertype.JobStateScheduled:
+		event = &Event{Kind: EventKindJobSnoozed, Job: job, JobStats: stats}
+	case rivertype.JobStateAvailable, rivertype.JobStateDiscarded, rivertype.JobStateRetryable, rivertype.JobStateRunning:
+		event = &Event{Kind: EventKindJobFailed, Job: job, JobStats: stats}
+	case rivertype.JobStatePending:
+		panic("completion subscriber unexpectedly received job in pending state, river bug")
+	default:
+		// linter exhaustive rule prevents this from being reached
+		panic("unreachable state to distribute, river bug")
+	}
+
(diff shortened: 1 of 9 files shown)


bgentry:
The test panic is almost certainly specific to the structuring of some specific tests and not reflective of anything we’ll hit in real usage. I’ll see if I can pin it down tomorrow.

subscription_manager.go
```diff
@@ -0,0 +1,271 @@
...
+
+		for subID, sub := range sm.subscriptions {
+			close(sub.Chan)
+			delete(sm.subscriptions, subID)
+		}
+	}()
+}
+
+func (sm *subscriptionManager) logStats(ctx context.Context, svcName string) {
+	// TODO: don't use mutex + struct internals externally like this
```
brandur, line 74: I'm having a hard time understanding this TODO — can we fix it?

subscription_manager.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+}
+
+// Distribute a single event into any listening subscriber channels.
+//
+// Job events should specify the job and stats, while queue events should only specify
+// the queue.
+//
+// MUST be called with sm.mu already held.
+func (sm *subscriptionManager) _distributeJobEvent(job *rivertype.JobRow, stats *JobStatistics) {
```
brandur, line 132: Hmm, can we drop this underscore convention? I don't think it's really a thing. Out of 22.5k struct functions in the Go codebase, I did find three instances of one, but they're all kind of weird special cases:

```
$ ag --no-break --nofilename 'func \([^)]+\) [A-Za-z]' | wc -l
   22438

$ ag --no-break --nofilename 'func \([^)]+\) _[A-Za-z]' | wc -l
       3
```

```
src/go/types/scope.go
119:func (s *Scope) _InsertLazy(name string, resolve func() Object) bool {

src/runtime/symtab.go
751:func (f funcInfo) _Func() *Func {

src/vendor/golang.org/x/text/unicode/norm/input.go
32:func (in *input) _byte(p int) byte {
```

How about renaming:

* `distributeJobEvents` -> `distributeJobUpdates`
* `_distributeJobEvent` -> `distributeJobEvent` 

Makes them a little more distinct.

If the concern is other places calling private members of this class, the underscore doesn't stop it, and the right fix is to try and get this into its own package.

subscription_manager.go
```diff
@@ -0,0 +1,271 @@
...
+
+		close(sub.Chan)
+
+		delete(sm.subscriptions, subID)
+	}
+
+	return subChan, cancel
+}
+
+// func newAsyncCompleterWithConcurrency(archetype *baseservice.Archetype, exec PartialExecutor, concurrency int, subscribeCh SubscribeChan) *AsyncCompleter {
```
brandur, line 231: Lots of dead code here. Maybe copy/paste error?

subscription_manager.go
```diff
@@ -0,0 +1,271 @@
...
+
+// ResetSubscribeChan is used to change the channel that the subscription
+// manager listens on. It must only be called when the subscription manager is
+// stopped.
+func (sm *subscriptionManager) ResetSubscribeChan(subscribeCh <-chan []jobcompleter.CompleterJobUpdated) {
+	sm.subscribeCh = subscribeCh
+}
+
+func (sm *subscriptionManager) Start(ctx context.Context) error {
+	sm.wg.Add(1)
```
brandur, line 46: Can we make this use the normal stop convention instead with `StartInit`? Protects against double starts, but also gives the client a way to check whether all its services are started.

brandur (commented):
Nice, definitely feels nicer breaking some logic out of the client which is already way too big.

Can you add `subscription_manager_test.go` and put in some basic test cases? It'll be nice because it lets us test the struct in relative isolation. Probably just need a basic check or two and the standard `StartStopStress`.

subscription_manager.go
bgentry: Derp, leftover note to self that I forgot to clean up after I finished the refactor. Thanks

subscription_manager.go
bgentry: Yep shouldn't have made it into the commit, sorry about that!

subscription_manager.go
bgentry: Can't recall where I picked up that convention (Miz maybe?) but I believe I've seen it used for indicating methods which are meant to be called only when a mutex is already held. No problem with changing to your suggestions though.

subscription_manager.go
bgentry: I think I've made the appropriate changes for this to use `startstop.BaseStartStop`, please take a look to see if I did so correctly!

bgentry:
> Can you add `subscription_manager_test.go` and put in some basic test cases? It'll be nice because it lets us test the struct in relative isolation. Probably just need a basic check or two and the standard `StartStopStress`.

@brandur the issue with `startstoptest.Stress()` is that it expects to be able to repeatedly `Start()` and `Stop()` a service with no additional steps. This `subscriptionManager` type doesn't quite work that way, at least not without the `subscribeCh` being closed prior to / concurrently with `Stop()`, and `ResetSubscribeChan` method being called in between runs. This is because its design relies on iterating over the `subscribeCh` in order to know when it has finished receiving all job events and can thus safely shut down, which is the best design I could think of for allowing these services to be capable of parallel shutdown while also waiting as necessary prior to shutting down.

I don't think there's an obvious way to extend the existing test to handle this different use case, however I will add some standalone test coverage to test the correct usage of this service.

bgentry:
@brandur should be good to go, please take another look :pray:

subscription_manager_test.go
```diff
@@ -0,0 +1,126 @@
...
+		// subsequent `Start` to succeed. It's also not friendly for multiple
+		// concurrent calls to `Start` and `Stop`, but this is fine because the only
+		// usage within `Client` is already protected by a mutex.
+		t.Parallel()
+
+		manager, bundle := setup(t)
+
+		subscribeCh := bundle.subscribeCh
+		for i := 0; i < 100; i++ {
+			go func() { close(subscribeCh) }()
```
brandur, line 116: This is racy because the goroutine can run after `subscribeCh` is reset below.

brandur (commented):
@bgentry So I think the difficulty of the stress test not working is somewhat revealing of a more fundamental problem in that as written, the manager basically doesn't respond to `Stop`. `Stop` can be called, but it'll never do anything unless the subscription channel is closed separately.

What do you think about this implementation instead? It modifies things such that it'll respect a stop, but in the case of one, makes sure to clear the subscription channel before leaving, which means that it still correctly clears all events on a client shutdown. This gives us a way to use the stress test because all we need to do is close the channel in advance before calling `startstoptest.Stress`.

Also normalizes things a bit by removing the custom `Stop` implementation, which most services shouldn't need.

``` go
func (sm *subscriptionManager) Start(ctx context.Context) error {
	ctx, shouldStart, stopped := sm.StartInit(ctx)
	if !shouldStart {
		return nil
	}

	go func() {
		// This defer should come first so that it's last out, thereby avoiding
		// races.
		defer close(stopped)

		sm.Logger.DebugContext(ctx, sm.Name+": Run loop started")
		defer sm.Logger.DebugContext(ctx, sm.Name+": Run loop stopped")

		// On shutdown, close and remove all active subscriptions.
		defer func() {
			sm.mu.Lock()
			defer sm.mu.Unlock()

			for subID, sub := range sm.subscriptions {
				close(sub.Chan)
				delete(sm.subscriptions, subID)
			}
		}()

		for {
			select {
			case <-ctx.Done():
				// Distribute remaining subscriptions until the channel is
				// closed. This does make the subscription manager a little
				// problematic in that it requires the subscription channel to
				// be closed before it will fully stop. This always happens in
				// the case of a real client by virtue of the completer always
				// stopping at the same time as the subscription manager, but
				// one has to be careful in tests.
				sm.Logger.DebugContext(ctx, sm.Name+": Stopping; distributing subscriptions until channel is closed")
				for updates := range sm.subscribeCh {
					sm.distributeJobUpdates(updates)
				}

				return

			case updates := <-sm.subscribeCh:
				sm.distributeJobUpdates(updates)
			}
		}
	}()

	return nil
}
```

Here's a patchable diff including addition of an extra test case:

``` diff
diff --git a/subscription_manager.go b/subscription_manager.go
index 1d74a6a..e1df69a 100644
--- a/subscription_manager.go
+++ b/subscription_manager.go
@@ -44,7 +44,7 @@ func (sm *subscriptionManager) ResetSubscribeChan(subscribeCh <-chan []jobcomple
 }
 
 func (sm *subscriptionManager) Start(ctx context.Context) error {
-	_, shouldStart, stopped := sm.StartInit(ctx)
+	ctx, shouldStart, stopped := sm.StartInit(ctx)
 	if !shouldStart {
 		return nil
 	}
@@ -54,34 +54,44 @@ func (sm *subscriptionManager) Start(ctx context.Context) error {
 		// races.
 		defer close(stopped)
 
-		for updates := range sm.subscribeCh {
-			sm.distributeJobUpdates(updates)
-		}
-	}()
-
-	return nil
-}
-
-func (sm *subscriptionManager) Stop() {
-	shouldStop, stopped, finalizeStop := sm.StopInit()
-	if !shouldStop {
-		return
-	}
+		sm.Logger.DebugContext(ctx, sm.Name+": Run loop started")
+		defer sm.Logger.DebugContext(ctx, sm.Name+": Run loop stopped")
 
-	<-stopped
+		// On shutdown, close and remove all active subscriptions.
+		defer func() {
+			sm.mu.Lock()
+			defer sm.mu.Unlock()
 
-	// Remove all subscriptions and close corresponding channels.
-	func() {
-		sm.mu.Lock()
-		defer sm.mu.Unlock()
+			for subID, sub := range sm.subscriptions {
+				close(sub.Chan)
+				delete(sm.subscriptions, subID)
+			}
+		}()
 
-		for subID, sub := range sm.subscriptions {
-			close(sub.Chan)
-			delete(sm.subscriptions, subID)
+		for {
+			select {
+			case <-ctx.Done():
+				// Distribute remaining subscriptions until the channel is
+				// closed. This does make the subscription manager a little
+				// problematic in that it requires the subscription channel to
+				// be closed before it will fully stop. This always happens in
+				// the case of a real client by virtue of the completer always
+				// stopping at the same time as the subscription manager, but
+				// one has to be careful in tests.
+				sm.Logger.DebugContext(ctx, sm.Name+": Stopping; distributing subscriptions until channel is closed")
+				for updates := range sm.subscribeCh {
+					sm.distributeJobUpdates(updates)
+				}
+
+				return
+
+			case updates := <-sm.subscribeCh:
+				sm.distributeJobUpdates(updates)
+			}
 		}
 	}()
 
-	finalizeStop(true)
+	return nil
 }
 
 func (sm *subscriptionManager) logStats(ctx context.Context, svcName string) {
diff --git a/subscription_manager_test.go b/subscription_manager_test.go
index 5406d4a..be00500 100644
--- a/subscription_manager_test.go
+++ b/subscription_manager_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/riverqueue/river/internal/jobcompleter"
 	"github.com/riverqueue/river/internal/jobstats"
 	"github.com/riverqueue/river/internal/riverinternaltest"
+	"github.com/riverqueue/river/internal/riverinternaltest/startstoptest"
 	"github.com/riverqueue/river/internal/riverinternaltest/testfactory"
 	"github.com/riverqueue/river/internal/util/ptrutil"
 	"github.com/riverqueue/river/riverdriver"
@@ -113,7 +114,7 @@ func Test_SubscriptionManager(t *testing.T) {
 
 		subscribeCh := bundle.subscribeCh
 		for i := 0; i < 100; i++ {
-			go func() { close(subscribeCh) }()
+			close(subscribeCh)
 			manager.Stop()
 
 			subscribeCh = make(chan []jobcompleter.CompleterJobUpdated, 1)
@@ -123,4 +124,16 @@ func Test_SubscriptionManager(t *testing.T) {
 		}
 		close(subscribeCh)
 	})
+
+	t.Run("StartStopStress", func(t *testing.T) {
+		t.Parallel()
+
+		svc, bundle := setup(t)
+
+		// Close the subscription channel in advance so that stops can leave
+		// successfully.
+		close(bundle.subscribeCh)
+
+		startstoptest.Stress(ctx, t, svc)
+	})
 }
```

bgentry:
@brandur it's possible I'm misunderstanding it, but from what I'm seeing your proposal doesn't guarantee that the subscription manager will wait for all events prior to shutting down. That was the whole purpose of the channel and it was the reason that the subscription manager shutdown was previously a separate step after the services were shut down: being able to know that there are no more events coming, and therefore if we start closing client subscriptions they will not miss a single event (even after `Stop()` is called).

That particular property is one I think we need to preserve, as some of the upcoming stuff needs to be able to ensure that it has consumed and handled all incoming events for all processed jobs prior to stopping.

Might be easier to discuss offline?

bgentry:
@brandur I definitely misread the patch on the first pass and it in fact looks good :pray: I've applied it here.

brandur (approved):
Awesome. Thx!

PR #383, opened 2024-06-04
customizable MaxAttempts default value

### LINKED ISSUE #381: change default retry count
I can't seem to find an easy way to configure the max attempts for a client. 

from what i see, i can only configure it on the args or insert level, and then it is the hard coded 25. 

https://github.com/riverqueue/river/blob/master/client.go#L1255

It would be nice if this could be customizable somehow through the clients config, either through configuring the said 25 number, or some sort of insertOpts middleware, so that i could set the default job to have 1 max attempt, and not have to configure the majority of my jobs to have 1 max attempt.

bgentry:
This adds a `MaxAttempts` setting to `Config` to enable the default max attempts to be customized (instead of the global default `MaxAttemptsDefault` of 25).

Fixes #381.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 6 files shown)


brandur (approved):
Wicked, LGTM.

PR #385, opened 2024-06-10
expose DisableStaggerStart via TestOnlyConfig

bgentry:
I ran into an issue with some external testing where I needed to run a full River client. These tests were quite slow, and I realized it was because of `StaggerStartStop`—a setting that is currently not exposed outside the `river` package.

I'm not sure this is the right way to expose it. If we end up having a set of related knobs that can be tweaked for test usage (lowering buffers, sleep durations, etc) it's likely most of the time they'll be used all together. Open to other ways to expose this.

### DIFF
--- a/client.go
+++ b/client.go
@@ -202,6 +202,10 @@ type Config struct {
 	// Defaults to DefaultRetryPolicy.
 	RetryPolicy ClientRetryPolicy
 
+	// TestOnly holds configuration for test-only settings that should generally
+	// not be used outside of test suites.
+	TestOnly TestOnlyConfig
+
 	// Workers is a bundle of registered job workers.
 	//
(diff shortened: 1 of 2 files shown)


brandur:
As discussed on Slack, my fear with the granularity here is that it's not very forward compatibility friendly. Meaning that in the future if more options were to be added like `DisableStaggerStart` that most tests would want to disable, people trying to disable this type of thing in tests wouldn't get the new ones unless they were paying very close attention to the changelog.

I was thinking about this one a little more overnight — what do you think about exposing `TestOnly boolean` for the time being, and then if the need called for it in the future, we could then expose a more granular `TestOnlyConfig TestOnlyConfig` property with more options?

bgentry:
Superseded by #414.

PR #390, opened 2024-06-11
add JobDelete / JobDeleteTx APIs to Client

bgentry:
This adds new APIs to permanently delete a job from the job table, so long as it is not currently running.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1047,6 +1047,23 @@ func (c *Client[TTx]) jobCancel(ctx context.Context, exec riverdriver.Executor, 
 	})
 }
 
+// JobDelete deletes the job with the given ID from the database, returning the
+// deleted row if it was deleted. Jobs in the running state are not deleted,
+// instead returning rivertype.ErrJobRunning.
+func (c *Client[TTx]) JobDelete(ctx context.Context, id int64) (*rivertype.JobRow, error) {
+	return c.driver.GetExecutor().JobDelete(ctx, id)
+}
+
+// JobDelete deletes the job with the given ID from the database, returning the
+// deleted row if it was deleted. Jobs in the running state are not deleted,
+// instead returning rivertype.ErrJobRunning. This variant lets a caller retry a
+// job atomically alongside other database changes. A deleted job isn't deleted
+// until the transaction commits, and if the transaction rolls back, so too is
+// the deleted job.
+func (c *Client[TTx]) JobDeleteTx(ctx context.Context, tx TTx, id int64) (*rivertype.JobRow, error) {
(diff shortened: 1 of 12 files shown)


riverdriver/riverpgxv5/river_pgx_v5_driver.go
```diff
@@ -99,6 +99,20 @@ func (e *Executor) JobCountByState(ctx context.Context, state rivertype.JobState
 	return int(numJobs), nil
 }
 
+func (e *Executor) JobDelete(ctx context.Context, id int64) (bool, error) {
+	job, err := e.queries.JobDelete(ctx, e.dbtx, id)
+	if err != nil {
+		return false, interpretError(err)
+	}
+	if job == nil {
+		return false, rivertype.ErrNotFound
```
brandur, line 108: Doesn't sqlc always return a not found error for a `:one` query rather than return a `nil`?

What do you think about adding a driver test case that verifies that a `JobDelete` on a non-existent row just to make sure the right thing happens here?

client.go
```diff
@@ -1047,6 +1047,22 @@ func (c *Client[TTx]) jobCancel(ctx context.Context, exec riverdriver.Executor,
 	})
 }
 
+// JobDelete deletes the job with the given ID from the database, returning
+// whether or not the job was deleted along with a possible error. Jobs in the
+// running state are not deleted.
+func (c *Client[TTx]) JobDelete(ctx context.Context, id int64) (bool, error) {
```
brandur, line 1053: Thoughts on returning the deleted job row instead of a bool? Feels like it gives you gives you more information and not a lot of downside.

bgentry: I think that will work if we also make the change to return an error when the job is running, I like that better.

riverdriver/riverpgxv5/river_pgx_v5_driver.go
bgentry: Added! Thought I already had that one but obviously missed it. And you're correct, this clause is unnecessary.

bgentry:
Also fixed another flaky test with the same cause as #391.

rivertype/river_type.go
```diff
@@ -13,6 +13,10 @@ import (
 // return this error.
 var ErrNotFound = errors.New("not found")
 
+// ErrJobRunning is returned when a job is attempted to be deleted while it's
+// running.
+var ErrJobRunning = errors.New("job is running")
```
brandur, line 18: What do you think about a slightly more descriptive error message here in case this error bubbles up to an end user somewhere? Something like:

```suggestion
var ErrJobRunning = errors.New("running jobs cannot be deleted")
```

brandur (approved):
lgtm.

PR #394, opened 2024-06-13
Prepare version 0.7.0

bgentry:
Prepare version 0.7.0 for release. This includes the bugfix from #376
(allow `StopAndCancel` to be called when a `Stop` is already ongoing),
and new features from #383 (allow customizing default `MaxAttempts` at
the client level) and #390 (`JobDelete` + `JobDeleteTx` APIs).


brandur (approved):
Awesome. Let's do it.

PR #399, opened 2024-06-24
Add AddQueue to the Client

### LINKED ISSUE #396: Dynamic queue handling
We have a use-case where we need to add queues dynamically. The description of the use-case: we have multiple `Kind` of jobs, and one particular is the `manage_rewards`. This `Kind` connects to a server where the rate limit handled per route. Sometimes happens that one route blocks the other jobs because it's not handled properly so we figured out that we create queues per route, since those handled parallel. But for this we want to add queues in runtime. 

I tweaked the code and created the following method:

```go
func (c *Client[TTx]) AddQueue(queue string, queueConfig QueueConfig) {
	archetype := &baseservice.Archetype{
		Logger:     c.config.Logger,
		Rand:       randutil.NewCryptoSeededConcurrentSafeRand(),
		TimeNowUTC: func() time.Time { return time.Now().UTC() },
	}

	c.producersByQueueName[queue] = newProducer(archetype, c.driver.GetExecutor(), &producerConfig{
		ClientID:          c.config.ID,
		Completer:         c.completer,
		ErrorHandler:      c.config.ErrorHandler,
		FetchCooldown:     c.config.FetchCooldown,
		FetchPollInterval: c.config.FetchPollInterval,
		JobTimeout:        c.config.JobTimeout,
		MaxWorkers:        queueConfig.MaxWorkers,
		Notifier:          c.notifier,
		Queue:             queue,
		RetryPolicy:       c.config.RetryPolicy,
		SchedulerInterval: c.config.schedulerInterval,
		StatusFunc:        c.monitor.SetProducerStatus,
		Workers:           c.config.Workers,
	})
	c.monitor.InitializeProducerStatus(queue)
}
```

After that I ran the following:

```go
if err := riverClient.Stop(ctx); err != nil {
	t.Fatal(err)
}

riverClient.AddQueue("new_queue", river.QueueConfig{
	MaxWorkers: 10,
})

	t.Fatal(err)
}
```

If you agree with these changes or you have any suggestion let me know and I can send a PR about it. We tested with 1800 queue's for the same Kind with 10.000 jobs per queue. That worked all fine. 

Note we are using River in production with millions of jobs and that works awesome.

PumpkinSeed:
Closes #396 

Based on the proposal I added AddQueue/RemoveQueue. This won't do anything just add the queues, so we need to "Restart" the client after that. 

What do you think about to add StartWorkContext in the AddQueue section

```go
if err := producer.StartWorkContext(fetchCtx, workCtx); err != nil {
	stopProducers()
	stopServicesOnError()
	return err
}
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -259,10 +259,7 @@ func (c *Config) validate() error {
 	}
 
 	for queue, queueConfig := range c.Queues {
-		if queueConfig.MaxWorkers < 1 || queueConfig.MaxWorkers > QueueNumWorkersMax {
-			return fmt.Errorf("invalid number of workers for queue %q: %d", queue, queueConfig.MaxWorkers)
-		}
-		if err := validateQueueName(queue); err != nil {
+		if err := queueConfig.validate(queue); err != nil {
 			return err
 		}
 	}
@@ -294,6 +291,17 @@ type QueueConfig struct {
 	//
 	// Requires a minimum of 1, and a maximum of 10,000.
 	MaxWorkers int
+}
+
+func (c QueueConfig) validate(queueName string) error {
+	if c.MaxWorkers < 1 || c.MaxWorkers > QueueNumWorkersMax {
+		return fmt.Errorf("invalid number of workers for queue %q: %d", queueName, c.MaxWorkers)
+	}
+	if err := validateQueueName(queueName); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 // Client is a single isolated instance of River. Your application may use
@@ -510,23 +518,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		client.services = append(client.services, client.elector)
 
 		for queue, queueConfig := range config.Queues {
-			client.producersByQueueName[queue] = newProducer(archetype, driver.GetExecutor(), &producerConfig{
-				ClientID:           config.ID,
-				Completer:          client.completer,
-				ErrorHandler:       config.ErrorHandler,
-				FetchCooldown:      config.FetchCooldown,
-				FetchPollInterval:  config.FetchPollInterval,
-				JobTimeout:         config.JobTimeout,
-				MaxWorkers:         queueConfig.MaxWorkers,
-				Notifier:           client.notifier,
-				Queue:              queue,
-				QueueEventCallback: client.subscriptionManager.distributeQueueEvent,
-				RetryPolicy:        config.RetryPolicy,
-				SchedulerInterval:  config.schedulerInterval,
-				StatusFunc:         client.monitor.SetProducerStatus,
-				Workers:            config.Workers,
-			})
-			client.monitor.InitializeProducerStatus(queue)
+			client.producersByQueueName[queue] = client.addProducer(queue, queueConfig)
 		}
 
 		client.services = append(client.services,
@@ -618,34 +610,16 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 	return client, nil
 }
 
-func (c *Client[TTx]) AddQueue(queueName string, queueConfig QueueConfig) {
+func (c *Client[TTx]) AddQueue(queueName string, queueConfig QueueConfig) error {
+	if err := queueConfig.validate(queueName); err != nil {
+		return err
+	}
+
 	c.producersByQueueNameMu.Lock()
 	defer c.producersByQueueNameMu.Unlock()
-	c.producersByQueueName[queueName] = newProducer(&c.baseService.Archetype, c.driver.GetExecutor(), &producerConfig{
-		ClientID:          c.config.ID,
-		Completer:         c.completer,
-		ErrorHandler:      c.config.ErrorHandler,
-		FetchCooldown:     c.config.FetchCooldown,
-		FetchPollInterval: c.config.FetchPollInterval,
-		JobTimeout:        c.config.JobTimeout,
-		MaxWorkers:        queueConfig.MaxWorkers,
-		Notifier:          c.notifier,
-		Queue:             queueName,
-		RetryPolicy:       c.config.RetryPolicy,
-		SchedulerInterval: c.config.schedulerInterval,
-		StatusFunc:        c.monitor.SetProducerStatus,
-		Workers:           c.config.Workers,
-	})
-	c.monitor.InitializeProducerStatus(queueName)
-}
+	c.producersByQueueName[queueName] = c.addProducer(queueName, queueConfig)
 
-func (c *Client[TTx]) RemoveQueue(queueName string) {
-	c.producersByQueueNameMu.Lock()
-	defer c.producersByQueueNameMu.Unlock()
-	delete(c.producersByQueueName, queueName)
-
-	// Remove queue from currentSnapshot.Producers
-	c.monitor.RemoveProducerStatus(queueName)
+	return nil
 }
 
 // Start starts the client's job fetching and working loops. Once this is called,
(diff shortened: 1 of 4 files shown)


client.go
```diff
@@ -616,6 +618,36 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
...
+		MaxWorkers:        queueConfig.MaxWorkers,
+		Notifier:          c.notifier,
+		Queue:             queueName,
+		RetryPolicy:       c.config.RetryPolicy,
+		SchedulerInterval: c.config.schedulerInterval,
+		StatusFunc:        c.monitor.SetProducerStatus,
+		Workers:           c.config.Workers,
+	})
+	c.monitor.InitializeProducerStatus(queueName)
+}
```
bgentry, line 640: This method is going to need to do some validation on the `QueueConfig` as [we currently do when initializing the client](https://github.com/riverqueue/river/blob/220a636820f37b9a7b16309d2863ef94124dc3db/client.go#L260-L267), and it will also need to return an `error` if validation fails.

We should probably extract that to an internal `func (c QueueConfig) validate() error` similar to what we have for the top level `Config` so that we can use the same logic in each place. And of course we'll need to make sure this new surface area (add/remove) is fully covered by tests in this regard.

client.go
```diff
@@ -616,6 +618,36 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
...
+
+func (c *Client[TTx]) RemoveQueue(queueName string) {
+	c.producersByQueueNameMu.Lock()
+	defer c.producersByQueueNameMu.Unlock()
+	delete(c.producersByQueueName, queueName)
```
bgentry, line 645: Should this one also return an `error`? We could reuse `rivertype.ErrNotFound` in case the queue does not exist.

client.go
```diff
@@ -616,6 +618,36 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
...
+
+	// Remove queue from currentSnapshot.Producers
+	c.monitor.RemoveProducerStatus(queueName)
```
bgentry, line 648: Safe removal is a bit trickier than this. We need to actually shut down the producer somehow, and then the producer itself should be responsible for reporting its status to the monitor until it is shut down. At that point, yeah, we need to somehow remove it from the monitor's map so we don't keep old state around in that map forever.

Maybe `SetProducerStatus` should just be smart enough to do that last part automatically when the producer is stopped?

@brandur thoughts on how to safely trigger the single producer to shut down? And is this API suitable for an operation which may take awhile (until all running jobs go away)? Keeping in mind there is no obvious way to cancel the context just for a single producer, and we may also need to differentiate between `Stop` and `StopAndCancel` for each individual producer—otherwise how will the producer know when it needs to cancel the context and try to more aggressively shut down running jobs?

In either case we still need to somehow wait for clean shutdown to complete, which may actually never happen before the program exits if the running jobs never exit.

client_monitor.go
```diff
@@ -67,6 +67,13 @@ func (m *clientMonitor) SetProducerStatus(queueName string, status componentstat
 	m.bufferStatusUpdate()
 }
 
+func (m *clientMonitor) RemoveProducerStatus(queueName string) {
+	m.statusSnapshotMu.Lock()
+	defer m.statusSnapshotMu.Unlock()
+	delete(m.currentSnapshot.Producers, queueName)
+	m.bufferStatusUpdate()
+}
```
bgentry, line 75: I think we can probably merge this into `SetProducer` and decide whether or not to remove based upon the status.

bgentry (changes requested):
Thanks for the idea! I'm not opposed to the idea of adding this, though I think it turns out to be a fair bit more complex than what this PR demonstrates. Adding is actually easy IMO, it's removal that is a lot trickier in practice and raises a lot of UX questions.

Interested in whether @brandur has ideas on how to address these. One option would be to reduce the scope of this PR to only `AddQueue` at this time and punt on `RemoveQueue` because of its much higher complexity.

client.go
```diff
@@ -616,6 +618,36 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
...
+		JobTimeout:        c.config.JobTimeout,
+		MaxWorkers:        queueConfig.MaxWorkers,
+		Notifier:          c.notifier,
+		Queue:             queueName,
+		RetryPolicy:       c.config.RetryPolicy,
+		SchedulerInterval: c.config.schedulerInterval,
+		StatusFunc:        c.monitor.SetProducerStatus,
+		Workers:           c.config.Workers,
+	})
+	c.monitor.InitializeProducerStatus(queueName)
```
brandur, line 639: Just given this is a lot of duplicative code, let's move this to a new `addProducer(queueName string, queueConfig QueueConfig)` internal helper that can share it with `Start`.

brandur (commented):
Thanks for the PR!

A couple larger points:

> What do you think about to add StartWorkContext in the AddQueue section

Yeah, I think we'll need something like this. Ideally, if the client is started, `AddQueue` starts the producer. If not, it doesn't.

Also, rather than stop all producers on failure (if the client was started), it maybe should just return an error from the `AddQueue` function.

Can you take a look at `PeriodicJobs()` too? See here:

https://github.com/riverqueue/river/blob/c9dd29156b59797a813f5459231c0cf80f8718eb/client.go#L1598

I think we should follow that pattern and make these `Queues().Add(...)` and `Queues.Remove(...)`. It sorts the functions together (since client has a lot of functions already), and future proofs us in case we want to add any future queues-related functions.

And lastly, tests of course! Although those can wait until the API has fully stabilized.

client.go
brandur: Yeah definitely pretty tricky. I find the existing `StopAndCancel` to be quite confusing to reason about, so it'd be nice if we could avoid adding another version of that here.

One possibility might be that `RemoveQueue` will only wait for a graceful shutdown, and if a hard shutdown is necessary, the caller will just have to use `StopAndCancel`.

Also, if it's not needed, I like your idea of just punting on `RemoveQueue` for now.

_Or_, a possible variation of that is to have `RemoveQueue`, but only allow it to be called if the client is stopped, which would allow it if necessary, but make a lot of the harder problems go away. If we went that route though, we should make sure to future-proof the API (e.g. by making it take a `ctx` if we think there will be a hard stop later) in case we want to remove this limitation in the future.

PumpkinSeed:
I agree with the reduction of the scope of the PR. I will focus on AddQueue and remove the RemoveQueue, I will address it in a next PR.

PumpkinSeed:
Is there any documentation about setting up development environment? I'm using Arch Linux and it would be nice to use postgres image/container for testing. I found docs/development.md but most of these using the raw postgres client, however it would be nice to set an env var or something like that so we can customize the postgres connection.

PumpkinSeed:
I added the requested changes. 

I still have a question whether I need to return `s.context` in the StartInit at the following section or not:

```
if s.started {
	return ctx, false, nil
}
```

PumpkinSeed:
@brandur @bgentry Do you have any update on this?

brandur:
Hey @PumpkinSeed, thanks for the updates here.

This is a bit of a tricky change, and there's quite a bit of nuance that's quite subtle — e.g. believe it or not, the way you're checking for `IsStarted` here is racy because a shutdown could have been finished between the check on `IsStarted` and the call to `StartWorkContext`. There's also a couple of other changes need like the addition of a `Queues()` bundle like I mentioned above.

Do you mind if we take this one over to get it over the finish line?

PumpkinSeed:
> Do you mind if we take this one over to get it over the finish line?

I'm totally fine with it. Thanks for that.

brandur:
Opened #410 as an alternative.

brandur:
Closing in favor of #410.

PR #402, opened 2024-06-26
Add transaction-based variants for all queue functions

brandur:
I noticed while trying to write tests for River UI that it wasn't
possible to use any of the queue endpoints with a test transaction
because unlike functions `Client.Insert` which have a transaction based
`InsertTx` variant, queue functions can only work on the main pool.

Here, add transaction variants for all queue functions:

* `QueueGetTx`
* `QueueListTx`
* `QueuePauseTx`
* `QueueResumeTx`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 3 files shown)


brandur:
@bgentry I was trying to write more tests for River UI endpoints, but found that since I was using test transactions, it wasn't possible to test the queue-related stuff because transaction-based functions were missing. Mind taking a look at this one?

bgentry (approved):
:fire:

PR #403, opened 2024-06-26
elector: resign leadership w/ WithoutCancel ctx, not background

bgentry:
Borrowing an idea from #401, rather than using a background context for a shutdown operation that we don't want to be prematurely immediately cancelled, we can instead use a `context.WithoutCancel(ctx)` to maintain non-cancel parts of the context.

I also updated logger calls in here to use the `Context` variants for consitency.

### DIFF
--- a/internal/leadership/elector.go
+++ b/internal/leadership/elector.go
@@ -177,7 +177,7 @@ func (e *Elector) Start(ctx context.Context) error {
 					continue // lost leadership reelection; unusual but not a problem; don't log
 				}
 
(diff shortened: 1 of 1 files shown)


internal/leadership/elector.go
```diff
@@ -375,14 +375,14 @@ func (e *Elector) attemptResignLoop(ctx context.Context) {
 
 	// This does not inherit the parent context because we want to give up leadership
 	// even during a shutdown. There is no way to short-circuit this.
```
brandur, line 377: Maybe tweak this comment a little?

brandur (approved):

PR #405, opened 2024-06-28
Fix broken links from Godoc

brandur:
I just happened to be looking at our Godoc and noticed that some of the
links are broken [1]. Some were missing their definitions, but the new
River UI ones use a Markdown syntax that's not supported by Godoc.

I also updated the README with the same changes. They're not necessary
there, but IMO better to keep these two docs as synchronized as possible
so if we need to make wider changes later it's easy to change both.

[1] https://pkg.go.dev/github.com/riverqueue/river

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -2,7 +2,7 @@
 Package river is a robust high-performance job processing system for Go and
 Postgres.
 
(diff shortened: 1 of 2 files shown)


bgentry (approved):
doh, thank you!

brandur:
Thanks!

PR #408, opened 2024-06-29
Idempotent queue pause and resume

brandur:
I was writing tests for River UI and found that when trying to pause an
already paused queue or resume an unpaused queue, River returns a "not
found" error.

This was quite surprising, so much so that it took me a good half hour
of debugging before I considered the fact that it might actually be a
problem in upstream River. Even if an argument were to be made that
pausing an already paused queue should be an error like "queue already
paused", returning "not found" is misleading and guaranteed to result
in confused people beyond just myself.

I think it's fine for pause and resume to be considered idempotent
operations. If a paused queue is paused again, there's no real damage
done, especially if we keep the original paused time, which we do in
this implementation.

Similarly, make an update so that when pausing or resuming using  the
all queues string (`*`), don't return "not found" error if there are no
queues in the database. Instead, just no-op.

### DIFF
--- a/riverdriver/riverpgxv5/internal/dbsqlc/river_queue.sql
+++ b/riverdriver/riverpgxv5/internal/dbsqlc/river_queue.sql
@@ -48,32 +48,49 @@ LIMIT @limit_count::integer;
 
 -- name: QueuePause :execresult
 WITH queue_to_update AS (
-    SELECT name
+    SELECT name, paused_at
     FROM river_queue
-    WHERE CASE WHEN @name::text = '*' THEN true ELSE river_queue.name = @name::text END
-      AND paused_at IS NULL
+    WHERE CASE WHEN @name::text = '*' THEN true ELSE name = @name END
     FOR UPDATE
+),
+updated_queue AS (
+    UPDATE river_queue
+    SET
+        paused_at = now(),
+        updated_at = now()
+    FROM queue_to_update
+    WHERE river_queue.name = queue_to_update.name
+        AND river_queue.paused_at IS NULL
+    RETURNING river_queue.*
 )
-
-UPDATE river_queue
-SET
-    paused_at = now(),
-    updated_at = now()
-FROM queue_to_update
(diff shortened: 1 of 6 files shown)


riverdriver/riverpgxv5/internal/dbsqlc/river_queue.sql
```diff
@@ -48,16 +48,14 @@ LIMIT @limit_count::integer;
...
     FOR UPDATE
 )
-
 UPDATE river_queue
 SET
-    paused_at = now(),
+    paused_at = coalesce(queue_to_update.paused_at, now()),
     updated_at = now()
 FROM queue_to_update
 WHERE river_queue.name = queue_to_update.name;
```
bgentry, line 61: To avoid editing a row for no reason, we could take an approach we've done in some other queries: do the update in a CTE, and then for the final return do a union w/ the updated row and original row to return only one of them.

Not a high throughput query so it doesn't matter much, but it would be a nice tweak to stay consistent in our approach.

bgentry (approved):
LGTM with a suggestion, this definitely should have worked this way from the beginning. Also we probably want a changelog entry.

riverdriver/riverpgxv5/internal/dbsqlc/river_queue.sql
brandur: Okay, good call. Updated.

brandur:
@bgentry Actually, while looking through that query code I found one other related problem: when you pause/resume using the all queues `*` string and there are no queues in the database, it returns a "not found" error.

This also feels a bit wrong to me, so I changed it to be a no-op. Do you agree with that behavior? If so, maybe take one more review pass since I added some more code and test cases.

bgentry:
Yes, I agree, a no op seems better in that case!

bgentry (approved):

brandur:
Great, thanks! Also added changelog.

PR #415, opened 2024-07-03
Track started state of start/stop services + cleaner functions instead of channels

brandur:
This is one that I've had on the backburner for a while, but
deprioritized since it'll take review bandwidth and we had more
important things going on. I pulled it up to the top because we've seen
seeing a crazy rate of intermittent failures out of `AddManyAfterStart`
in the periodic job enqueuer [1] [2]. I'm not entirely sure yet because
I'm having trouble reproducing it locally, but reading the code I
believe the problem is that we start the client and then start adding
periodic jobs, but the start is a race condition because we're not
guaranteed that the service performed a loop before the periodic jobs
were added.

    startService(t, svc)

    svc.Add(
            &PeriodicJob{ScheduleFunc: periodicIntervalSchedule(500 * time.Millisecond), ConstructorFunc: jobConstructorFunc("periodic_job_500ms", false)},
    )
    svc.Add(
            &PeriodicJob{ScheduleFunc: periodicIntervalSchedule(500 * time.Millisecond), ConstructorFunc: jobConstructorFunc("periodic_job_500ms_start", false), RunOnStart: true},
    )

If that is the case, I believe this change can help with that.

Its main feature is that it adds a `Started` channel to the start/stop
infrastructure that's closed when a service finishes starting up:

    type Service interface {
        ...

        // Started returns a channel that's closed when a service finishes starting,
        // or if failed to start and is stopped instead. It can be used in
        // conjunction with WaitAllStarted to verify startup of a constellation of
        // services.
        Started() <-chan struct{}

We then provide an easy helper called `WaitAllStarted` that lets callers
easily wait for a series of startups to come up:

    // WaitAllStarted waits until all the given services are started (or stopped in
    // a degenerate start scenario, like if context is cancelled while starting up).
    //
    // Unlike StopAllParallel, WaitAllStarted doesn't bother with parallelism
    // because the services themselves have already backgrounded themselves, and we
    // have to wait until the slowest service has started anyway.
    func WaitAllStarted(services ...Service) {
        ...

This allows us to modify tests like the periodic job enqueuer above to
wait on a service start, thereby hopefully fixing our intermittency
problems:

    startService := func(t *testing.T, svc *PeriodicJobEnqueuer) {
        t.Helper()

        require.NoError(t, svc.Start(ctx))
        t.Cleanup(svc.Stop)

        startstop.WaitAllStarted(svc)
    }

We also modify start/stop's returns slightly so that a `started`
function is added to enable this new feature, but also changes `stopped`
from a channel to a function, which looks a bit nicer and has better
ergonomics compared to closing a channel:

    func (m *QueueMaintainer) Start(ctx context.Context) error {
        ctx, shouldStart, started, stopped := m.StartInit(ctx)
        if !shouldStart {
            return nil
        }

        go func() {
            started()
            defer stopped() // this defer should come first so it's last out

            ...

[1] https://github.com/riverqueue/river/actions/runs/9770889034/job/26972789182?pr=413
[2] https://github.com/riverqueue/river/actions/runs/9770889034/job/26972789334?pr=413

### DIFF
--- a/client.go
+++ b/client.go
@@ -321,7 +321,7 @@ type Client[TTx any] struct {
 	queueMaintainer      *maintenance.QueueMaintainer
 	services             []startstop.Service
 	subscriptionManager  *subscriptionManager
-	stopped              chan struct{}
+	stopped              <-chan struct{}
 	testSignals          clientTestSignals
 	uniqueInserter       *dbunique.UniqueInserter
 
@@ -633,17 +633,17 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 // jobs, but will also cancel the context for any currently-running jobs. If
 // using StopAndCancel, there's no need to also call Stop.
 func (c *Client[TTx]) Start(ctx context.Context) error {
-	fetchCtx, shouldStart, stopped := c.baseStartStop.StartInit(ctx)
+	fetchCtx, shouldStart, started, stopped := c.baseStartStop.StartInit(ctx)
 	if !shouldStart {
 		return nil
 	}
 
-	c.stopped = stopped
+	c.stopped = c.baseStartStop.Stopped()
 
-	stopProducers := func() {
-		startstop.StopAllParallel(sliceutil.Map(
+	producersAsServices := func() []startstop.Service {
+		return sliceutil.Map(
 			maputil.Values(c.producersByQueueName),
-			func(p *producer) startstop.Service { return p }),
+			func(p *producer) startstop.Service { return p },
 		)
 	}
 
@@ -723,7 +723,7 @@ func (c *Client[TTx]) Start(ctx context.Context) error {
 			producer := producer
 
 			if err := producer.StartWorkContext(fetchCtx, workCtx); err != nil {
-				stopProducers()
+				startstop.StopAllParallel(producersAsServices())
 				stopServicesOnError()
 				return err
 			}
@@ -731,7 +731,7 @@ func (c *Client[TTx]) Start(ctx context.Context) error {
 
 		return nil
 	}(); err != nil {
-		defer close(stopped)
+		defer stopped()
 		if errors.Is(context.Cause(fetchCtx), startstop.ErrStop) {
 			return rivercommon.ErrShutdown
 		}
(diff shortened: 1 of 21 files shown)


brandur:
@bgentry Okay this turned out not to fix the intermittency problem (opened #416 for that instead), but I still think it's not a bad refactor. Mind taking a look?

client.go
```diff
@@ -718,23 +718,36 @@ func (c *Client[TTx]) Start(ctx context.Context) error {
...
 		return err
 	}
 
 	go func() {
-		defer close(stopped)
+		// Wait for all subservices to start up before signaling our own start.
+		// This isn't strictly needed, but gives tests a way to fully confirm
+		// that all goroutines for subservices are spun up before continuing.
+		//
+		// Stops also cancel the "started" channel, so in case of a context
```
bgentry, line 741: At first I thought there was a typo, but maybe it's just worded awkwardly. This feels a bit clearer imo, up to you:

```suggestion
		// Stop also cancels the "started" channel, so in case of a context
```

internal/maintenance/startstop/start_stop.go
```diff
@@ -97,19 +103,37 @@ type BaseStartStop struct {
 //	}
 //
 //	...
-func (s *BaseStartStop) StartInit(ctx context.Context) (context.Context, bool, chan struct{}) {
+func (s *BaseStartStop) StartInit(ctx context.Context) (context.Context, bool, func(), func()) {
```
bgentry, line 106: This is where I start disliking the lint rule discouraging named return values. I mainly just want to see a label on things when there are this many, especially when there are consecutive values of the same type.

bgentry (approved):
Seems like a good improvement. I'm now even more feeling the desire to reuse this `startstop` stuff elsewhere so we might want to think about pulling it into a separate package.

internal/maintenance/startstop/start_stop.go
brandur: Yeah, definitely fair. I tried to write some good docs to compensate for it in this case. Will hopefully help.

brandur:
Thanks!

Issue #386, opened 2024-06-10
flaky TestPeriodicJobEnqueuer


bgentry:
From [this run](https://github.com/riverqueue/river/actions/runs/9442380723/job/26004445991):

```
--- FAIL: TestPeriodicJobEnqueuer (0.00s)
    --- FAIL: TestPeriodicJobEnqueuer/RemoveAfterStart (0.08s)
        periodic_job_enqueuer_test.go:447: 
            	Error Trace:	/home/runner/work/river/river/internal/maintenance/periodic_job_enqueuer_test.go:87
            	            				/home/runner/work/river/river/internal/maintenance/periodic_job_enqueuer_test.go:447
            	Error:      	"[0xc00042c480]" should have 0 item(s), but has 1
            	Test:       	TestPeriodicJobEnqueuer/RemoveAfterStart
            	Messages:   	Expected to find exactly 0 job(s) of kind: periodic_job_500ms, but found 1
```

PR #419, opened 2024-07-04
Move `startstop` package up to root of `./internal/`

brandur:
A very small change that's been on my list for a while, but I didn't do
because it'd be conflicting with some other branches I had: move the
`startstop` package up to root of `./internal/`. Until now it's been in
`./internal/maintenance` because it was originally only used for
maintenance services, but since it's now used much more widely, it makes
more sense for it to live in `./internal`.


brandur:
OMG, more `*AfterStart` flakiness like #386. Somehow wasn't running into any of those last night.

bgentry (approved):

brandur:
thx.

PR #420, opened 2024-07-04
fix: use the constructor function every time a periodic job is scheduled, instead of when its registered

elee1766:
right now, the constructor gets called when converting to an internal job, which causes the constructor to use the value at add time.

however, the comments say
```
// The constructor function is invoked each time a periodic job's schedule
// elapses, returning job arguments to insert along with optional insertion
// options.
```
so im assuming this is whats intended?


brandur:
Oops, yes, thanks! I'll try to add a test case to verify this.

brandur (approved):

PR #429, opened 2024-07-07
Move parts of the project over to `rivershared` (monorepo alternative)

brandur:
As discussed over Slack, we've been having trouble wherein we often find
ourselves needing pieces of the main River Go project over in other
projects like River UI, and having to copy a bunch of stuff over.

Here, we establish a new `rivershared` Go Module in the main River
project, then move references to it from other main modules over to it.

Basically my methodology for this one is that I'd been working on the
plugin system, wanted to see if `startstop.Service` could be moved so we
didn't have to duplicate it in `rivertype`, moved it to `rivershared`,
then moved all of its dependencies over to `rivershared` as well. (Most
of these dependencies come from `startstop`'s test suite.)


bgentry (approved):
Nice, thanks for taking care of this!

brandur:
Awesome. Alright let's give this a shot.

PR #430, opened 2024-07-07
Add basic plugin system for clients

brandur:
Here, follow up #429 to add a basic plugin system for River clients
which allows a driver to add maintenance and non-maintenance services to
a client before it starts up. The plugin interface is implemented by the
drivers themselves, and looks like this:

    type driverPlugin[TTx any] interface {
        // PluginInit initializes a plugin with an archetype and client. It's
        // invoked on Client.NewClient.
        PluginInit(archetype *baseservice.Archetype, client *Client[TTx])

        // PluginMaintenanceServices returns additional maintenance services (will
        // only run on an elected leader) for a River client.
        PluginMaintenanceServices() []startstop.Service

        // PluginServices returns additional non-maintenance services (will run on
        // all clients) for a River client.
        PluginServices() []startstop.Service
    }

The change is fairly straightforward, and we make sure to bring in some 
test cases verifying the plugin services were indeed added correctly.

### DIFF
--- a/plugin.go
+++ b/plugin.go
@@ -1,0 +1,11 @@
+package river
+
+import (
+	"log/slog"
+
+	"github.com/riverqueue/river/rivertype"
+)
+
+type Pluginable[TTx any] interface {
+	Plugin(client *Client[TTx], logger *slog.Logger) rivertype.Plugin
+}
--- a/plugin.go
+++ b/plugin.go
@@ -1,11 +1,25 @@
 package river
 
 import (
-	"log/slog"
-
-	"github.com/riverqueue/river/rivertype"
+	"github.com/riverqueue/river/rivershared/baseservice"
+	"github.com/riverqueue/river/rivershared/startstop"
 )
 
-type Pluginable[TTx any] interface {
-	Plugin(client *Client[TTx], logger *slog.Logger) rivertype.Plugin
+// A plugin API that drivers may implement to extend a River client. Driver
+// plugins may, for example, add additional maintenance services.
+//
+// This should be considered a River internal API and its stability is not
+// guaranteed. DO NOT USE.
+type driverPlugin[TTx any] interface {
+	// PluginInit initializes a plugin with an archetype and client. It's
+	// invoked on Client.NewClient.
+	PluginInit(archetype *baseservice.Archetype, client *Client[TTx])
+
+	// PluginMaintenanceServices returns additional maintenance services (will
+	// only run on an elected leader) for a River client.
+	PluginMaintenanceServices() []startstop.Service
+
+	// PluginServices returns additional non-maintenance services (will run on
+	// all clients) for a River client.
+	PluginServices() []startstop.Service
 }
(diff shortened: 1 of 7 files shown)


brandur:
Hold on review on this. Needs more work.

brandur:
@bgentry As mentioned before, I was having trouble finding a good spot to put `Plugin` because especially with the new `rivershared` module, I was getting dependency cycles.

I want to pitch a slightly alternative design here where the `Plugin` type goes away in favor of the driver itself optionally implementing plugin-like functions:

``` go
type driverPlugin[TTx any] interface {
	// PluginInit initializes a plugin with an archetype and client. It's
	// invoked on Client.NewClient.
	PluginInit(archetype *baseservice.Archetype, client *Client[TTx])

	// PluginMaintenanceServices returns additional maintenance services (will
	// only run on an elected leader) for a River client.
	PluginMaintenanceServices() []startstop.Service

	// PluginServices returns additional non-maintenance services (will run on
	// all clients) for a River client.
	PluginServices() []startstop.Service
}
```

It's a little less elegant maybe, but it has the sizable advantage that no new types are needed anywhere — a plugin can implement it by just referencing `rivershared` and `river` root. I prefixed the function names with `Prefix` so they can group nicely on a driver that's implementing them.

Thoughts?

rivershared/startstop/main_test.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"testing"
+
+	"github.com/riverqueue/river/rivershared/riversharedtest"
+)
+
+func TestMain(m *testing.M) {
+	riversharedtest.WrapTestMain(m)
+}
```
brandur, line 11: Added this because I found some goroutines leaks in the tests. Fixed those, but this will prevent future ones from occurring.

plugin.go
```diff
@@ -0,0 +1,25 @@
+package river
+
+import (
+	"github.com/riverqueue/river/rivershared/baseservice"
+	"github.com/riverqueue/river/rivershared/startstop"
+)
+
+// A plugin API that drivers may implemented to extend a River client. Driver
```
bgentry, line 8:
```suggestion
// A plugin API that drivers may implement to extend a River client. Driver
```

plugin_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (d *TestDriverWithPlugin) PluginInit(archetype *baseservice.Archetype, client *Client[pgx.Tx]) {}
+
+func (d *TestDriverWithPlugin) PluginMaintenanceServices() []startstop.Service {
+	return []startstop.Service{d.maintenanceService}
+}
+
+func (d *TestDriverWithPlugin) PluginServices() []startstop.Service {
+	return []startstop.Service{d.service}
+}
```
bgentry, line 99: Might be helpful to have the test driver panic or `t.Fatal()` if it hasn't been init'd yet when these service methods are called.

bgentry (approved):
Solid improvement on the original design. LGTM with minor comments!

plugin_test.go
brandur: Yep, good call. Added.

brandur:
Thanks!

PR #432, opened 2024-07-07
Use longer `WaitTimeout` for test manager pool acquisition context timeout

brandur:
I'm not sure what changed exactly, but about 25% of my builds have been
failing for the last couple days on the context deadline set when
acquiring a test database.

Here, attempt to resolve the problem by changing the hard-coded 5 second
wait timeout to acquire a test database manager to the timeout specified
by `WaitTimeout`, which is a tight 3 seconds locally, but which allows
more time for GitHub Actions, which seems to regularly be quite slow.

    func WaitTimeout() time.Duration {
        if os.Getenv("GITHUB_ACTIONS") == "true" {
                return 10 * time.Second
        }

        return 3 * time.Second
    }

### DIFF
--- a/internal/riverinternaltest/riverinternaltest.go
+++ b/internal/riverinternaltest/riverinternaltest.go
@@ -164,7 +164,7 @@ func DrainContinuously[T any](drainChan <-chan T) func() []T {
 func TestDB(ctx context.Context, tb testing.TB) *pgxpool.Pool {
 	tb.Helper()
 
-	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, riversharedtest.WaitTimeout())
 	defer cancel()
 
 	testPool, err := dbManager.Acquire(ctx)


brandur:
@bgentry I ran the matrix three times and can confirm that this seems to take us from the new intermittent error we're seeing back to the _old_ intermittent error we used to see (which thankfully, is less common). I don't have a clue why this is happening, but it's so frequent that I think we need something quick-ish like this to try and patch it up.

Counterintuitively, the longer wait doesn't seem to affect the timing on runs at all ... all checks finish in about two minutes (+/- 10 seconds), which is about what it was before.

internal/riverinternaltest/riverinternaltest.go
```diff
@@ -164,7 +164,7 @@ func DrainContinuously[T any](drainChan <-chan T) func() []T {
 func TestDB(ctx context.Context, tb testing.TB) *pgxpool.Pool {
 	tb.Helper()
 
-	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
```
bgentry, line 167: It's so weird that this is not long enough. Are we creating enough test DBs for the `GOMAXPROCS`? They really shouldn't be waiting at all if so, at least not longer than whatever the time is to close the pool and wait for `TruncateRiverTables` 🤔

bgentry (approved):

internal/riverinternaltest/riverinternaltest.go
brandur: Yeah it's so weird. This definitely merits deeper investigation, but ugh, don't feel like going down this rabbit hole right now. I was thinking the same around number of test DBs — seems like it's got to be something of that nature.

PR #433, opened 2024-07-07
Remove `slogtest` and `startstoptest` in main River module

brandur:
This is a follow up from #429 this morning. I moved `slogtest` and
`startstoptest` over to the new `rivershared` module, but then forgot to
remove the existing ones out of the main River module. Here, do that.
All references have already been updated.


bgentry (approved):

brandur:
ty.

PR #434, opened 2024-07-08
Include `rivershared` in `update-submodule-versions` script

brandur:
Small change to follow up #429 that makes sure that we also update
`rivershared`'s `go.mod` with changing internal package versions prior
to release.


bgentry (approved):

brandur:
Thanks!

PR #435, opened 2024-07-08
Add migration that brings `line` column to the migration table

brandur:
Here, add a new River migration (version 005) that brings a `line`
column to the `river_migration` table, allowing non-main lines to be
supported. We also teach the migrator how to use it.

This one is a little trickier than it sounds (partly thanks to sqlc)
because the migrator needs to behave somewhat differently depending on
whether the `line` column exits yet. So for example, when migrating to
004 it needs to upsert migration records that _do not_ include a `line`
field, but then when migrating to 005 or beyond, a `line` value is
included. The same consideration goes for performing a down migration,
or checking the existing state of migrations in the database.

This is implemented by pairing driver functions related to migrations
into one for pre-`line` and another for post-`line`. e.g.

    // MigrationDeleteAssumingMainMany deletes many migrations assuming
    // everything is on the main line. This is suitable for use in databases on
    // a version before the `line` column exists.
    MigrationDeleteAssumingMainMany(ctx context.Context, versions []int) ([]*Migration, error)

    // MigrationDeleteByLineAndVersionMany deletes many migration versions on a
    // particular line.
    MigrationDeleteByLineAndVersionMany(ctx context.Context, line string, versions []int) ([]*Migration, error)

We don't yet pull line support into the CLI, but this should make
alternate lines fully supported up to that point. The CLI needs a little
more thought because it might involve building a separate binary.


rivermigrate/river_migrate.go
```diff
@@ -324,9 +325,21 @@ func (m *Migrator[TTx]) migrateDown(ctx context.Context, exec riverdriver.Execut
...
-		if _, err := exec.MigrationDeleteByVersionMany(ctx, sliceutil.Map(res.Versions, migrateVersionToInt)); err != nil {
-			return nil, fmt.Errorf("error deleting migration rows for versions %+v: %w", res.Versions, err)
+	if !opts.DryRun && len(res.Versions) > 0 {
+		versions := sliceutil.Map(res.Versions, migrateVersionToInt)
+
+		// Version 005 is hard-coded here because that's the version in which
+		// the migration `line` comes in. If migration to a point equal or above
+		// 005, we can remove migrations with a line included, but otherwise we
+		// must omit the `line` column from queries because it doesn't exist.
+		if m.line == riverdriver.MigrationLineMain && slices.Min(versions) <= 5 {
```
bgentry, line 335: The value 5 here might be clearer if put in a named constant

bgentry (approved):

rivermigrate/river_migrate.go
brandur: Yep, good call. Changed.

brandur:
Thanks!

PR #438, opened 2024-07-08
Definitively fix race in periodic job enqueuer

### LINKED ISSUE #409: Intermittently failing test: `TestPeriodicJobEnqueuer/AddAfterStart`

brandur:
Related to #409. I think I finally cracked it. For real this time. No
workarounds, no hacks.

If you look at the "after start" tests, they look roughly like this,
start the client, then add some jobs:

    startService(t, svc)

    svc.Add(
        &PeriodicJob{ScheduleFunc: periodicIntervalSchedule(500 * time.Millisecond), ConstructorFunc: jobConstructorFunc("periodic_job_500ms", false)},
    )
    svc.Add(
        &PeriodicJob{ScheduleFunc: periodicIntervalSchedule(500 * time.Millisecond), ConstructorFunc: jobConstructorFunc("periodic_job_500ms_start", false), RunOnStart: true},
    )

This would usually work as expected, with the enqueuer starting up, then
jobs added, then those new jobs scheduled.

_However_, what can happen is that the "after start" jobs are actually
added in _before_ the enqueuer has finished entering its run loop. So,
the service is starting up, makes a first call to `timeUntilNextRun`,
and because the newly added periodic jobs have not been assigned a first
run time, the code below would get a `timerUntilNextRun` of zero, which
would cause it to enter the loop and run an insert iteration
immediately, despite the jobs not actually being appropriate for
scheduling as of yet:

    timerUntilNextRun := time.NewTimer(s.timeUntilNextRun())

    for {
            select {
            case <-timerUntilNextRun.C:

The fix is to modify `timeUntilNextRun` so that it ignores periodic jobs
that have not yet been scheduled:

    for _, periodicJob := range s.periodicJobs {
        // Jobs may have been added after service start, but before this
        // function runs for the first time. They're not scheduled properly yet,
        // but they will be soon, at which point this function will run again.
        // Skip them for now.
        if periodicJob.nextRunAt.IsZero() {
                continue
        }

Those jobs will be scheduled soon, then `timeUntilNextRun` called again,
and the expected "time until" values returned.

I'm confident enough in the fix that I've reverted most of #416 so the
5s jobs go back to 500ms.

Fixes #409.

### DIFF
--- a/internal/maintenance/periodic_job_enqueuer.go
+++ b/internal/maintenance/periodic_job_enqueuer.go
@@ -268,7 +268,7 @@ func (s *PeriodicJobEnqueuer) Start(ctx context.Context) error {
 
 			s.insertBatch(ctx, insertParamsMany, insertParamsUnique)
 
-			if len(insertParamsMany) > 0 {
+			if len(insertParamsMany) > 0 || len(insertParamsUnique) > 0 {
 				s.Logger.DebugContext(ctx, s.Name+": Inserted RunOnStart jobs", "num_jobs", len(insertParamsMany)+len(insertParamsUnique))
 			}
 		}
@@ -300,7 +300,7 @@ func (s *PeriodicJobEnqueuer) Start(ctx context.Context) error {
 					defer s.mu.RUnlock()
 
 					for _, periodicJob := range s.periodicJobs {
-						if !periodicJob.nextRunAt.Before(nowWithMargin) {
+						if periodicJob.nextRunAt.IsZero() || !periodicJob.nextRunAt.Before(nowWithMargin) {
 							continue
 						}
 
@@ -405,6 +405,7 @@ func (s *PeriodicJobEnqueuer) insertBatch(ctx context.Context, insertParamsMany 
 		s.Logger.ErrorContext(ctx, s.Name+": Error committing transaction", "error", err.Error())
 		return
 	}
+
 	s.TestSignals.InsertedJobs.Signal(struct{}{})
 }
 
@@ -427,6 +428,8 @@ func (s *PeriodicJobEnqueuer) insertParamsFromConstructor(ctx context.Context, c
 	return insertParams, uniqueOpts, true
 }
 
+const periodicJobEnqueuerVeryLongDuration = 24 * time.Hour
+
 func (s *PeriodicJobEnqueuer) timeUntilNextRun() time.Duration {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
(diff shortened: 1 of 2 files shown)


client_test.go
```diff
@@ -3956,7 +3956,7 @@ func Test_Client_JobCompletion(t *testing.T) {
 		require.NotNil(updatedJob.FinalizedAt)
 
 		// Make sure the FinalizedAt is approximately ~now:
-		require.WithinDuration(now, *updatedJob.FinalizedAt, time.Second)
+		require.WithinDuration(now, *updatedJob.FinalizedAt, 2*time.Second)
```
brandur, line 3959: Totally unrelated, but one that cropped up as an intermittent failure as I reran this PR's test matrix over and over again. Most duration checks use 2s rather than 1s by convention in this file, so this change makes it match up a little better with other assertions.

bgentry (approved):
As much as I'm in agreement with GitHub Actions' inconsistent performance, this is bad enough that I'm really becoming convinced there must be something else going on here. This is just an eternity in computer time.

brandur:
Thx. Ugh, yeah you're right that there's definitely something else at work here. I'm going to leave this open and convert it to a draft to help me debug further.

brandur:
@bgentry I think I FINALLY cracked this one. It was a legit bug, although not one that would ever occur IRL probably. I did a complete rewrite of the PR description to describe it. Mind taking another look?

internal/maintenance/periodic_job_enqueuer.go
```diff
@@ -427,14 +428,16 @@ func (s *PeriodicJobEnqueuer) insertParamsFromConstructor(ctx context.Context, c
 	return insertParams, uniqueOpts, true
 }
 
+const periodicJobEnqueurVeryLongDuration = 24 * time.Hour
```
bgentry, line 431: Small typo in the name of this long constant:

```suggestion
const periodicJobEnqueuerVeryLongDuration = 24 * time.Hour
```

brandur: Oops, thanks. Fixed.

PR #439, opened 2024-07-09
Bump `WithinDuration` check up to two seconds

brandur:
This fixes an intermittent failure. It's a fairly rare one, but one that
does appear on the order of ~dozens of test runs in GitHub Actions. I'd
originally lumped this into #438, but since I'm likely to discard that
pull request, breaking this change out separately since it's still
useful.

Most of the other time comparisons in this file already use two seconds
for `WithinDuration` instead of one, so the change also serves to make
things a little more consistent.


brandur:
ty!

PR #440, opened 2024-07-09
Use `TestOnly` property in example tests

brandur:
I was looking into what was making the example tests so slow (some like
the periodic job enqueuer one often take 3 seconds to run), and it turns
out that the lion's share of it is from from service start up jitter.

Example tests aren't in the main `river` package (they're in
`river_test`) so they didn't previously have a way of interacting with
the client in such a way as to disable stagger. Luckily, we do have a
way now. Take advantage of the recently introduced `TestOnly` client
property from #414 to make all the examples faster.

### DIFF
--- a/example_batch_insert_test.go
+++ b/example_batch_insert_test.go
@@ -52,7 +52,8 @@ func Example_batchInsert() {
 		Queues: map[string]river.QueueConfig{
 			river.QueueDefault: {MaxWorkers: 100},
 		},
-		Workers: workers,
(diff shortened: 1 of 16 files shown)


bgentry (approved):
Awesome! Do you think we should add the flag to higher level docs as well? https://riverqueue.com/docs/testing

brandur:
Oh yeah, good idea. I'll take that a bit later.

PR #443, opened 2024-07-11
Tighten up schema of the migrations table a little more

brandur:
I'd totally forgotten I'd even created it this way, but going back
through migrations recently I realized that the table's primary key is a
`bigserial`.

This isn't totally bad, and even has some advantages in that rows are
easier to target from a psql session, but in this instance it feels a
little on the sloppy side because it's fairly easy to conflate the
serial with migration `version` -- both are expected to be small
integers that a human could easily mix up.

Since we haven't shipped a new version yet containing the addition of
`line` in #435, here I propose that we modify that a little further to
reshape `river_migration` into a slightly smaller table with a compound
key on `(line, version)`:

    CREATE TABLE river_migration(
        created_at timestamptz NOT NULL DEFAULT NOW(),
        line TEXT NOT NULL,
        version bigint NOT NULL,
        CONSTRAINT line_length CHECK (char_length(line) > 0 AND char_length(line) < 128),
        CONSTRAINT version_gte_1 CHECK (version >= 1),
        PRIMARY KEY (line, version)
    );

This has two tiny side benefits that (1) the table is smaller (drops a
column), and (2) we don't need an extra index on `(line, version)` since
it's already covered by the primary key. Very similar in spirit to what
it was before, but I think a little cleaner in design.

### DIFF
--- a/riverdriver/riverdatabasesql/migration/main/005_river_migration_add_line.down.sql
+++ b/riverdriver/riverdatabasesql/migration/main/005_river_migration_add_line.down.sql
@@ -1,6 +1,36 @@
-DROP INDEX river_migration_line_version_idx;
-CREATE UNIQUE INDEX river_migration_version_idx ON river_migration USING btree(version);
+--
+-- If any non-main migration are present, 005 is considered irreversible.
+--
+
+DO
+$body$
+BEGIN
+    IF EXISTS (
+        SELECT *
+        FROM river_migration
+        WHERE line <> 'main'
+    ) THEN
+        RAISE EXCEPTION 'Found non-main migration lines in the database; version 005 migration is irreversible because it would result in loss of migration information.';
+    END IF;
+END;
+$body$
+LANGUAGE 'plpgsql'; 
 
 ALTER TABLE river_migration
-    DROP CONSTRAINT line_length,
-    DROP COLUMN line;
+    RENAME TO river_migration_old;
+
+CREATE TABLE river_migration(
+  id bigserial PRIMARY KEY,
+  created_at timestamptz NOT NULL DEFAULT NOW(),
+  version bigint NOT NULL,
+  CONSTRAINT version CHECK (version >= 1)
+);
+
+CREATE UNIQUE INDEX ON river_migration USING btree(version);
+
+INSERT INTO river_migration
+    (created_at, version)
+SELECT created_at, version
+FROM river_migration_old;
+
+DROP TABLE river_migration_old;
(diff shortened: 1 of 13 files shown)


brandur:
@bgentry Not high priority, but this design feels a bit cleaner to me. Thoughts?

riverdriver/riverdatabasesql/migration/main/005_river_migration_add_line.down.sql
```diff
@@ -1,6 +1,22 @@
-DROP INDEX river_migration_line_version_idx;
-CREATE UNIQUE INDEX river_migration_version_idx ON river_migration USING btree(version);
+--
+-- If any non-main migration are present, 005 should be considered irreversible.
+--
```
bgentry, line 3: Any way we could make this intentionally error? Thinking of something like selecting from a nonexistent column `irreversible_migration` if any values are in the set of `SELECT FROM river_migration WHERE line != 'main'` or something

riverdriver/riverdatabasesql/migration/main/005_river_migration_add_line.up.sql
```diff
@@ -1,12 +1,18 @@
 ALTER TABLE river_migration
-    ADD COLUMN line text;
+    RENAME TO river_migration_old;
 
-UPDATE river_migration
-SET line = 'main';
+CREATE TABLE river_migration(
+    created_at timestamptz NOT NULL DEFAULT NOW(),
+    line TEXT NOT NULL,
+    version bigint NOT NULL,
```
bgentry, line 7: imo it's a little funky to not put the PK fields first just to make the default `SELECT * FROM river_migration` a bit cleaner. Seems like a warranted variance from the typical desire to alphabetize.

bgentry (approved):
Looks great in principle despite my minor nits :shipit:

riverdriver/riverdatabasesql/migration/main/005_river_migration_add_line.down.sql
brandur: Heh, yeah I thought about that too, but never willingly try to remember how to write PL/pgSQL ;)

Added.

riverdriver/riverdatabasesql/migration/main/005_river_migration_add_line.up.sql
brandur: Fair enough. Changed.

brandur:
thx.

PR #444, opened 2024-07-11
Clean up extraneous SQL annotations

brandur:
A small change that goes through and cleans up SQL a bit by removing
some unnecessary annotations, mostly typecasting that's already implied
by the type of the column where the values are going.


bgentry (approved):

brandur:
Thanks!

PR #447, opened 2024-07-12
Add River Python link from README + `doc.go`

brandur:
The README links out to River Ruby already. Here, add a link to Stripe
Python as well. Also, add this section to `doc.go` (where it'd
previously been forgotten) to keep these two docs in sync.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -151,6 +151,13 @@ See the [`InsertAndWork` example] for complete code.
 
   - [Work functions] for simplified worker implementation.
 
+## Cross language enqueueing
+
+River supports inserting jobs in some non-Go languages which are then worked by Go implementations. This may be desirable in performance sensitive cases so that jobs can take advantage of Go's fast runtime.
+
+  - [Inserting jobs from Python].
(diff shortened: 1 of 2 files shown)


doc.go
```diff
@@ -151,6 +151,13 @@ See the [`InsertAndWork` example] for complete code.
 
   - [Work functions] for simplified worker implementation.
 
+## Cross language enqueueing
+
+River supports inserting jobs in some non-Go languages which are then be worked by Go implementations. This may be desirable in performance sensitive cases so that jobs can take advantage of Go's fast runtime.
```
bgentry, line 156:
```suggestion
River supports inserting jobs in some non-Go languages which are then worked by Go implementations. This may be desirable in performance sensitive cases so that jobs can take advantage of Go's fast runtime.
```

bgentry: Same typo is in the file below too

bgentry (approved):
LGTM except for typos

doc.go
brandur: Thx. Fixed the other instance too.

PR #450, opened 2024-07-13
Makefile `make db/reset` target + rename everything to use `river_test` convention

brandur:
A couple small ones:

* I often find myself wanting to reset my dev database from scratch, and
  it's quite inconvenient to do because you want to use the current code
  (rather than installed River CLI), and doing so requires cding into
  the `./cmd/river` directory because `go run` won't let you run a
  directory that's not in the current project. Here, add a `make
  db/reset` that'll easily reset the dev database, test databases, or
  both, all at once.

* While we're at it, change over to the naming convention of
  `river_test` established in the River Ruby and River Python projects.


brandur:
ty.

PR #451, opened 2024-07-13
Substantially (~20-45x) faster unique insertion using unique index

brandur:
Here, rebuild the unique job insertion infrastructure so that insertions
become substantially faster, in the range of 20 to 45x.

    $ go test -bench=. ./internal/dbunique
    goos: darwin
    goarch: arm64
    pkg: github.com/riverqueue/river/internal/dbunique
    BenchmarkUniqueInserter/FastPathEmptyDatabase-8                     9632            126446 ns/op
    BenchmarkUniqueInserter/FastPathManyExistingJobs-8                  9718            127795 ns/op
    BenchmarkUniqueInserter/SlowPathEmptyDatabase-8                      468           3008752 ns/op
    BenchmarkUniqueInserter/SlowPathManyExistingJobs-8                   214           6197776 ns/op
    PASS
    ok      github.com/riverqueue/river/internal/dbunique   13.558s

The speed up is accomplished by mostly abandoning the old methodology
that took an advisory lock, did a job look up, and then did an insertion
if no equivalent unique job was found. Instead, we add a new
`unique_key` field to the jobs table, put a partial index on it, and
use it in conjunction with `kind` to do upserts for unique insertions.
Its value is similar to what we used for advisory locks -- a hash of a
string representing the unique opts in question.

There is however, a downside. `unique_key` is easy when all we need to
think about are uniqueness based on something immutable like arguments
or queue, but more difficult when we have to factor in job state, which
may change over the lifetime of a job.

To compensate for this, we clear `unique_key` on a job when setting it
to states not included in the default unique state list, like when it's
being cancelled or discarded. This allows a new job with the same unique
properties to be inserted again.

But the corollary of this technique is that if a state like `cancelled`
or `discarded` is included in the `ByState` property, the technique
obviously doesn't work anymore. So instead, in these cases we _keep_ the
old insertion technique involving advisory locks, and fall back to this
slower insertion path when we have to. So while we get the benefits of
substantial performance improvements, we have the downside of more
complex code -- there's now two paths to think about and which have to
be tested. Overall though, I think the benefit is worth it.

The addition does require a new index. Luckily it's a partial so it only
gets used on unique inserts, and I benchmarked before/after, and found
no degradation in non-unique insert performance. I added instructions to
the CHANGELOG for building the index with `CONCURRENTLY` for any users
who may already have a large jobs table, giving them an operationally
safer alternative to use.


cmd/river/go.mod
```diff
@@ -10,6 +10,8 @@ replace github.com/riverqueue/river/riverdriver/riverdatabasesql => ../../riverd
 
 replace github.com/riverqueue/river/riverdriver/riverpgxv5 => ../../riverdriver/riverpgxv5
 
+replace github.com/riverqueue/river/rivertype => ../../rivertype
```
brandur, line 13: Need to remember to re-comment all these `replaces` in the CLI module before final release.

riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
```diff
@@ -276,9 +278,43 @@ INSERT INTO river_job(
...
+    @queue,
+    coalesce(sqlc.narg('scheduled_at')::timestamptz, now()),
+    @state,
+    coalesce(@tags::varchar(255)[], '{}'),
+    @unique_key
+)
+ON CONFLICT (kind, unique_key) WHERE unique_key IS NOT NULL
+    -- Something needs to be updated for a row to be returned on a conflict.
+    DO UPDATE SET kind = EXCLUDED.kind
+RETURNING *, (xmax != 0) AS unique_skipped_as_duplicate;
```
brandur, line 316: This `(max != 0)` thing is kind of a well known upsert trick for figuring out whether the operation resulted in an insert or an update. It's not formally documented, but it's worked since the introduction of upsert, and general consensus is that it's unlikely to be broken at this point. Unfortunately, upsert provides no formal way of differentiating an insert versus an update.

riverdriver/riverpgxv5/river_pgx_v5_driver.go
```diff
@@ -272,13 +272,61 @@ func (e *Executor) JobInsertFull(ctx context.Context, params *riverdriver.JobIns
...
+		Kind:        insertRes.Kind,
+		MaxAttempts: insertRes.MaxAttempts,
+		Metadata:    insertRes.Metadata,
+		Priority:    insertRes.Priority,
+		Queue:       insertRes.Queue,
+		ScheduledAt: insertRes.ScheduledAt,
+		State:       insertRes.State,
+		Tags:        insertRes.Tags,
+		UniqueKey:   insertRes.UniqueKey,
+	})
```
brandur, line 319: This is a little unfortunate, but the addition of `(xmax != 0)` to `RETURNING *` means that sqlc won't return this as a `RiverJob` anymore, so we have to map it back to one as seen here. It's not great, but I copied over all the property tests from other insert functions so can verify we're not missing anything.

riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
```diff
@@ -276,9 +278,43 @@ INSERT INTO river_job(
     @queue,
     coalesce(sqlc.narg('scheduled_at')::timestamptz, now()),
     @state,
-    coalesce(@tags::varchar(255)[], '{}')
+    coalesce(@tags::varchar(255)[], '{}'),
+    @unique_key
 ) RETURNING *;
 
+-- name: JobInsertUnique :one
```
brandur, line 285: This op is ~the same as `JobInsertFast` except for the `ON CONFLICT`.

We may want to consider just rollling the two up together so that every job insert is just an upsert to reduce on the number of separate ops in here we have.

I left them separate for now because I wrote a benchmark and the upsert version regularly seems to come out as a couple percent slower:

```
$ go test -bench=BenchmarkDriverRiverPgxV5Insert .
goos: darwin
goarch: arm64
pkg: github.com/riverqueue/river
BenchmarkDriverRiverPgxV5Insert/InsertFast-8               12199             97738 ns/op
BenchmarkDriverRiverPgxV5Insert/InsertUnique-8             10000            100346 ns/op
PASS
ok      github.com/riverqueue/river     15.302s
```

It's not that big of a difference though, so an argument could be made that long term the extra operations aren't worth it.

brandur: Actually, those bench results show quite a large different. But depending on the run, some are much closer:

```
$ go test -bench=BenchmarkDriverRiverPgxV5Insert .
goos: darwin
goarch: arm64
pkg: github.com/riverqueue/river
BenchmarkDriverRiverPgxV5Insert/InsertFast-8               12384            102295 ns/op
BenchmarkDriverRiverPgxV5Insert/InsertUnique-8             12135            112301 ns/op
PASS
ok      github.com/riverqueue/river     18.364s
```

Most runs tend to be a closer to the first than the second, so probably good reason to keep them as separate ops for now.

riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
bgentry: Oh wild, nice trick!

riverdriver/riverpgxv5/river_pgx_v5_driver.go
bgentry: Yeah, I always hate this effect when I want to select one or two little things in addition to the entire row. I sorta wish sqlc adopted an embedded struct for this situation instead of inventing all new structs with a few extra fields.

bgentry (commented):
Obviously I love the perf gains, but I'm a little uncertain whether this approach is fully fleshed out. For example, what happens when jobs change states by another mechanism besides the completer (such as the rescuer, or a job cancellation update)? Won't they potentially be left in an inconsistent state with a unique key left in place potentially incorrectly? And what about when a job is manually retried, what happens to the unique key there?

CHANGELOG.md
```diff
@@ -14,6 +14,20 @@ go install github.com/riverqueue/river/cmd/river@latest
...
 ```
 
+The migration **includes a new index**. Users with a very large job table may want to consider raising the index separately using `CONCURRENTLY` (which must be run outside of a transaction), then run `river migrate-up` to finalize the process (it will tolerate an index that already exists):
+
+```sql
+ALTER TABLE river_job
+    ADD COLUMN unique_key bytea;
+
+CREATE UNIQUE INDEX CONCURRENTLY river_job_kind_unique_key_idx ON river_job (kind, unique_key) WHERE unique_key IS NOT NULL;
+```
```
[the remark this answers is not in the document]
brandur: Dang it, I'm not sure we have a choice except this for the time being. The migration framework automatically uses transactions so that in case a statement within a migration fails, the whole thing can be rolled back without tainting the database.

We'd have to add some new kind of "no transaction" mode to the migrator. Definitely possible, but would take a little more work right now.

riverdriver/riverpgxv5/river_pgx_v5_driver.go
brandur: Yeah, found myself wishing that _exactly_ thing. An option for an embedding struct when `*` was combined with something else would solve this problem completely.

brandur:
> Obviously I love the perf gains, but I'm a little uncertain whether this approach is fully fleshed out. For example, what happens when jobs change states by another mechanism besides the completer (such as the rescuer, or a job cancellation update)? Won't they potentially be left in an inconsistent state with a unique key left in place potentially incorrectly? And what about when a job is manually retried, what happens to the unique key there?

Cool, I think we covered most of this sync, but LMK if you still have concerns about it.

I've since updated the job cancellation paths. This should be ready for review now.

CHANGELOG.md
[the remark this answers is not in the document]
irreal: Might be a silly idea, but I felt the need to throw it out there:

inside the migration you could detect if the index was not created manually and the jobs table is above a certain (unfortunately arbitrary) threshold of rows, and fail the migration with an error indicating that the index should be created concurrently manually?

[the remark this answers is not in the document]
irreal: @brandur btw, creating a new type of migration job that runs without a tx would also not help with users who export sql to run in their own migration frameworks.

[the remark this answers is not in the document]
brandur: Yeah that's a good thought.

It's still a little difficult because given a very large table, even doing a `SELECT count(*)` can sometimes be a bit of a debacle. Sill though, cheaper than building an index.

brandur:
Actually, one tweak I made here is to switch from SHA-3 to SHA-2 because SHA-2 has the nice property of being available in the stdlibs of all of Go, Ruby, and Python, and since we're not using this for security or anything, we don't need the exact state of the art in cryptographic robustness. Even in Go, SHA-3 is still in the /x/ tree, so this is better. I was using the 256-bit variant either way, so plenty of hash space to work with.

bgentry (approved):

brandur:
Thx.

riverdriver/riverpgxv5/river_pgx_v5_driver.go
tadejsv: Hey @brandur, I think this is already possible when using `sqlc.embed`, check out this example: https://play.sqlc.dev/p/fa54ae82af605631d9ed837d0d71b0d661876e891217ff0c35903888f4d26df1

brandur: Oh, amazing! Had no idea this feature existed.

Totally works. Opened #464. Thanks for pointing it out!

PR #452, opened 2024-07-13
Demote many log lines from info to debug

brandur:
We've had reports that a running River client can be quite a noisy
affair, producing too much logging in general.

Here, we move many log lines from info to debug. I suspect the most
impactful is that we had a bunch in the elector and notifier that would
fire quite often during run of the mill operation.

I left a couple in place, although I'm not 100% sure it's the right
thing:

* The 5 second heartbeat loop in the client that prints job statistics.
  I figure that if this is the _main_ log line being emitted, with
  others being quite rare, it might be okay. We could also consider
  moving it to be more infrequent like every 10 seconds.

* I left the maintenance service "done run" log lines as info for the
  time being, but I changed them so that they only print in case there
  were >0 results, so I think the result is that most of them are silent
  most of the time, except perhaps in an actual production situation.

I figure we can try this as a pass, see how it is, and then possibly
demote more if it's still annoying.

### DIFF
--- a/internal/maintenance/job_cleaner.go
+++ b/internal/maintenance/job_cleaner.go
@@ -124,9 +124,11 @@ func (s *JobCleaner) Start(ctx context.Context) error { //nolint:dupl
 				continue
 			}
 
-			s.Logger.InfoContext(ctx, s.Name+logPrefixRanSuccessfully,
-				slog.Int("num_jobs_deleted", res.NumJobsDeleted),
-			)
+			if res.NumJobsDeleted > 0 {
+				s.Logger.InfoContext(ctx, s.Name+logPrefixRanSuccessfully,
+					slog.Int("num_jobs_deleted", res.NumJobsDeleted),
+				)
+			}
 		}
 	}()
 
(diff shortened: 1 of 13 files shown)


internal/maintenance/job_cleaner.go
```diff
@@ -124,9 +124,11 @@ func (s *JobCleaner) Start(ctx context.Context) error { //nolint:dupl
...
 			}
 
-			s.Logger.InfoContext(ctx, s.Name+logPrefixRanSuccessfully,
-				slog.Int("num_jobs_deleted", res.NumJobsDeleted),
-			)
+			if res.NumJobsDeleted > 0 {
+				s.Logger.DebugContext(ctx, s.Name+logPrefixRanSuccessfully,
+					slog.Int("num_jobs_deleted", res.NumJobsDeleted),
+				)
+			}
```
brandur, line 131: There's some argument to be made that these changes are detrimental to debugging tests, but I generally find that if I'm debugging anything non-trivial I need to add more log statements anyway because the existing ones just aren't enough.

internal/maintenance/job_rescuer.go
```diff
@@ -232,11 +234,6 @@ func (s *JobRescuer) runOnce(ctx context.Context) (*rescuerRunOnceResult, error)
 			break
 		}
 
-		s.Logger.InfoContext(ctx, s.Name+": Rescued batch of jobs",
-			slog.Int64("num_jobs_discarded", res.NumJobsDiscarded),
-			slog.Int64("num_jobs_retried", res.NumJobsRetried),
-		)
```
brandur, line 238: A couple of these services sort of had doubled up run lines. I took one half out and just left the line above.

bgentry (approved):
:shipit:  🤫

brandur:
Thanks!

PR #454, opened 2024-07-14
Break out CLI functionality to make it modular + `--line` flag

brandur:
Here, break out the CLI so that most of its implementation no longer
lives in a `main.go`, and rather in a subpackage that can be imported
from somewhere else. This lets the majority of it to be reused in
another executable and augmented in various ways.

An addition is a new `DriverProcurer` interface that can provide a
driver for various databases:

    type DriverProcurer interface {
        ProcurePgxV5(pool *pgxpool.Pool) riverdriver.Driver[pgx.Tx]
    }

For the main CLI, this gets a trivial implementation using `riverpgxv5`,
but could potentially be reimplemented elsewhere to swap in something
else;

    type DriverProcurer struct{}

    func (p *DriverProcurer) ProcurePgxV5(pool *pgxpool.Pool) riverdriver.Driver[pgx.Tx] {
            return riverpgxv5.New(pool)
    }

To make this more workable, I end up going through and refactoring quite
a lot of code, making it into a bit of a mini framework, and one that
could potentially support additional databases in the future without
having to refactor the world again.

We also add a `--line` flag to support the feature from #435 to the CLI.


bgentry (approved):

brandur:
Awesome. Ty.

PR #456, opened 2024-07-15
Add `river_client` table for tracking state of active client

brandur:
As discussed yesterday, I wanted to quickly pitch the schema for a new
`river_client` table, whose job it would be to track active clients and
how much worked they're performing or have performed. The idea would be
to include the new table in the v0.10.0 migration, then implement the
feature at our leisure after we've knocked out higher priority work, but
avoiding another migration while doing so.

I've largely based the design off the existing `river_queue` table,
including a `paused_at` column that we could use to implement per-client
pause/resume (although I think just having stats information visible
from the UI would be very useful and quite easy to do, so that might be
a nice first pass).

One difference is that I'm proposing a `river_client_queue` table
derivative as well because after thinking about it some, knowing only
the aggregate stats for a client isn't all that useful. For example, say
you have a `high_priority` queue/producer that's operating at 0 / 100
capacity, but then a `default` queue/producer that's maxed out at 100 /
100. Showing that the client's capacity is 100 / 200 would be borderline
useless information -- it's only sensible if you're able to show per
queue detail.

We could jam queue info into a `jsonb` on `river_client` and that might
work okay for Go code, but I tend to dislike this make it decreases the
schema's legibility (may as well use Mongo), and makes the table harder
for an operator to query from something like a `psql` shell (they have
to do digging for JSON functions, `\d+` means nothing, and tab complete
goes away).


brandur:
@bgentry Just wanted to get your rough thoughts on this schema.

riverdriver/riverpgxv5/migration/main/005_river_migration_add_line.up.sql
```diff
@@ -15,4 +15,35 @@ INSERT INTO river_migration
...
+
+-- Differs from `river_queue` in that it tracks the queue state for a particular
+-- active client.
+CREATE UNLOGGED TABLE river_client_queue (
+    river_client_name text NOT NULL REFERENCES river_client (name) ON DELETE CASCADE,
+    name text NOT NULL,
+    created_at timestamptz NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    max_workers bigint NOT NULL DEFAULT 0,
+    num_jobs_completed bigint NOT NULL DEFAULT 0,
+    num_jobs_running bigint NOT NULL DEFAULT 0,
```
bgentry, line 45: While these are named very similarly, they seem to have very different purposes (if I'm understanding it correctly). One is a historical aggregate stat of how many jobs the client has worked on that queue, while the other is showing a real-time count of in-progress jobs. Almost wondering if we should name it more similarly to `max_workers`.

Minor though, I don't feel strongly about it.

bgentry, line 45: Do you think any of these should have a `> 0` constraint?

riverdriver/riverpgxv5/migration/main/005_river_migration_add_line.up.sql
```diff
@@ -15,4 +15,35 @@ INSERT INTO river_migration
...
+    updated_at timestamptz NOT NULL,
+    CONSTRAINT name_length CHECK (char_length(name) > 0 AND char_length(name) < 128)
+);
+
+-- Differs from `river_queue` in that it tracks the queue state for a particular
+-- active client.
+CREATE UNLOGGED TABLE river_client_queue (
+    river_client_name text NOT NULL REFERENCES river_client (name) ON DELETE CASCADE,
+    name text NOT NULL,
+    created_at timestamptz NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
bgentry, line 42: huh, I've always used and seen `DEFAULT now()` and never seen this form. Maybe we want to stick with `now()` for consistency within the project since we use it elsewhere?

riverdriver/riverpgxv5/migration/main/005_river_migration_add_line.up.sql
```diff
@@ -15,4 +15,35 @@ INSERT INTO river_migration
...
+--
+-- Create `river_client` and derivative.
+--
+-- This feature hasn't quite yet been implemented, but we're taking advantage of
+-- the migration to add the schema early so that we can add it later without an
+-- additional migration.
+--
+
+CREATE UNLOGGED TABLE river_client (
+    name text PRIMARY KEY NOT NULL,
```
bgentry, line 29: notably this creates a hard constraint on unique client names, which may be breaking when rolled out

bgentry (approved):
Mostly LGTM with a few questions & comments :v:

riverdriver/riverpgxv5/migration/main/005_river_migration_add_line.up.sql
brandur: Yeah, I'll change them back to `now()`.

One thing I learned while trying to implement SQLite is that `now()` is actually a Postgres thing while `CURRENT_TIMESTAMP` is a standard that works everywhere. It's not a problem for now, but if we bring SQLite in we may want to switch to `CURRENT_TIMESTAMP` everywhere for consistency since SQLite will need it that way anyway.

riverdriver/riverpgxv5/migration/main/005_river_migration_add_line.up.sql
brandur: WFM. I'll add it.

riverdriver/riverpgxv5/migration/main/005_river_migration_add_line.up.sql
brandur: Hmm, so default names should always produce relatively unique names. You're thinking if someone overrides `ID` and names everything the same thing?

I think probably what would happen in that case rather than throwing an error would be that the stats produced would end up being nonsense since the clients would be overriding each other's work. Not great, but at least it's not too harmful beyond bad numbers.

riverdriver/riverpgxv5/migration/main/005_river_migration_add_line.up.sql
brandur: Hmm, I kind of feel like it's pretty descriptive and I'm having a hard time thinking of a better name. Did you have one in mind?

riverdriver/riverpgxv5/migration/main/005_river_migration_add_line.up.sql
bgentry: Yeah, specifically for overriding. Like what if somebody used Heroku style `web.1` names instead of the unique dyno ID. That may interfere with leader election anyway. Just wanted to be mindful of this as we implement it, it's probably fine for a stats table as long as it doesn't become an extra breaking point.

brandur: I guess we could do something like try to detect that a newly started client updates a record that was inserted recently instead of inserting a fresh one, which might suggest a duplicate ID, but even that could only be a warning at best because it could be legitimate restart that just happened to have occurred quite quickly since the last one.

But yeah, I'm not too compelled to fix a problem that'd be entirely user-induced. There's only so much you can do to protect people from themselves, and like you point out, leader election won't work either.

bgentry (approved):

brandur:
Thanks! I also added one more note on `Config.ID` that duplicate IDs will cause problems and if in doubt, to leave the property empty.

PR #458, opened 2024-07-16
Show names in migration output + trim duration precision

brandur:
Currently, running a migration via the CLI produces output like this:

    applied migration 001 [up] [7.855584ms]
    applied migration 002 [up] [28.0565ms]
    applied migration 003 [up] [9.860208ms]
    applied migration 004 [up] [10.549625ms]
    applied migration 005 [up] [10.458333ms]

It's not the worst, but notice a few things:

* Only a migration's version number is included, not the human-friendly
  name from its filename.

* We use Go's default `time.Duration` output, so all the durations look
  ugly and have way more significant digits than is necessary.

Here, we make a few tweaks to fix those problems. New output looks like:

    applied migration 001 [up] create river migration   [4.74ms]
    applied migration 002 [up] initial schema           [18.61ms]
    applied migration 003 [up] river job tags non null  [522.12µs]
    applied migration 004 [up] pending and more         [3.45ms]
    applied migration 005 [up] river migration add line [1.43ms]

### DIFF
--- a/cmd/river/rivercli/river_cli.go
+++ b/cmd/river/rivercli/river_cli.go
@@ -329,17 +329,33 @@ func (c *migrateDown) Run(ctx context.Context, opts *migrateOpts) (bool, error) 
 	return true, nil
 }
 
+// Rounds a duration so that it doesn't show so much cluttered and not useful
+// precision in printf output.
+func roundDuration(duration time.Duration) time.Duration {
+	switch {
+	case duration > 1*time.Second:
+		return duration.Truncate(10 * time.Millisecond)
+	case duration < 1*time.Millisecond:
+		return duration.Truncate(10 * time.Nanosecond)
+	default:
+		return duration.Truncate(10 * time.Microsecond)
+	}
+}
+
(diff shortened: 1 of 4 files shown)


brandur:
@bgentry Just a small one that I was finding kind of annoying.

cmd/river/main.go
```diff
@@ -359,17 +359,33 @@ func migrateDown(ctx context.Context, logger *slog.Logger, out io.Writer, opts *
...
+func roundDuration(duration time.Duration) time.Duration {
+	switch {
+	case duration > 1*time.Second:
+		return duration.Truncate(10 * time.Millisecond)
+	case duration < 1*time.Millisecond:
+		return duration.Truncate(10 * time.Nanosecond)
+	default:
+		return duration.Truncate(10 * time.Microsecond)
+	}
+}
```
bgentry, line 373: hah, thanks, I noticed this the other day and I'm glad you did too took the initiative to fix it!

bgentry (approved):

cmd/river/main.go
brandur: NP! Yeah, surprised I didn't do it before, but I guess it was just pure shipping mode.

PR #460, opened 2024-07-19
Added an argument type for river.AddWorker

vacoo:
Снимок экрана 2024-07-19 в 10 25 01

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -53,7 +53,7 @@ start up so that River knows how to assign jobs to workers:
 ```go
 workers := river.NewWorkers()
 // AddWorker panics if the worker is already registered or invalid:
-river.AddWorker(workers, &SortWorker{})
+river.AddWorker[SortArgs](workers, &SortWorker{})
 ```
 
 ## Starting a client


brandur:
@vacoo That type parameter will be inferred from the compiler. You can see working code here that's confirmed to be correct by the test suite:

https://github.com/riverqueue/river/blob/master/example_insert_and_work_test.go#L51

It looks like you've got a different problem somewhere. Your tooltip seems to be saying about `NextRetry`. Do you want to see if something weird is going on there?

vacoo:
> @vacoo That type parameter will be inferred from the compiler. You can see working code here that's confirmed to be correct by the test suite:
> 
> https://github.com/riverqueue/river/blob/master/example_insert_and_work_test.go#L51
> 
> It looks like you've got a different problem somewhere. Your tooltip seems to be saying about `NextRetry`. Do you want to see if something weird is going on there?

The NextRetry method is already available in river.WorkerDefaults.

brandur:
> The NextRetry method is already available in river.WorkerDefaults.

Yep, I know. Are you sure you don't have any additional customization?

vacoo:
> > The NextRetry method is already available in river.WorkerDefaults.
> 
> Yep, I know. Are you sure you don't have any additional customization?

I don't think so. I use go land and go1.22.5 darwin/arm64
Снимок экрана 2024-07-19 в 10 45 32

brandur:
Hm, maybe check whether the code compiles outside Goland with `go build .`?

If that doesn't work, then maybe you could try copying the code from the example test just in case there's a typo or something somewhere:

https://github.com/riverqueue/river/blob/master/example_insert_and_work_test.go

vacoo:
> Hm, maybe check whether the code compiles outside Goland with `go build .`?
> 
> If that doesn't work, then maybe you could try copying the code from the example test just in case there's a typo or something somewhere:
> 
> https://github.com/riverqueue/river/blob/master/example_insert_and_work_test.go

Strangely, everything works. It just seems to be just a warning

brandur:
Ah weird. Possibly a GoLand issue somewhere ...

PR #461, opened 2024-07-19
Add `pending` job state to default unique job states

brandur:
Here, add the `pending` job state to the set of default unique states.
`pending` was a relatively recent addition in #301 and probably should 
always have been in the default unique states, but `pending` isn't used yet
so no one noticed one way or the other.


brandur:
@bgentry There's a test case in `dbunique` that iterates through the default unique states and makes sure they're included to return the same job. I added `pending` to that and I think that's sufficient test coverage for this.

bgentry (approved):

brandur:
Thanks!

PR #462, opened 2024-07-19
Prepare version v0.10.0

brandur:
Prepare version v0.10.0 which contains a broad bucket of changes
including a new migration. See the changelog for details.

Notably, this release will require a separate CLI release step, which
will follow this one.


brandur:
Ugh, build is failing because the CLI can't find the v0.10.0 release tag for `rivershared`, which I think it's picking up transitively. We might just have to push through this failure and we should see green as I cut a CLI release.

PR #464, opened 2024-07-20
Use `sqlc.embed` to embed a `RiverJob` row during unique insertion

brandur:
Follows up #451 with a significant code quality improvement in which we
can use `sqlc.embed` (which TIL about) to embed a `RiverJob` row
directly on the returned struct, which means we can use our normal
`jobRowFromInternal` to map it to a driver result instead of having to
manually construct `RiverJob` property by property.

Tipped off to the existence of `sqlc.embed` by @tadejsv [1]. Thank you!

[1] https://github.com/riverqueue/river/pull/451#discussion_r1684925329


bgentry (approved):
THE WHOLE TIME 🤦🏼‍♂️ glad to learn about this too.

brandur:
> THE WHOLE TIME 🤦🏼‍♂️ glad to learn about this too.

Haha, I know right??

PR #465, opened 2024-07-22
Modify migration 005 to tolerate the absence of `river_migration`

brandur:
Version 0.10.0's migration (005) brought in a number of changes
including one that adds a new `line` field to the migration table.

It works, but could present a problem for River installations that are
purposely not using River's migration system, which is a path that even
if not explicitly recommended, we've been trying to support.

Here, add some conditionals to 005 so that the migration-related changes
only run if `river_migration` exists. A test verifies that even if 001
(which brings in `river_migration`) had never run, the up and own
migrations still work.

### DIFF
--- a/riverdriver/riverdatabasesql/migration/main/005_migration_unique_client.down.sql
+++ b/riverdriver/riverdatabasesql/migration/main/005_migration_unique_client.down.sql
@@ -7,35 +7,40 @@
 DO
 $body$
 BEGIN
-    IF EXISTS (
-        SELECT *
-        FROM river_migration
-        WHERE line <> 'main'
-    ) THEN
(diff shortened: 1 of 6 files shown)


riverdriver/riverdatabasesql/migration/main/005_migration_unique_client.down.sql
```diff
@@ -7,36 +7,41 @@
...
-);
-
-CREATE UNIQUE INDEX ON river_migration USING btree(version);
-
-INSERT INTO river_migration
-    (created_at, version)
-SELECT created_at, version
-FROM river_migration_old;
-
-DROP TABLE river_migration_old;
```
brandur, line 38: No actual changes here really, but everything moves up into the `IF` block/script above.

bgentry (approved):

brandur:
Thanks!

PR #468, opened 2024-07-24
Prepare version 0.10.1 for CLI

brandur:
Follow up to #467 with a separate release for the CLI so that it can
internalize changes to the 005 migration.


bgentry (approved):

brandur:
Hm, this is odd, but this PR is still open although we can see it was merged in bf61772d4693ed498f66a7e1697819554bf3f0cb including the custom GitHub squash message. going to close this out.

PR #480, opened 2024-07-29
Give `database/sql` driver Go files a `_driver` suffix

brandur:
A very small change, but here rename `database/sql` driver
implementation files from:

    river_database_sql.go -> river_database_sql_driver.go

This gives them a consistent suffix with `river_pgx_v5_driver.go` files,
and makes them a little easier to match via partial quick find in tools
like VSCode.


bgentry (approved):

brandur:
Thx.

Issue #482, opened 2024-07-30
Panic when using cli migration


nathankawalec:
I'm running into a panic when using the command:

`river migrate-up --database-url postgres://host:5432/db  `

The stack trace is:
```
PS C:\Users\Nathan\source\voicelab\voicelab.ai\voicelab-api> river migrate-up --database-url postgres://host:5432/db  
panic: runtime error: slice bounds out of range [5:4]

goroutine 1 [running]:
github.com/riverqueue/river/rivermigrate.migrationsFromFS.func1({0xc00010d080, 0xe}, {0x0?, 0x0?}, {0x0?, 0x0?})
        C:/Users/Nathan/go/pkg/mod/github.com/riverqueue/river@v0.10.1/rivermigrate/river_migrate.go:596 +0x9a6
io/fs.walkDir({0xb50f80, 0xb65ee0}, {0xc00010d080, 0xe}, {0xb57c98, 0xb65f28}, 0xc000139950)
        C:/Program Files/Go/src/io/fs/walk.go:73 +0x6c
io/fs.walkDir({0xb50f80, 0xb65ee0}, {0xa66e5f, 0x9}, {0xb57ad8, 0xc00010a7c0}, 0xc000139950)
        C:/Program Files/Go/src/io/fs/walk.go:95 +0x2bf
io/fs.WalkDir({0xb50f80, 0xb65ee0}, {0xa66e5f, 0x9}, 0xc000139950)
        C:/Program Files/Go/src/io/fs/walk.go:122 +0xa5
github.com/riverqueue/river/rivermigrate.migrationsFromFS({0xb50f80?, 0xb65ee0?}, {0xa6514e, 0x4})
        C:/Users/Nathan/go/pkg/mod/github.com/riverqueue/river@v0.10.1/rivermigrate/river_migrate.go:573 +0xbd
github.com/riverqueue/river/rivermigrate.New[...]({0xb5b650, 0xc00010e150}, 0x1)
        C:/Users/Nathan/go/pkg/mod/github.com/riverqueue/river@v0.10.1/rivermigrate/river_migrate.go:122 +0x333
github.com/riverqueue/river/cmd/river/rivercli.RunCommand[...].func1.4()
        C:/Users/Nathan/go/pkg/mod/github.com/riverqueue/river/cmd/river@v0.10.1/rivercli/command.go:103 +0x25
[...]

PR #483, opened 2024-07-30
Set up Windows CI for the CLI

bgentry:
In order to help ensure that issues like #482 don't happen again, this adds windows CI for the river CLI. You can see that it exposes the panic from that issue, which is hopefully fixed by #485.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -96,88 +96,94 @@ jobs:
         run: go test -race ./... -timeout 2m
 
   cli:
-    runs-on: ubuntu-latest
-    timeout-minutes: 3
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
 
-    services:
-      postgres:
-        image: postgres
-        env:
-          POSTGRES_PASSWORD: postgres
-        options: >-
-          --health-cmd pg_isready
-          --health-interval 2s
-          --health-timeout 5s
-          --health-retries 5
-        ports:
-          - 5432:5432
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 10
 
     steps:
+      - uses: ikalnytskyi/action-setup-postgres@v6
+        with:
+          database: river_dev
+          password: postgres
+
       - uses: actions/setup-go@v5
         with:
           go-version: "stable"
           check-latest: true
 
+      - name: Setup GOBIN (Linux)
+        if: runner.os == 'Linux'
+        run: |
+          echo "GOBIN=$(go env GOPATH)/bin" >> $GITHUB_ENV
+          echo "PATH=$(go env GOPATH)/bin:$PATH" >> $GITHUB_ENV
+        shell: bash
+
+      - name: Setup GOBIN (Windows)
+        if: runner.os == 'Windows'
+        run: |
+          $gobin = "$(go env GOPATH)\bin"
+          echo "GOBIN=$gobin" >> $env:GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -96,88 +96,94 @@ jobs:
...
-        ports:
-          - 5432:5432
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 10
 
     steps:
+      - uses: ikalnytskyi/action-setup-postgres@v6
+        with:
+          database: river_dev
+          password: postgres
```
bgentry, line 110: This action is cross-platform, unlike containerized services.

.github/workflows/ci.yaml
```diff
@@ -96,88 +96,94 @@ jobs:
...
+          elif [ "$RUNNER_OS" == "Windows" ]; then
+            echo "RIVER_CMD_DIR=.\cmd\river" >> $GITHUB_ENV
+          fi
+        shell: bash
 
-      - name: Create database
-        run: psql --echo-errors --quiet -c '\timing off' -c "CREATE DATABASE river_dev;" ${ADMIN_DATABASE_URL}
+      - name: Build and install CLI
+        run: go install -v .
+        working-directory: ${{ env.RIVER_CMD_DIR }}
```
bgentry, line 146: `go install` instead of `go build` here means we don't have to set the working dir for every subsequent command.

.github/workflows/ci.yaml
```diff
@@ -96,88 +96,94 @@ jobs:
...
+          echo "PATH=$(go env GOPATH)/bin:$PATH" >> $GITHUB_ENV
+        shell: bash
+
+      - name: Setup GOBIN (Windows)
+        if: runner.os == 'Windows'
+        run: |
+          $gobin = "$(go env GOPATH)\bin"
+          echo "GOBIN=$gobin" >> $env:GITHUB_ENV
+          echo $gobin | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+        shell: pwsh
```
bgentry, line 130: It got a little messy trying to do this in a reliably cross-platform way using `bash`, so I decided it was easier to split out the windows-specific form.

.github/workflows/ci.yaml
```diff
@@ -96,88 +96,94 @@ jobs:
...
-          POSTGRES_PASSWORD: postgres
-        options: >-
-          --health-cmd pg_isready
-          --health-interval 2s
-          --health-timeout 5s
-          --health-retries 5
-        ports:
-          - 5432:5432
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 10
```
brandur, line 104: Dang, so just TBC, it looks like this will be by far the slowest job now right? (Last run took 4 min, which is 33% slower than the 3 min the test suite takes.)

River core's CI is still within reasonable bounds time-wise, although it's pretty slow already. We can see how bad it could get by looking at River UI though, where the container jobs are _so_ slow that they really do have a substantial effect on the productivity loop.

brandur (approved):

.github/workflows/ci.yaml
bgentry: Yeah, I haven't really looked into whether it can be optimized yet, though I suspect it's not caching properly due to the fact the step hasn't yet succeeded.

For the time being I'm not going to make it a required check, we'll see where we're at once it's running successfully.

bgentry:
@brandur FYI I updated branch protection rules to require the linux version of the CLI check (removing the plain old `cli` one). Any open branches will need to be rebased.

PR #491, opened 2024-08-01
Prepare v0.10.2 release

brandur:
We have a variety of bug fixes that should go out, so here, prepare a
v0.10.2 release that bundles them up.

We have migration changes, so we'll also need to do a separate phase two
release of the CLI for this version.


bgentry (approved):

brandur:
Thanks!

PR #493, opened 2024-08-02
specify bash shell for windows CI steps needing envs

bgentry:


bgentry:
@brandur this should be just about there, just dealing with the fact that graceful shutdowns with sigterm aren't a thing on windows and we'd need to come up with another way to do this cleanly.

.github/workflows/ci.yaml
```diff
@@ -242,56 +259,3 @@ jobs:
         run: |
           echo "Make sure that all sqlc changes are checked in"
           make verify/sqlc
-
-  producer_sample:
```
brandur, line 246: Just to make sure: you took this out on purpose?

brandur (approved):
Great!

.github/workflows/ci.yaml
bgentry: Yeah, I don’t think it’s needed anymore. We have enough coverage elsewhere imo.

PR #505, opened 2024-08-06
Account for schema when checking whether a column exists

### LINKED ISSUE #503: ColumnExists query breaks when there are multiple river queues across different schemas
I am using a River setup where we have multiple river queues with the db for each located in a different schema. This was based on a discussion with one of my coworkers [here](https://github.com/riverqueue/river/discussions/343). When trying to run the most recent migrations the migration succeeded for the first schema/queue but then failed for the second and third schema. The error message was 
```
❯ river migrate-up --database-url "postgres://user:pass@localhost:1234/chariot?search_path=SCHEMA_NAME"
failed: error getting existing migrations for line "main": ERROR: column "line" does not exist (SQLSTATE 42703)
```
After digging into the code I found that the River CLI checks whether the "line" column exists is through [the ColumnExists query](https://github.com/riverqueue/river/blob/08ef6349316815d667e672d19b4fb389af464249/riverdriver/riverpgxv5/internal/dbsqlc/river_migration.sql#L71-L76).

The issue with this is it doesn't take into account the schema on the search path, so after one migration has run, the query will always return true. This caused [this code](https://github.com/riverqueue/river/blob/08ef6349316815d667e672d19b4fb389af464249/rivermigrate/river_migrate.go#L540-L551) to be skipped, which I assume was meant to be run when the line was missing and then we see the error above. 

I'm not sure why the query doesn't respect the search path we provide in the database url. I'm not sure if there's a workaround here - I'd appreciate any help!

brandur:
This one's aimed at addressing #503, in which migrations won't run when
using an alternate schema with `search_path` because the `ColumnExists`
query isn't accounting for schema in any way.

Here, modify `ColumnExists` so that it looks up within `CURRENT_SCHEMA`,
which will respect search path. Add test assertions to verify that this
works as expected, and similar ones in `TableExists` to avoid any
regressions there (it already does respect search path as written, but a
future change could potentially break that).

Fixes #503.

### DIFF
--- a/internal/riverinternaltest/riverdrivertest/riverdrivertest.go
+++ b/internal/riverinternaltest/riverdrivertest/riverdrivertest.go
@@ -192,6 +192,26 @@ func Exercise[TTx any](ctx context.Context, t *testing.T,
 		exists, err = exec.ColumnExists(ctx, "river_job", "does_not_exist")
 		require.NoError(t, err)
 		require.False(t, exists)
+
+		exists, err = exec.ColumnExists(ctx, "does_not_exist", "id")
+		require.NoError(t, err)
+		require.False(t, exists)
+
+		// Will be rolled back by the test transaction.
+		_, err = exec.Exec(ctx, "CREATE SCHEMA another_schema_123")
+		require.NoError(t, err)
+
+		_, err = exec.Exec(ctx, "SET search_path = another_schema_123")
+		require.NoError(t, err)
+
+		// Table with the same name as the main schema, but without the same
+		// columns.
+		_, err = exec.Exec(ctx, "CREATE TABLE river_job (another_id bigint)")
+		require.NoError(t, err)
+
+		exists, err = exec.ColumnExists(ctx, "river_job", "id")
+		require.NoError(t, err)
+		require.False(t, exists)
 	})
 
 	t.Run("Exec", func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


internal/riverinternaltest/riverdrivertest/riverdrivertest.go
```diff
@@ -192,6 +192,26 @@ func Exercise[TTx any](ctx context.Context, t *testing.T,
...
+
+		exists, err = exec.ColumnExists(ctx, "does_not_exist", "id")
+		require.NoError(t, err)
+		require.False(t, exists)
+
+		// Will be rolled back by the test transaction.
+		_, err = exec.Exec(ctx, "CREATE SCHEMA another_schema_123")
+		require.NoError(t, err)
+
+		_, err = exec.Exec(ctx, "SET search_path = another_schema_123")
```
bgentry, line 204: Does this pollute the search path for this connection for future tests? Or does the conn pool get thrown away entirely after each test before the test DB goes back into the DB pool?

bgentry (approved):

internal/riverinternaltest/riverdrivertest/riverdrivertest.go
brandur: It gets rolled back with the transaction:

``` sql
river_dev=# show search_path;
                 search_path
---------------------------------------------
 "$user", crunchy_cloud, infra, logs, public
(1 row)


river_dev=# begin;
BEGIN

river_dev=*# set search_path = 'hello';
SET

river_dev=*# show search_path;
 search_path
-------------
 hello
(1 row)

river_dev=*# rollback;
ROLLBACK

river_dev=# show search_path;
                 search_path
---------------------------------------------
 "$user", crunchy_cloud, infra, logs, public
(1 row)
```

PR #514, opened 2024-08-07
Derive all internal contexts from Client context

### LINKED ISSUE #512: Cannot test work that inserts jobs using `river.ClientFromContext`
If I use `river.ClientFromContext()` within a job, and test it using the `Work` method (as docs suggest), I cannot insert the client into the context. Could be solved by adding a `river.ContextWithClient` helper, exposing `ctxKey`, etc.

bgentry:
We were previously using `context.Background()` in some situations to avoid cancellation of things that really shouldn't be cancelled by the user context cancellation. However, we can now do that with `context.WithoutCancel` instead in order to preserve context values without any cancellation or deadline.

Fixes #512.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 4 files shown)


brandur (approved):
Looks great!

PR #516, opened 2024-08-08
rivertest.RequireNotInsertedTx: fix incorrect description

bgentry:
Noticed this description was wrong from some copy pasta 🍝  https://pkg.go.dev/github.com/riverqueue/river/rivertest#RequireNotInsertedTx

Set to auto-merge, make edits before approving if you want to please 🙏


bgentry:
Whoops @brandur, it looks like those branch protections changes I made allowed this to merge without approvals once CI had passed. I put them back in place but with an option to bypass, that should do it. Please take a look at this regardless if you can!

PR #517, opened 2024-08-08
Add `startstop.StartAll` helper to start a series of services

brandur:
Add a `StartAll` helper that starts a number of services at once, also
taking care that in the event of a problem, any services that had
started are once stopped. I found the use of something like this
convenient while trying some service code in River UI.

I also make a couple tweaks so that the `Stopped` channel becomes a
little safer to use. Like `Started`, it will preallocate itself if
called before a service is started. It's still not safe to call while
stopping or after a stop, but its ergonomics still measurably improve.

Lastly, I pull a somewhat unrelated change to add a `NewArchetype`
function in for easily creating an archetype in non-test environments.
Also found this useful when bringing use of archetypes into River UI.

### DIFF
--- a/client.go
+++ b/client.go
@@ -26,7 +26,6 @@ import (
 	"github.com/riverqueue/river/rivershared/startstop"
 	"github.com/riverqueue/river/rivershared/testsignal"
 	"github.com/riverqueue/river/rivershared/util/maputil"
-	"github.com/riverqueue/river/rivershared/util/randutil"
 	"github.com/riverqueue/river/rivershared/util/sliceutil"
 	"github.com/riverqueue/river/rivershared/util/valutil"
 	"github.com/riverqueue/river/rivertype"
@@ -337,8 +336,8 @@ type Client[TTx any] struct {
 	queueMaintainer      *maintenance.QueueMaintainer
 	queues               *QueueBundle
 	services             []startstop.Service
-	subscriptionManager  *subscriptionManager
 	stopped              <-chan struct{}
+	subscriptionManager  *subscriptionManager
 	testSignals          clientTestSignals
 	uniqueInserter       *dbunique.UniqueInserter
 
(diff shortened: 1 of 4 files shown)


bgentry (commented):
There seems to be something wrong w/ this breaking all the tests, but aside from that it looks good!

client.go
```diff
@@ -858,12 +848,14 @@ func (c *Client[TTx]) StopAndCancel(ctx context.Context) error {
...
+// the channel must be taken _before_ calling Stop or StopAndCancel. The channel
+// is reset after the service is stopped, so if a reference isn't taken before
+// stopping, then the caller might (it's a race) receive a channel intended for
+// use the next time the client is stopped.
 //
 // It is not affected by any contexts passed to Stop or StopAndCancel.
-func (c *Client[TTx]) Stopped() <-chan struct{} {
-	return c.stopped
-}
+func (c *Client[TTx]) Stopped() <-chan struct{} { return c.baseStartStop.Stopped() }
```
brandur, line 858: So because `BaseStartStop.Stopped` now locks a mutex, I had to change the client here so that instead of take a reference to `Stopped` in `Start` (which no longer works because the `BaseStartStop`'s mutex is held during `Start`), I had to change this so that `Stopped` is a direct passthrough into `BaseStartStop.Stopped`.

This does make the code nicer (no longer necessary to have a local `Client.stopped` variable and initialize it, but does result in a subtle change in behavior which I'll mention in the comment below.

client_test.go
```diff
@@ -1071,15 +1071,21 @@ func Test_Client_StopAndCancel(t *testing.T) {
 
 		client, _ := setup(t)
 
+		// A reference to the stopped channel should be taken before calling Stop.
+		stopped := client.Stopped()
```
brandur, line 1075: Making `Client.Stopped` a pass through to `BaseStartStop.Stopped` does now require that a reference to the `Stopped` channel be taken _before_ calling `Stop` or `StopAndCancel`. Calling it after bears the risk that you might be getting the _next_ `Stopped` channel which is being preemptively initialized for the next time the client is `Started`.

This does have the advantage though that you can take a reference to `Stopped` before calling a client's `Start`, something that previously didn't work.

brandur:
Hm, I think this one may have snowballed too much.

client.go
brandur: (NM, changed this.)

client_test.go
brandur: (NM, changed this.)

brandur:
@bgentry Okay I ended up reverting the behavior changes that I mentioned in comments above. I still did have to tweak how `Stopped` work in `BaseStartStop` so I could accurately test the new `StartAll`, but I put in an escape hatch so that `river.Client`'s `Stopped` could behave identically to how it did before. Mind taking a look?

bgentry (approved):

brandur:
Thx.

PR #518, opened 2024-08-08
Simplify names of base service types involving generics for logging

brandur:
A small improvement to base service so that when it's using reflect to
get a name of the service's type for logging purposes, it strips a long
path out of a generic type for a more succinct result.

So this long, unsightly blemish:

    QueryCacher[[]*github.com/riverqueue/riverui/internal/dbsqlc.JobCountByStateRow]

Would become this, which is quite a bit more tolerable:

    QueryCacher[[]*dbsqlc.JobCountByStateRow]

### DIFF
--- a/rivershared/baseservice/base_service.go
+++ b/rivershared/baseservice/base_service.go
@@ -11,6 +11,7 @@ import (
 	"math"
 	"math/rand"
 	"reflect"
+	"regexp"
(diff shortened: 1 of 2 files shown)


bgentry (approved):

brandur:
thx.

PR #522, opened 2024-08-09
Establish minimum Go version of 1.21, with toolchain of Go 1.22

### LINKED ISSUE #520: Version in `go.mod` could be lowered
I need to use a slightly older Docker base image that only has Go 1.22.3 available. I had to pin back to River 0.9, because River 0.10 on uses 1.22.5. Ideally it'd use 1.22.0 unless there is some compelling reason, so people stuck on older 1.22 versions do not need to pin to older versions.

Hopefully I'm not off-base with this suggestion- given Go language versioning rules, it seems like I should be able to ignore the patch version, but as of Go 1.21 I don't think that's possible. So River version 0.10 and newer is only compatible with Go versions 1.22.5 and higher, released July 2 2024. But please let me know if I'm misunderstanding the issue I'm having, I don't follow Go packaging closely so may have missed something.

brandur:
This one related to #520, in which it turns out that our Go 1.21 build
step has actually been automatically upgrading itself to Go 1.22, so
under the radar that we didn't notice.

Here, leverage a Go "toolchain" to establish a preferred version of Go
for the project, but keeping a `go` directive in `go.mod`s that's at the
1.21 lower bound that we're trying to support. The presence of a
`toolchain` directive prevents `go mod tidy` from upgrading the `go`
directive to the latest version of Go installed.

I found the easiest way to add the toolchain directive to be with `go
get` [2] like:

    go get go@1.21 toolchain@1.22.5

Maintaining this correctly is going to be a little tricky. This is one
of those classic Go features that kind of works, but various Go commands
will perform all kinds of confusing behavior like stripping directives
out of your file if there's anything even a tiny bit wrong, and with no
explanation whatsoever.

I found that to get this working I had to started with the "inner"
dependencies like `rivershared` and `rivertype`, `go get` Go/toolchain
there first, then work my way up the stack all the way up to the main
project.

I'm going to try and follow this up with some tooling to help make this
easier for ourselves, and a build check that verifies in CI that nothing
gets accidentally regressed as we make changes because this is _very_
easy to do accidentally.

Fixes #520.

[1] https://go.dev/doc/toolchain
[2] https://go.dev/doc/toolchain#get


cmd/river/go.mod
```diff
@@ -1,23 +1,23 @@
...
-// replace github.com/riverqueue/river/riverdriver/riverdatabasesql => ../../riverdriver/riverdatabasesql
+replace github.com/riverqueue/river/riverdriver => ../../riverdriver
 
-// replace github.com/riverqueue/river/riverdriver/riverpgxv5 => ../../riverdriver/riverpgxv5
+replace github.com/riverqueue/river/riverdriver/riverdatabasesql => ../../riverdriver/riverdatabasesql
 
-// replace github.com/riverqueue/river/rivertype => ../../rivertype
+replace github.com/riverqueue/river/riverdriver/riverpgxv5 => ../../riverdriver/riverpgxv5
+
+replace github.com/riverqueue/river/rivertype => ../../rivertype
```
brandur, line 15: This'll need the two stage release. Without these `replace` directives, `go mod tidy` detects that Go 1.21 can't be supported and therefore removes the `toolchain` directive and changes `go` from 1.21 to 1.22.

`replace` directives to be removed during second phase of release.

bgentry: Is this why setup-go is still upgrading us to 1.22?

```
Setup go version spec 1.21
Found in cache @ /opt/hostedtoolcache/go/1.21.[12](https://github.com/riverqueue/river/actions/runs/10312078359/job/28546717441#step:4:13)/x64
Added go to the path
Successfully set up Go version 1.21
go: downloading go1.22.5 (linux/amd64)
/opt/hostedtoolcache/go/1.21.12/x64/bin/go env GOMODCACHE
/opt/hostedtoolcache/go/1.21.12/x64/bin/go env GOCACHE
/home/runner/go/pkg/mod
/home/runner/.cache/go-build
Cache Size: ~1[16](https://github.com/riverqueue/river/actions/runs/10312078359/job/28546717441#step:4:17) MB (122068102 B)
/usr/bin/tar -xf /home/runner/work/_temp/3fd374b5-c372-4fe0-bb5a-72c9ec40afc9/cache.tzst -P -C /home/runner/work/river/river --use-compress-program unzstd
Received 12[20](https://github.com/riverqueue/river/actions/runs/10312078359/job/28546717441#step:4:21)68102 of 122068102 (100.0%), 116.4 MBs/sec
Cache restored successfully
Cache restored from key: setup-go-Linux-ubuntu22-go-1.22.5-7e42e876c7fc43348578ac5ddfec93c5643eb4a8c880ab849c10c6fbcbbb6cc8
go version go1.[22](https://github.com/riverqueue/river/actions/runs/10312078359/job/28546717441#step:4:23).5 linux/amd64
```

brandur: I think that might be something else -- even Go 1.21 is capable of upgrading to a Go 1.22 toolchain, so the default `GOTOOLCHAIN` value of `auto` is causing Go to install the matrix version, but then fetch one according to `toolchain` in `go.mod`.

.github/workflows/ci.yaml
```diff
@@ -18,6 +18,12 @@ on:
 
 jobs:
   build_and_test:
+    env:
+      # The special value "local" tells Go to use the bundled Go version rather
+      # than trying to fetch one according to a `toolchain` value in `go.mod`.
+      # This ensures that we're really running the Go version in the CI matrix
+      # rather than one that the Go command has upgraded to automatically.
+      GOTOOLCHAIN: local
```
brandur, line 26: @bgentry I think this is the magic invocation needed to stop CI from autoupgrading its toolchain. A value of "local" tells Go to ignore a `go.mod` `toolchain` directive and use the version in the installed Go command.

See this section of the docs:

https://go.dev/doc/toolchain#select

> When GOTOOLCHAIN is set to local, the go command always runs the bundled Go toolchain.

cmd/river/go.mod
```diff
@@ -1,16 +1,20 @@
 module github.com/riverqueue/river/cmd/river
 
-go 1.22.5
+go 1.21
```
brandur, line 3: @bgentry I changed all of these to 1.21 instead of a specific patch version like 1.21.13.

1.21 == 1.21.0

I think this is probably right because it gives maximum flexibility for people installing the package. They should still use the latest possible patch version of course, but it's not always possible due to their local build toolchain. Specifying 1.21.0 just seems like the much easier choice rather than trying to find a minimum patch version that's most universally available.

Do you agree?

riverdriver/riverdatabasesql/river_database_sql_driver_test.go
```diff
@@ -78,6 +78,8 @@ func TestReplaceNamed(t *testing.T) {
 		{Desc: "SliceUint64", ExpectedSQL: "SELECT ARRAY[123,124]", InputSQL: "SELECT @slice_uint64", InputArgs: map[string]any{"slice_uint64": []uint64{123, 124}}},
 	}
 	for _, tt := range testCases {
+		tt := tt
```
brandur, line 81: Something I hadn't realized is that golangci-lint is actually reading the Go version from `go.mod`. So as we reset the min version down to 1.21, the lint requiring that loop vars are captured kicked back in.

bgentry (approved):

brandur:
Thx.

PR #528, opened 2024-08-13
Move submodule update command to `rivershared` + check for go/toolchain + workspace

brandur:
This one's a bit of a grab bucket, but do a few refactors and additions:

* Bring the project over to use a Go workspace, thereby letting us run
  tests and programs across module directories and cut out our `replace`
  statements in `go.mod` all over the place.

* Move the River version update program for `go.mod` files over to
  `rivershared` so that it can be used by other River projects. It's
  modified so that instead of hardcoding a submodule list, it can now
  read them straight out of the project's workspace in `go.work`.

* Add a new program that's similar to this one whose job it is to check
  that the `go`/`toolchain` directives across all `go.mod`s in the
  workspace match, which will help prevent accidental regressions of
  #522. It can also be used locally to update `go`/`toolchain`
  directives to new values in case we want to do that in the future.

* Do a lot of repaving in `Makefile` so that we can use list of
  submodules from the workspace to run a series of commands instead of
  needing to maintain independent lists for each target, a practice that
  was often unreliable.


Makefile
```diff
@@ -1,71 +1,87 @@
...
-	cd riverdriver && golangci-lint run --fix
-	cd riverdriver/riverdatabasesql && golangci-lint run --fix
-	cd riverdriver/riverpgxv5 && golangci-lint run --fix
-	cd rivershared && golangci-lint run --fix
-	cd rivertype && golangci-lint run --fix
+lint:: ## Run linter (golangci-lint) for all submodules
+define lint-target
+    lint:: ; cd $1 && golangci-lint run --fix
+endef
+$(foreach mod,$(submodules),$(eval $(call lint-target,$(mod))))
```
brandur, line 51: This looks a little weird, but out of all the options I tried, it produces by far the best output. Basically equivalent to what we had before:

```
$ make lint
cd /Users/brandur/Documents/projects/river && golangci-lint run --fix
cd /Users/brandur/Documents/projects/river/cmd/river && golangci-lint run --fix
cd /Users/brandur/Documents/projects/river/riverdriver && golangci-lint run --fix
cd /Users/brandur/Documents/projects/river/riverdriver/riverdatabasesql && golangci-lint run --fix
cd /Users/brandur/Documents/projects/river/riverdriver/riverpgxv5 && golangci-lint run --fix
cd /Users/brandur/Documents/projects/river/rivershared && golangci-lint run --fix
cd /Users/brandur/Documents/projects/river/rivertype && golangci-lint run --fix
```

The problem with using something like a `for` loop is that you just get the for loop printed, so the output is completely useless because you can't tell which module is currently running.

Makefile
```diff
@@ -1,71 +1,87 @@
...
-generate/sqlc:
+generate/sqlc: ## Generate sqlc
 	cd riverdriver/riverdatabasesql/internal/dbsqlc && sqlc generate
 	cd riverdriver/riverpgxv5/internal/dbsqlc && sqlc generate
 
+# Looks at comments using ## on targets and uses them to produce a help output.
+.PHONY: help
+help: ALIGN=22
+help: ## Print this message
+	@awk -F '::? .*## ' -- "/^[^':]+::? .*## /"' { printf "'$$(tput bold)'%-$(ALIGN)s'$$(tput sgr0)' %s\n", $$1, $$2 }' $(MAKEFILE_LIST)
```
brandur, line 36: This is kind of a minor nicety that I've been putting in a few projects. Just lets you run `make help` and get this:

```
$ make help
db/reset               Drop, create, and migrate dev and test databases
db/reset/dev           Drop, create, and migrate dev database
db/reset/test          Drop, create, and migrate test databases
generate               Generate generated artifacts
generate/migrations    Sync changes of pgxv5 migrations to database/sql
generate/sqlc          Generate sqlc
help                   Print this message
lint                   Run linter (golangci-lint) for all submodules
test                   Run test suite for all submodules
tidy                   Run `go mod tidy` for all submodules
update-mod-go          Update `go`/`toolchain` directives in all submodules to match `go.work`
update-mod-version     Update River packages in all submodules to $VERSION
verify                 Verify generated artifacts
verify/migrations      Verify synced migrations
verify/sqlc            Verify generated sqlc
```

rivershared/cmd/update-mod-version/main.go
```diff
@@ -14,23 +18,9 @@ import (
...
-// `go.sum`, so any changes to its `go.mod` file would require a `go mod tidy`
-// be run afterwards.
-var allProjectModules = []string{ //nolint:gochecknoglobals
-	".",
-	"./riverdriver",
-	"./riverdriver/riverdatabasesql",
-	"./riverdriver/riverpgxv5",
-	"./rivershared",
-	"./rivertype",
-}
```
brandur, line 29: Nice: these bespoke lists of modules in various files like here and `Makefile` all go away.

brandur:
@bgentry Just to summarize a bit, this one brings in:

* Workspace.
* Modified the `go.mod` version update script so it goes to `rivershared` and can be used by other River projects (so we don't have the code duplicated in every repo).
* Added new script that verifies all versions are in sync. See new `submodule_check` CI job.

rivershared/cmd/update-mod-go/main.go
```diff
@@ -0,0 +1,124 @@
+// Package main provides a command to help bump the `go/`toolchain` directives
```
bgentry, line 1: The recommended convention on `main` packages is to name this first word after the program name, rather than `main`: https://tip.golang.org/doc/comment#cmd

bgentry (approved):
Fantastic stuff! This was the missing piece in our workflow that we needed months ago 😆

rivershared/cmd/update-mod-go/main.go
brandur: Ah interesting, TIL. Updated both programs.

brandur:
Thanks!

As discussed, let's give this a shot in the main project for a while, and revert in case it doesn't work out for some reason.

PR #529, opened 2024-08-14
Only log hearbeat if status changed

arp242:
Right now it will log this every five seconds, which really clobbers my terminal/logs, mostly with fairly useless debug messages:

	 15:27 INFO  producer: Heartbeat
		     num_completed_jobs = 0
		     num_jobs_running   = 0
		     queue              = default
	 15:27 INFO  producer: Heartbeat
		     num_completed_jobs = 0
		     num_jobs_running   = 0
		     queue              = default
	 15:27 INFO  producer: Heartbeat
		     num_completed_jobs = 0
		     num_jobs_running   = 0
		     queue              = default
	 15:27 INFO  producer: Heartbeat
		     num_completed_jobs = 0
		     num_jobs_running   = 0
		     queue              = default

I originally considered changing the InfoContext() to DebugContext(), but doing an Info log only if the status changed seemed like a good middle ground.

### DIFF
--- a/producer.go
+++ b/producer.go
@@ -535,16 +535,25 @@ func (p *producer) dispatchWork(workCtx context.Context, count int, fetchResultC
 func (p *producer) heartbeatLogLoop(ctx context.Context) {
 	ticker := time.NewTicker(5 * time.Second)
 	defer ticker.Stop()
+	type jobCount struct {
+		ran    uint64
+		active int
+	}
+	var prevCount jobCount
 	for {
 		select {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			p.Logger.InfoContext(ctx, p.Name+": Heartbeat",
-				slog.Uint64("num_completed_jobs", p.numJobsRan.Load()),
-				slog.Int("num_jobs_running", int(p.numJobsActive.Load())),
-				slog.String("queue", p.config.Queue),
-			)
+			curCount := jobCount{ran: p.numJobsRan.Load(), active: int(p.numJobsActive.Load())}
+			if curCount != prevCount {
+				p.Logger.InfoContext(ctx, p.Name+": Producer job counts",
+					slog.Uint64("num_completed_jobs", curCount.ran),
+					slog.Int("num_jobs_running", curCount.active),
+					slog.String("queue", p.config.Queue),
+				)
+			}
+			prevCount = curCount
 		}
 	}
 }


producer.go
```diff
@@ -535,16 +535,25 @@ func (p *producer) dispatchWork(workCtx context.Context, count int, fetchResultC
 func (p *producer) heartbeatLogLoop(ctx context.Context) {
 	ticker := time.NewTicker(5 * time.Second)
 	defer ticker.Stop()
+	type state struct {
```
brandur, line 538: I  think the naming on this is just a little too generic. Maybe `jobStatistics`?

producer.go
```diff
@@ -535,16 +535,25 @@ func (p *producer) dispatchWork(workCtx context.Context, count int, fetchResultC
...
+		jobsRan    uint64
+		jobsActive int
+	}
+	prev := state{}
```
brandur, line 542: And similarly here too. Maybe:

* `cur` -> `curStats`
* `prev` -> `prevStats`

producer.go
```diff
@@ -535,16 +535,25 @@ func (p *producer) dispatchWork(workCtx context.Context, count int, fetchResultC
...
 			return
 		case <-ticker.C:
-			p.Logger.InfoContext(ctx, p.Name+": Heartbeat",
-				slog.Uint64("num_completed_jobs", p.numJobsRan.Load()),
-				slog.Int("num_jobs_running", int(p.numJobsActive.Load())),
-				slog.String("queue", p.config.Queue),
-			)
+			cur := state{jobsRan: p.numJobsRan.Load(), jobsActive: int(p.numJobsActive.Load())}
+			if cur != prev {
+				p.Logger.InfoContext(ctx, p.Name+": Heartbeat",
```
brandur, line 550: Removing the guaranteed tick means this isn't really a heartbeat anymore. Maybe rephrase to?

```suggestion
				p.Logger.InfoContext(ctx, p.Name+": Producer job stats",
```

brandur (commented):
Left a couple comments, but WFM.

producer.go
arp242: It's not really "statistics", right? At least, not how I would use the word: It's the state at the time this is run. I changed it to `jobState`, but if you want `jobStatistics` then I can change it to that.

producer.go
arp242: Changed to `p.Logger.InfoContext(ctx, p.Name+": producer job state changed"`

producer.go
brandur: I'm open to alternatives, but the new log message is definitely wrong. Try to zoom out and re-read "producer job state changed" — that sounds like the state of a job is changing. "State" is a common word used in the API for jobs indicating their currently state. e.g. "available", "completed", etc.

Maybe "job counts" or something?

ccoVeille:
I like your changes, but I would like to suggest something.

The heartbeat could be logged on each call with debug level, and the state change is logged with info level.

It would help a lot when a debug is needed.

I'm a bit afraid no log at all could lead to problem

producer.go
arp242: > Maybe "job counts" or something?

Yes, that's better :+1:

producer.go
```diff
@@ -535,16 +535,25 @@ func (p *producer) dispatchWork(workCtx context.Context, count int, fetchResultC
...
 			return
 		case <-ticker.C:
-			p.Logger.InfoContext(ctx, p.Name+": Heartbeat",
-				slog.Uint64("num_completed_jobs", p.numJobsRan.Load()),
-				slog.Int("num_jobs_running", int(p.numJobsActive.Load())),
-				slog.String("queue", p.config.Queue),
-			)
+			curCount := jobCount{ran: p.numJobsRan.Load(), active: int(p.numJobsActive.Load())}
+			if curCount != prevCount {
+				p.Logger.InfoContext(ctx, p.Name+": producer job count changed",
```
brandur, line 550: Small, but can you change back to capitalizing the start of sentence? We're pretty consistent on this throughout the rest of the code.

```suggestion
				p.Logger.InfoContext(ctx, p.Name+": Producer job count changed",
```

arp242: Ah sorry; I'm so used to lower-casing after the `:` I just didn't notice.

arp242:
> I'm a bit afraid no log at all could lead to problem

@ccoVeille  I don't think this will actually log less information? It just won't log if the message will be identical (because none of the log parameters changed).

ccoVeille:
heartbeat will be silent if nothing changed.

That's what your PR is about.

But won't it log nothing for hours if nothing changed. Or maybe I missed something in the conditions.

What will affect the count?

brandur:
@ccoVeille So I think the idea is that while yes, nothing gets logged if nothing is changing, this might be a reasonable compromise for most systems. i.e. Any systems were any jobs are moving through queues at all, you'll see logging basically the same as you did before. But for systems where nothing is happening (like in say dev), we bias towards quieter log outputs.

This seems to make sense to me, although I can imagine an alternative where we should make it a `Debug` line and the client gets even quieter, and those who want a higher output client can either use debug level, or attach their own hooks into the even system.

Thoughts?

producer.go
```diff
@@ -535,16 +535,25 @@ func (p *producer) dispatchWork(workCtx context.Context, count int, fetchResultC
...
 			return
 		case <-ticker.C:
-			p.Logger.InfoContext(ctx, p.Name+": Heartbeat",
-				slog.Uint64("num_completed_jobs", p.numJobsRan.Load()),
-				slog.Int("num_jobs_running", int(p.numJobsActive.Load())),
-				slog.String("queue", p.config.Queue),
-			)
+			curCount := jobCount{ran: p.numJobsRan.Load(), active: int(p.numJobsActive.Load())}
+			if curCount != prevCount {
+				p.Logger.InfoContext(ctx, p.Name+": Producer job count changed",
```
brandur, line 550: Can we just make this "counts"? Saying "changed" seems to suggest the log was fired by something changing, which isn't really the case since it's just periodic output.

```suggestion
				p.Logger.InfoContext(ctx, p.Name+": Producer job counts",
```

brandur (approved):

brandur:
Thanks!

ccoVeille:
I'm fine and agree with what you said.

I might open a PR with what I was thinking about. But my initial comment was a suggestion while the changes involved in the PR were already great.

So no worries

PR #530, opened 2024-08-15
Upgrade to golangci-lint 1.60 + linting fixes

brandur:
This is in preparation for an eventual upgrade to Go 1.23. Golangci-lint
1.60+ is required for 1.23 support, so here upgrade to 1.23 and fix some
new linting problems that fell out of the process.


bgentry (approved):

brandur:
Thanks!

PR #532, opened 2024-08-18
Accept `postgresql://`-schemed database URLs

yawboakye:
According to the [official PostgresQL documentation](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS) (referenced 18/08/2024), `postgresql://` is an acceptable database URL scheme. In fact that's what I use in my application. So I was surprised when River's migrations were failing. Turns out River only supports `postgres://`-schemed URLs.

Given that `postgresq://` is accepted, and perhaps the preferred, I suggest that River accepts it to. Other changes are: fix grammar typo (unsupported instead of unsupport) and printing the URL the command was called with. I think it's good UX, especially if the user passed an environment variable.


cmd/river/rivercli/river_cli.go
```diff
@@ -94,7 +94,7 @@ Provides command line facilities for the River job queue.
 	}
 
 	addDatabaseURLFlag := func(cmd *cobra.Command, databaseURL *string) {
-		cmd.Flags().StringVar(databaseURL, "database-url", "", "URL of the database to benchmark (should look like `postgres://...`")
+		cmd.Flags().StringVar(databaseURL, "database-url", "", "URL of the database to benchmark (should look like `postgres://...` or `postgresql://...`")
```
brandur, line 97:
```suggestion
		cmd.Flags().StringVar(databaseURL, "database-url", "", "URL of the database to benchmark (should look like `postgres://...`")
```

Maybe revert his line ... `postgres://` implies that the `postgresql://` is also supported, and generally it's better if these help lines are terser so they don't line wrap.

cmd/river/rivercli/command.go
```diff
@@ -17,6 +16,16 @@ import (
 	"github.com/riverqueue/river/rivermigrate"
 )
 
+const (
+	// Referencing the offical PostgreSQL documentation, it appears that
+	// `postgresql://` is the default/preferred URI scheme, with `postgres://`
+	// as the fallback. River accepts both.
+	//
+	// See https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS.
```
brandur, line 24: Thank you for the comment (usually I think these are quite a good idea), but maybe drop this one. It feels oddly unsure of itself ("it appears that"), and also the naming of the constants make it so obvious what's going on here that I don't think it's really needed.

brandur (commented):
Wow, odd oversight! The fact that no one's reported this before seems to suggest the sheer ubiquity of the `postgres://` convention.

Asked for a couple tweaks here, but LGTM.

yawboakye:
@brandur i had been waiting for the checks to turn green before requesting for another review, but for reasons i don’t know they’ve stayed amber.

brandur:
>  had been waiting for the checks to turn green before requesting for another review, but for reasons i don’t know they’ve stayed amber.

Ah yeah, GitHub requires us to approve CI runs from external contributors for security reasons (in case private vars or such leak into the build and could be extracted).

brandur (approved):

brandur:
Thanks!

arp242:
> The fact that no one's reported this before seems to suggest the sheer ubiquity of the postgres:// convention.

I ran into this as well a few weeks ago; I just changes postgresql:// to postgres://and called it a day 🙃

In-between doing a ton of other things and "just trying out this River thing to see if it works for us" (it does), I just never got around to reporting it.

I would be surprised if I was the first person. I've often found that getting feedback on these types of small things can be very difficult.

brandur:
@arp242 Ah, fair enough. We used to have a rule of thumb at my last job that "for every one user that opened a support ticket, 10 users ran into the same problem and didn't". That's probably what happened here.

PR #533, opened 2024-08-18
Add Go 1.23 to build matrix + update toolchain

brandur:
Add Go 1.23 to the build matrix and update toolchain in `go.work` and
all `go.mod` files. Add development instructions for how to use the new
`make update-mod-go` to do so in an automated way.

I didn't drop Go 1.21 support yet because I figure there isn't a hurry
in jettisoning it immediately, although, we'll have to see how the build
stability is with the expanded matrix.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,6 +30,7 @@ jobs:
         go-version:
           - "1.21"
           - "1.22"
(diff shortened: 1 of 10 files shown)


bgentry (approved):

brandur:
ty.

PR #535, opened 2024-08-19
Prepare release v0.11.3

brandur:
Prepare release v0.11.3, which includes largely #532, which is a
legitimate bug in the CLI.

This will be the first time I try for a release under workspaces (#528),
so it includes a simultaneous release of the CLI and the rest of the
project, although I'm expecting the build to fail due to lack of
`replace` statements.


bgentry (approved):

brandur:
Thanks!

PR #536, opened 2024-08-20
Break `BaseService` functions into `serviceutil`

brandur:
I was going through trying to refactor the demo project to use
`rivershared`, and while doing so, noticed that it'd broken out some
of the `BaseService` functions like `CancellableSleep` into a new
package so that they could be used without an available `BaseService`.

Looking over this API again, there's no good reason for functions like
`CancellableSleep` and `ExponentialBackoff` to be tied so tight to
`BaseService`. The reason they are is that `BaseService` provides a
random source that both can conveniently use, but that can be moved to a
function argument instead.

The reason that `BaseService` has a random source is to avoid the old
`math/rand` seeding problem. A single random source is initialized and
seeding securely once, then placed onto an archetype that's inherited
by all `BaseService` instances.

Soon, even that can go away. One of the biggest things fixed by Go
1.22's `math/rand/v2` is that it's no longer necessary to seed the top
level functions. Once we drop support for Go 1.21, we'll be able to
simplify this code dramatically by dropping `BaseService.Rand` and the
random source arguments from functions like `ExponentialBackoff` in
favor of just using top level `math/rand/v2` functions.

I also drop the variant `CancellableSleepBetween` in favor of having
callers use `CancellableSleep` combined with `randutil.DurationBetween`,
a new random helper similar to `IntBetween`.

With all this in, we'll be able to fully jettison all utilities from the
demo project in favor of going all in with `rivershared`'s equivalents.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -59,9 +59,6 @@ jobs:
       - name: Display Go version
         run: go version
 
-      - name: Install dependencies
-        run: go get -t ./...
-
       - name: Set up test DBs
         run: go run ./internal/cmd/testdbman create
         env:
--- a/client_test.go
+++ b/client_test.go
@@ -33,6 +33,8 @@ import (
 	"github.com/riverqueue/river/rivershared/riversharedtest"
 	"github.com/riverqueue/river/rivershared/startstoptest"
 	"github.com/riverqueue/river/rivershared/util/ptrutil"
+	"github.com/riverqueue/river/rivershared/util/randutil"
+	"github.com/riverqueue/river/rivershared/util/serviceutil"
 	"github.com/riverqueue/river/rivershared/util/sliceutil"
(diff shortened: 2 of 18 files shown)


.github/workflows/ci.yaml
```diff
@@ -59,9 +59,6 @@ jobs:
       - name: Display Go version
         run: go version
 
-      - name: Install dependencies
-        run: go get -t ./...
```
brandur, line 63: A workspace side effect: adding a new package to rivershared seems to make it hard to resolve by `go get -t` despite `go test` and another commands working fine. No functional change by removing this since all Go commands will get dependencies, but another strike against workspaces, and a weird one.

brandur:
@bgentry With these changes I think we can get the demo app fully switched over to rivershared.

bgentry (approved):

brandur:
thx.

PR #537, opened 2024-08-20
Add CLI version subcommand and flag

brandur:
I was trying to help which River CLI I had installed and realized
there's no easy way to know this currently because it can't currently
reveal its version information in any built-in way.

Here, add a `river version` subcommand that prints the version of the Go
module it's distributed as, along with the version of Go used to build
it, making accessing version information easy.

I also added support for a `river --version` flag. There's no common
standard on whether version should be revealed by subcommand or flag,
and since `--version`'s (unlike `-v`) never going to be used for
anything else, we may as well be permissable in what we accept.

I also add a basic framework for a CLI "integration" test suite that
allows CLI commands to be tested from the level of the Cobra command,
which is useful for checking flags and such. It has the downside in that
there's no way to inject a test transaction into it, so it's not always
appropriate for use right now.

### DIFF
--- a/cmd/river/rivercli/command.go
+++ b/cmd/river/rivercli/command.go
@@ -73,6 +73,7 @@ type RunCommandBundle struct {
 	DatabaseURL    *string
 	DriverProcurer DriverProcurer
 	Logger         *slog.Logger
+	OutStd         io.Writer
 }
 
 // RunCommand bootstraps and runs a River CLI subcommand.
@@ -85,7 +86,7 @@ func RunCommand[TOpts CommandOpts](ctx context.Context, bundle *RunCommandBundle
 		commandBase := &CommandBase{
 			DriverProcurer: bundle.DriverProcurer,
 			Logger:         bundle.Logger,
-			Out:            os.Stdout,
+			Out:            bundle.OutStd,
 		}
 
 		switch {
@@ -124,7 +125,7 @@ func RunCommand[TOpts CommandOpts](ctx context.Context, bundle *RunCommandBundle
 
 	ok, err := procureAndRun()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "failed: %s\n", err)
+		fmt.Fprintf(os.Stdout, "failed: %s\n", err)
 	}
 	if err != nil || !ok {
 		os.Exit(1)
--- a/cmd/river/rivercli/river_cli_test.go
+++ b/cmd/river/rivercli/river_cli_test.go
@@ -3,12 +3,19 @@ package rivercli
 import (
 	"bytes"
 	"context"
+	"fmt"
+	"runtime/debug"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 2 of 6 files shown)


bgentry (approved):

brandur:
Thanks!

Also, I had to tie golangci-lint to 1.60.1. It was upgrading to the new 1.60.2 in CI (not even available yet on Homebrew) which despite a patch-level release, came with a bunch of breaking changes. Ugh.

ccoVeille:
> Thanks!
> 
> Also, I had to tie golangci-lint to 1.60.1. It was upgrading to the new 1.60.2 in CI (not even available yet on Homebrew) which despite a patch-level release, came with a bunch of breaking changes. Ugh.

Yes, golangci-lint 1.60.2 is quite broken.

cmd/river/rivercli/command.go
```diff
@@ -124,7 +125,7 @@ func RunCommand[TOpts CommandOpts](ctx context.Context, bundle *RunCommandBundle
 
 	ok, err := procureAndRun()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "failed: %s\n", err)
+		fmt.Fprintf(os.Stdout, "failed: %s\n", err)
```
ccoVeille, line 128: I'm surprised by this change.

If you had used command.StdOut, I might have understand, but here reporting errors to stdout seems strange.

Maybe I simply missed something about your PR

If it's OK to use stdout here for error reporting, adding an inline comment could help

cmd/river/rivercli/river_cli_test.go
```diff
@@ -69,20 +76,102 @@ var (
...
+
+		cmd.SetArgs([]string{"--version"})
+		require.NoError(t, cmd.Execute())
+
+		buildInfo, _ := debug.ReadBuildInfo()
+
+		require.Equal(t, strings.TrimSpace(fmt.Sprintf(`
+River version (unknown)
+Built with %s
+		`, buildInfo.GoVersion)), strings.TrimSpace(bundle.out.String()))
```
ccoVeille, line 141: I would have used require.Contains to do not link the test and the templating

```suggestion
		require.Contains(t, bundle.out.String(), "River version (unknown)")
		require.Contains(t, bundle.out.String(), "Built with " + buildInfo.GoVersion)
```

cmd/river/rivercli/command.go
brandur: Good catch. That's actually an accident — I was refactoring lots of stuff related to output to find a good way of capturing it for comparison in tests, and this accidentally got hit. Opened #547 with a fix.

cmd/river/rivercli/river_cli_test.go
brandur: Use of `Contains` does seem plausible.  I think this is a little more of a personal preference thing though — I kind of like how when you match the entire output your tests are more strict. i.e. If there was any extra output in there accidentally, they'd catch the problem. This comes with the tradeoff of possibly being a little more brittle in case of minor whitespace changes that aren't really that important to know about.

ccoVeille: OK 👍

cmd/river/rivercli/command.go
ccoVeille: Thanks

Issue #538, opened 2024-08-20
Unable to install the CLI in v0.11.3


youssefsiam38:
after running
```bash
go install github.com/riverqueue/river/cmd/river@latest
```
an error appeared
```
# github.com/riverqueue/river
../../../../go/pkg/mod/github.com/riverqueue/river@v0.11.3/client.go:481:27: undefined: baseservice.NewArchetype
../../../../go/pkg/mod/github.com/riverqueue/river@v0.11.3/client.go:659:30: c.baseStartStop.StoppedUnsafe undefined (type startstop.BaseStartStop has no field or method StoppedUnsafe)
../../../../go/pkg/mod/github.com/riverqueue/river@v0.11.3/client.go:720:23: undefined: startstop.StartAll
```

fixed when I downgraded to v0.11.2

brandur:
Hmm, sorry for the trouble. I thought I'd tested this. Fix incoming.

nexovec:
@brandur You're a legend.

brandur:
Hey, we recently cut v0.11.4, which fixes this issue.

The Go proxy caches for some amount of time. so if `@latest` is still resolving to v0.11.3, then you may need to grab v0.11.4 explicitly:

```
$ go install github.com/riverqueue/river/cmd/river@v0.11.4
```

If that is showing some 404s due to Go proxy caching, then you may need to use env vars to bypass it for now:

```
$ GOPROXY=direct GOSUMDB=off go install github.com/riverqueue/river/cmd/river@v0.11.4
go: downloading github.com/riverqueue/river v0.11.4
go: downloading github.com/riverqueue/river/riverdriver v0.11.4
go: downloading github.com/riverqueue/river/riverdriver/riverpgxv5 v0.11.4
go: downloading github.com/riverqueue/river/rivertype v0.11.4
go: downloading github.com/riverqueue/river/rivershared v0.11.4
```

(None of that should be necessary a half hour or so from now.)

@nexovec Hah, dunno about that, but thanks for the sentiment :)

PR #540, opened 2024-08-20
Update indirect dependencies from `update-mod-version` script

brandur:
This is driven by [1] and the same bug we'd noticed in another project
previously too. Until now, `update-mod-version` hadn't been updating
versions on indirect dependencies, and `rivershared` in the CLI had been
indirect, thereby causing it to try and resolve v0.11.2 instead of
v0.11.3.

Here, update the script so that indirect dependencies are updated as
well.

[1] https://github.com/riverqueue/river/issues/538


bgentry (approved):

brandur:
ty.

PR #543, opened 2024-08-21
Update checksums in `go.work.sum` + tidy everything

brandur:
So this is going to be a small annoyance with workspaces. After cutting
a new release, Go **really** wants to put the new checksums in
`go.work.sum`, and even having the LSP running will update it for you
automatically. If you revert the changes, the LSP will update them
again, which will block you from rebasing and switching branches, which
as you can imagine, is great fun.

So here, add the new checksums to `go.work.sum`, something which we'll
probably have to do after each release goes out.

Also, since we're doing a checksum change anyway, I also ran `make tidy`
to clean up all the package `go.sum` files.


bgentry (approved):

brandur:
ty.

PR #547, opened 2024-08-22
Report CLI failures to stderr, not stdout

brandur:
This corrects a problem from #537, wherein I was doing a lot of
refactoring for testing purposes, and accidentally changed a place where
failures were reported from stderr to stdout.


brandur:
ty.

PR #548, opened 2024-08-22
Upgrade to golangci-lint 1.60.1

brandur:
Here, take the project to golangci-lint 1.60.1, which is now widely
available on Homebrew and elsewhere. It comes with a number of breaking
changes, so I figure it's not the worst idea to keep us on latest so
that future upgrades are less painful.

All breakages are related to overflow checking in `gosec` getting more
strict. The strictness is kind of useful in that it reminds us to check
bounds with something like `min(limit, math.MaxInt32)`, but it can't
detect that a bounds check is done with `min(...)`, which is kind of
annoying because it means you need a `nolint` tag on all of these.

### DIFF
--- a/job_list_params.go
+++ b/job_list_params.go
@@ -327,7 +327,7 @@ func (p *JobListParams) First(count int) *JobListParams {
 		panic("count must be <= 10000")
 	}
 	paramsCopy := p.copy()
-	paramsCopy.paginationCount = int32(count)
+	paramsCopy.paginationCount = int32(count) //nolint:gosec
 	return paramsCopy
 }
 
--- a/queue_list_params.go
+++ b/queue_list_params.go
@@ -35,6 +35,6 @@ func (p *QueueListParams) First(count int) *QueueListParams {
 		panic("count must be <= 10000")
 	}
 	result := p.copy()
-	result.paginationCount = int32(count)
+	result.paginationCount = int32(count) //nolint:gosec
 	return result
 }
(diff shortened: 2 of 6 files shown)


job_list_params.go
```diff
@@ -327,7 +327,7 @@ func (p *JobListParams) First(count int) *JobListParams {
 		panic("count must be <= 10000")
 	}
 	paramsCopy := p.copy()
-	paramsCopy.paginationCount = int32(count)
+	paramsCopy.paginationCount = int32(count) //nolint:gosec
```
brandur, line 330: Bounds are checked a little higher up in this function.

queue_list_params.go
```diff
@@ -35,6 +35,6 @@ func (p *QueueListParams) First(count int) *QueueListParams {
 		panic("count must be <= 10000")
 	}
 	result := p.copy()
-	result.paginationCount = int32(count)
+	result.paginationCount = int32(count) //nolint:gosec
```
brandur, line 38: Bounds are checked a little higher up in this function.

ccoVeille:
A lot of nolint you add are caused by the gosec G115 that was added recently and that are currently being reviewed

I would wait a few days if I was you

Side remark,I thought the gosec G115 was only added in golangci-lint 1.60.2

https://github.com/securego/gosec/pull/1149#issuecomment-2302227236

brandur:
Ah interesting. I'm glad they're reconsidering. There's some overflow check logic that's clearly just wrong (e.g. `uint8 -> int64`).

brandur:
@bgentry Actually, can we move forward with this? Even if some of the gosec stuff is retooled, it's not like golangci-lint 1.60.2 is going to re-released. If they do anything, they'll roll forward to a new 1.6.3, at which point we can remove some `nolint`s if they're no longer necessary. This is basically blocking me from running lint locally because linting always fails on the top level package, so `make lint` can't descend into any of the submodules.

bgentry (approved):
Sweet, good to go. I guess we'd better get the other projects upgrade too then.

brandur:
ty.

Yeah I checked the other projects and a couple fixes will be required, but they're similar to these ones and should only take a couple min.

PR #557, opened 2024-08-23
Preinitialize `workCancel` function in case `StopAndCancel` called before start

### LINKED ISSUE #549: Graceful shutdown - StopAndCancel panics
I'm not sure why `client.StopAndCancel` panics while `client.Stop` returns without any error

```go
package main

import (
	"context"
	"fmt"
	"os"
	"os/signal"
	"sync"
	"syscall"
	"time"

	"github.com/jackc/pgx/v5/pgxpool"
	"github.com/riverqueue/river"
	"riverqueue.com/riverpro/driver/riverpropgxv5"
)

func main() {
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	pool, err := pgxpool.New(ctx, "postgres://postgres:password@localhost:5432/postgres")
	die(err)
	defer pool.Close()

	riverClient, err := river.NewClient(
		riverpropgxv5.New(pool, nil),
		&river.Config{
				river.QueueDefault: {MaxWorkers: 100},
			},
			Workers: river.NewWorkers(),
		},
	)
	die(err)
	riverClient.Start(ctx)
	fmt.Println("started")

	quit := make(chan os.Signal, 1)
	signal.Notify(quit, os.Interrupt, syscall.SIGTERM)
	<-quit
	cancel()

	shutdownCtx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
	defer cancel()
	fmt.Println("shutting down")

	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		// die(riverClient.Stop(shutdownCtx)) // -> OK
		riverClient.StopAndCancel(shutdownCtx) // -> panic
		wg.Done()
	}()
	wg.Wait()
	fmt.Println("shutdown complete")
}

func die(err error) {
	if err == nil {
		return
	}
	panic(err)
}

```

stacktrace
```
started
^Cshutting down
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x829d0d]

goroutine 23 [running]:
github.com/riverqueue/river.(*Client[...]).StopAndCancel(0x0, {0x9ec8e8?, 0xc00017bf10})
        /home/user/pkg/mod/github.com/riverqueue/river@v0.11.4/client.go:838 +0xad
main.main.func1()
        /home/user/src/rift/archipelagoapp/backend/cmd/main.go:51 +0x2f
created by main.main in goroutine 1
        /home/user/src/rift/archipelagoapp/backend/cmd/main.go:49 +0x46e
exit status 2
```

go.mod
```go.mod
module main

go 1.23.0

require (
    github.com/jackc/pgx/v5 v5.6.0
    github.com/riverqueue/river v0.11.4
    riverqueue.com/riverpro/driver/riverpropgxv5 v0.2.1
)

require (
    github.com/davecgh/go-spew v1.1.1 // indirect
    github.com/jackc/pgservicefile v0.0.0-20240606120523-5a60cdf6a761 // indirect
    github.com/pmezard/go-difflib v1.0.0 // indirect
    github.com/riverqueue/river/riverdriver v0.11.4 // indirect
    github.com/riverqueue/river/riverdriver/riverpgxv5 v0.11.4 // indirect
    github.com/riverqueue/river/rivershared v0.11.4 // indirect
    github.com/riverqueue/river/rivertype v0.11.4 // indirect
    github.com/stretchr/testify v1.9.0 // indirect
    github.com/tidwall/gjson v1.17.1 // indirect
    github.com/tidwall/match v1.1.1 // indirect
    github.com/tidwall/pretty v1.2.1 // indirect
    go.uber.org/goleak v1.3.0 // indirect
    golang.org/x/crypto v0.25.0 // indirect
    golang.org/x/sync v0.8.0 // indirect
    golang.org/x/text v0.17.0 // indirect
    gopkg.in/yaml.v3 v3.0.1 // indirect
    riverqueue.com/riverpro v0.2.1 // indirect
    riverqueue.com/riverpro/driver v0.2.1 // indirect
)
```

brandur:
This one's related to #549. Although the most proximate problem in that
repro code is that there was no error check when calling `Start`, it did
reveal a legitimate problem in that the River client will panic in case
`StopAndCancel` is called before `Start` because `workCancel` was never
set.

Here, initialize `workCancel` in the client's constructor. During normal
operation this will be overwritten almost immediately on `Start` as the
client starts up, but in case `Start` was never called or didn't run
successfully, it provides a function for `StopAndCancel` to call so that
it doesn't panic.

Fixes #549.

### DIFF
--- a/client.go
+++ b/client.go
@@ -491,6 +491,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		uniqueInserter: baseservice.Init(archetype, &dbunique.UniqueInserter{
 			AdvisoryLockPrefix: config.AdvisoryLockPrefix,
 		}),
+		workCancel: func(cause error) {}, // replaced on start, but here in case StopAndCancel is called before start up
 	}
 	client.queues = &QueueBundle{addProducer: client.addProducer}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - A new `river migrate-list` command is available which lists available migrations and which version a target database is migrated to. [PR #534](https://github.com/riverqueue/river/pull/534).
 - `river version` or `river --version` now prints River version information. [PR #537](https://github.com/riverqueue/river/pull/537).
 
+## Fixed
+
+- Fixed a panic that'd occur if `StopAndCancel` was invoked before a client was started. [PR #557](https://github.com/riverqueue/river/pull/557).
+
 ## [0.11.4] - 2024-08-20
 
 ### Fixed
--- a/client_test.go
+++ b/client_test.go
@@ -1050,18 +1050,6 @@ func Test_Client_StopAndCancel(t *testing.T) {
 
 		client := runNewTestClient(ctx, t, config)
 
-		insertRes, err := client.Insert(ctx, &callbackArgs{}, nil)
-		require.NoError(t, err)
-
-		startedJobID := riversharedtest.WaitOrTimeout(t, jobStartedChan)
-		require.Equal(t, insertRes.Job.ID, startedJobID)
-
-		select {
-		case <-client.Stopped():
-			t.Fatal("expected client to not be stopped yet")
-		default:
-		}
-
 		return client, &testBundle{
 			jobDoneChan:    jobDoneChan,
 			jobStartedChan: jobStartedChan,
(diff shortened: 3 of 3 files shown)


bgentry:
@brandur oh, probably deserves a changelog entry too

client.go
```diff
@@ -491,6 +491,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		uniqueInserter: baseservice.Init(archetype, &dbunique.UniqueInserter{
 			AdvisoryLockPrefix: config.AdvisoryLockPrefix,
 		}),
+		workCancel: func(cause error) {}, // replaced on start, but here in case StopAndCancel is called before start up
```
ccoVeille, line 494: for such usecase I would define a `defautWorkCanceler := func(error) {}`

Then use it here, it helps to document separately, and using 


```suggestion
		workCancel: defautWorkCanceler // replaced on start
```

brandur: Hm, not sure I agree on that one :) You're never going to reuse this function, and I think adding a named variable for it introduces a layer of indirection that in the end just serves to make it less clear what the code's doing.

brandur:
> @brandur oh, probably deserves a changelog entry too

Added.

Issue #546, opened 2024-08-22
docs: riverpro missing line type workflows -> workflow 


krhubert:
Hi,

When I run cli with `--line workflows` it panics.

```
riverpro migrate-up --database-url  "postgres://postgres:@localhost:5432/postgres" --line workflows
panic: migration line does not exist: workflows
```

But with `--line workflow` all is good

```
riverpro migrate-up --database-url  "postgres://postgres:@localhost:5432/postgres" --line workflow
```

Links with typos

https://riverqueue.com/docs/pro/go-proxy#installing-in-ci-and-build-environments
https://riverqueue.com/docs/pro/getting-started#running-pro-migrations

brandur:
Thanks Hubert!

Wow, believe it or not, we've had multiple people run through those steps already, but you're the first to report the problem to us :/

I just pushed a fix. We should really also have some basic [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) check that'd be able to detect a one letter difference and suggest a correction too. I'll look into that separately.

Please don't hesitate to open any more bug reports if you find other problems!

krhubert:
Sure, I was quite surprised that nobody ran into trouble with this one. Awesome it is fixed so fast. I'll open issues as they raise. I actually have one more already.

brandur:
Going to try and do one other small follow up here to suggest alternatives when a migration line is found to be non-existent in #558.

PR #558, opened 2024-08-23
Better handle non-existent migration line + suggestions with Levenshtein distance

brandur:
This is related to #546. That was a documentation problem, but there are
some code problems that are related.

The first is that an unknown migration line is a panic in the CLI. This
is a byproduct of the fact that `rivermigrate.New` checks the migration
line name, but doesn't return an error, so it'd panic instead.

Here, I'm suggesting that we do a small breaking change by having
`rivermigrate.New` return a possible error. It's not great, but it's
something that'll easy for people to fix, and might help avoid some
future panics. An alternative possibility would be to add a new
`NewWithError` or something of that nature, but it'd make the API a
little uglier.

Along with that change, we also bring in a change to suggest migration
lines in case of an unknown one using Levenshtein distances. This is
aimed at preventing totally unactionable errors in the event of very
simple misspellings (e.g. "workflow" versus "workflows"). Cobra actually
already has a similar feature built-in for command misspellings.

I vendored in this repo [1] as a Levenshtein implementation and did a
little code clean up. I don't want to add another Go module dependency
for such a simple algorithm (it's just a dozen lines), and I believe
their MIT license should be compatible.

[1] https://github.com/agnivade/levenshtein/tree/master

### DIFF
--- a/rivershared/levenshtein/License.txt
+++ b/rivershared/levenshtein/License.txt
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2015 Agniva De Sarker
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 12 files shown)


rivershared/levenshtein/License.txt
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2015 Agniva De Sarker
```
bgentry, line 3: Trying to confirm if this is the correct copyright holder for this bit, where did this name come from?

bgentry (approved):

rivershared/levenshtein/License.txt
brandur: I vendored the code originally from this repo:

https://github.com/agnivade/levenshtein

I'm thinking I may write my own implementation from scratch at some point when I have a bit more time. I ended up rebuilding a lot of the code anyway to modernize and satisfy lint.

brandur: I added a comment pointing to the source repo.

PR #569, opened 2024-08-28
improve InsertTx documentation

bgentry:
I got this feedback from a River Pro customer who mentioned that we had a better explanation of the reasons to use `InsertTx` on the docs website than we had in the godoc. This attempts to bridge that gap by providing a better explanation in the method comments.


brandur (approved):
SGTM.

PR #584, opened 2024-09-09
Add middleware system for jobs

brandur:
Here, experiment with a middleware-like system that adds middleware
functions to job lifecycles, which results in them being invoked during
specific phases of a job like as it's being inserted or worked.

The most obvious unlock for this is telemetry (e.g. logging, metrics),
but it also acts as a building block for features like encrypted jobs.

### DIFF
--- a/client.go
+++ b/client.go
@@ -131,6 +131,10 @@ type Config struct {
 	// If in doubt, leave this property empty.
 	ID string
 
+	// JobMiddleware are optional functions that can be called around different
+	// parts of each job's lifecycle.
+	JobMiddleware []rivertype.JobMiddleware
+
 	// JobTimeout is the maximum amount of time a job is allowed to run before its
 	// context is cancelled. A timeout of zero means JobTimeoutDefault will be
 	// used, whereas a value of -1 means the job's context will not be cancelled
@@ -459,6 +463,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		FetchCooldown:               valutil.ValOrDefault(config.FetchCooldown, FetchCooldownDefault),
 		FetchPollInterval:           valutil.ValOrDefault(config.FetchPollInterval, FetchPollIntervalDefault),
 		ID:                          valutil.ValOrDefaultFunc(config.ID, func() string { return defaultClientID(time.Now().UTC()) }),
+		JobMiddleware:               config.JobMiddleware,
 		JobTimeout:                  valutil.ValOrDefault(config.JobTimeout, JobTimeoutDefault),
 		Logger:                      logger,
 		MaxAttempts:                 valutil.ValOrDefault(config.MaxAttempts, MaxAttemptsDefault),
@@ -1150,7 +1155,7 @@ func (c *Client[TTx]) ID() string {
 	return c.config.ID
 }
 
-func insertParamsFromConfigArgsAndOptions(archetype *baseservice.Archetype, config *Config, args JobArgs, insertOpts *InsertOpts) (*riverdriver.JobInsertFastParams, *dbunique.UniqueOpts, error) {
+func insertParamsFromConfigArgsAndOptions(archetype *baseservice.Archetype, config *Config, args JobArgs, insertOpts *InsertOpts) (*rivertype.JobInsertParams, *dbunique.UniqueOpts, error) {
 	encodedArgs, err := json.Marshal(args)
 	if err != nil {
 		return nil, nil, fmt.Errorf("error marshaling args to JSON: %w", err)
@@ -1212,12 +1217,12 @@ func insertParamsFromConfigArgsAndOptions(archetype *baseservice.Archetype, conf
 		metadata = []byte("{}")
 	}
 
-	insertParams := &riverdriver.JobInsertFastParams{
+	insertParams := &rivertype.JobInsertParams{
 		CreatedAt:   createdAt,
-		EncodedArgs: json.RawMessage(encodedArgs),
+		EncodedArgs: encodedArgs,
 		Kind:        args.Kind(),
 		MaxAttempts: maxAttempts,
-		Metadata:    json.RawMessage(metadata),
+		Metadata:    metadata,
 		Priority:    priority,
 		Queue:       queue,
 		State:       rivertype.JobStateAvailable,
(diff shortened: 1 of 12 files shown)


rivertype/river_type.go
```diff
@@ -218,6 +219,128 @@ type AttemptError struct {
...
+	// and (provided there was no duplicate), inserting it to the database. A
+	// lifecycle hook may make modifications to insert params, but should be
+	// careful doing so as it'll fundamentally modify the job.
+	//
+	// Returns a context that'll be used during the insert operation. Hooks that
+	// don't need to modify context should pass through the input ctx.
+	//
+	// InsertBegin is *not* invoked on a batch insertion with InsertMany or
+	// InsertManyTx. Integrations should implement InsertManyBegin separately.
+	// InsertBegin(ctx context.Context, params *JobLifecycleInsertParams) (context.Context, error)
```
brandur, line 284: So I originally tried to write this system as "hooks" instead of middleware, and you can see what the old entrypoints looked like here (left them in as gravestones for the time being). They almost worked, but I found that I was running into two major problems with them compared to middleware:

* It'd be a common thing to want to be put something in context in the "begin" part, then extract that thing in the "end" part (e.g. for metrics or whatever). I accomplished that by returning a context here as you can see, but it was fairly awkward, and I don't think users would've like it.
* The bigger issue is that it was hard to know how the "end" functions should be called (or not called) under various error conditions like a panic or error returned. We would've either need to send back parameters for all of (1) successful return, (2) possible error, (3) panic val, OR just have totally separate functions for when errors occurred, but both those options were extremely awkward. With middleware, the return values are right there, and it's up to the caller to just do whatever they want with them.

client.go
```diff
@@ -1150,7 +1155,7 @@ func (c *Client[TTx]) ID() string {
 	return c.config.ID
 }
 
-func insertParamsFromConfigArgsAndOptions(archetype *baseservice.Archetype, config *Config, args JobArgs, insertOpts *InsertOpts) (*riverdriver.JobInsertFastParams, *dbunique.UniqueOpts, error) {
```
brandur, line 1153: So not 100% sure on this one yet, but the "insert" part of the middleware needs to receive a type that's _not_ a `JobRow` because we don't have a job row yet. I basically took `JobInsertParams`, duplicated it, and promoted it to `rivertype`. The types are different for now, but they can be type converted to one another because they have the same fields. I basically did this because I like the naming of `JobInsertParams`, but they could also be the same type or even slightly different types with a couple fields dropped (`CreatedAt` for example, which is only needed for time injection).

Either way, `JobInsertParams` stays internal for now, so it should leave refactoring possibility ...

brandur:
@bgentry Wanted to get a minimal POC out there just so I don't spend too much time on this in case there's backpressure, so didn't write tests or anything like that. Rough thoughts?

middleware_defaults.go
```diff
@@ -0,0 +1,47 @@
...
+	return doInner(ctx)
+}
+
+func (l *JobMiddlewareDefaults) InsertMany(ctx context.Context, manyParams []*rivertype.JobInsertParams, doInner func(ctx context.Context) (int, error)) (int, error) {
+	return doInner(ctx)
+}
+
+func (l *JobMiddlewareDefaults) Work(ctx context.Context, job *rivertype.JobRow, doInner func(ctx context.Context) error) error {
+	return doInner(ctx)
+}
```
bgentry, line 25: I think for this to be the most useful it would need to be customizable on a per-job basis. I'm wondering what that looks like in practice with this design. Like what if I wanted to add a middleware that uses some aspect of the worker or args to dynamically determine what to do? (maybe some optional interface gets fulfilled by either of those types to indicate to the middleware what it should do).

The problem with trying to do that here is the args have already been encoded, so there's no longer any access to the underlying `JobArgs` type. Is there any path to potentially having the middleware stack get called _before_ the JSON encoding part? That could more easily enable dynamic behavior based on the type.

Additionally, this might be further exposing the somewhat confusing split between `JobArgs` and `Worker` implementations. We had some recent customer feedback about it being a little weird that i.e. the timeout must be customized on the `Worker` and can't easily be tweaked at insertion time via the args. In this case though you mentioned potentially allowing for middleware to be configured at the `JobArgs` level, which seems fine for insert time but IMO doesn't make any sense for the `Work()` middleware. I don't want to have two separate middleware stacks/concepts, but it does feel a bit odd to have both of these on a single interface given the way this split is designed today 🤔

rivertype/river_type.go
bgentry: Makes sense, I think this is probably the cleaner & more flexible option.

client.go
```diff
@@ -1405,32 +1424,50 @@ func (c *Client[TTx]) InsertManyTx(ctx context.Context, tx TTx, params []InsertM
...
 
-func (c *Client[TTx]) insertFastMany(ctx context.Context, tx riverdriver.ExecutorTx, insertParams []*riverdriver.JobInsertFastParams) (int, error) {
-	inserted, err := tx.JobInsertFastMany(ctx, insertParams)
-	if err != nil {
-		return inserted, err
-	}
+func (c *Client[TTx]) insertFastMany(ctx context.Context, tx riverdriver.ExecutorTx, manyParams []*rivertype.JobInsertParams) (int, error) {
+	doInner := func(ctx context.Context) (int, error) {
+		manyInsertParams := sliceutil.Map(manyParams, func(params *rivertype.JobInsertParams) *riverdriver.JobInsertFastParams {
+			return (*riverdriver.JobInsertFastParams)(params)
```
bgentry, line 1430: I worry a bit about the number of places where we have to do this kind of struct pointer casting and the potential danger of it. A lot of those are in tests and presumably any issues would be caught by tests, but I still felt like I should point out that it makes me feel a little uneasy.

bgentry (commented):
Awesome progress. I had a few thoughts/concerns to consider whether this design may need some tweaks, but I'm generally good to move forward with it.

client.go
bgentry: I guess the compiler actually catches these, so they're nbd?

brandur: Yeah exactly. The structs have to match _exactly_ or the compiler won't let you do it. Even having a non-comparable field in there like a `[]byte` will prevent this.

FWIW, I don't worry two much about two separate structs that are set manually from each other, but I figure as long as they're exactly identical, may as well cast from one to the other.

bgentry:
I rebased this branch on top of #589 now that it's merged. I also refactored our bulk insert methods so they use the same underlying code aside from a narrow adapter function for each query. I think it can be improved further, but IMO it's a good start: https://github.com/riverqueue/river/compare/bg-lifecycle-hooks-on-insert-many-refactor?expand=1

Finally, I had the realization that given the goal of aligning our `Insert` and `InsertMany` APIs to use a single code path (each supporting the same set of features too) that maybe we don't want to introduce a middleware interface that differentiates between those two. I made that change in my above branch as well.

middleware_defaults.go
bgentry: I ended up putting `JobArgs` into the struct in my uniqueness PR and I think it's a great thing to have available. Can still encode args in advance but just keep the original around for introspection.

bgentry:
I know we talked about potentially wanting to have the database transaction available as part of the middleware interface, but I think I've talked myself out of that. With the driver concept it becomes pretty tough to do that cleanly, especially given we don't want the driver interfaces to be considered stable. I also don't think it's needed for anything I'm doing at the moment (#627 seems like the way for me to do database-related customizations).

bgentry:
A version of this was merged in #632 and will be in the next release.

PR #587, opened 2024-09-09
move testfactory to rivershared

bgentry:
This will be useful for tests in other River libraries (riverui, riverpro).

The `replace` in here is necessary to avoid issues with `go mod tidy` trying to fetch the `rivershared` module from remote and it not containing the `testfactory` package. I believe we should be able to simply remove it prior to release, but you may have more insight on that given the workspace testing you did.


brandur (approved):
Damn that `replace` thing is absurd. (Odd that I don't remember running into this at all during development, but I guess it's because I never ran `go mod tidy`, and nothing complained that I should.)

Yet another run on Go workspace thread on the problem with no clear resolution in sight:

https://github.com/golang/go/issues/50750

PR #590, opened 2024-09-15
Bulk unique insertion, uniqueness with subset of args

bgentry:
## Background

Our first two attempts at unique insertions have been limited:

1. The initial version relied on advisory locks, which were slow enough and had enough potential for conflicts that we restricted them to the single job insertion path. They were fairly flexible in terms of allowing a full customization of which states should be considered in the uniqueness check.
2. The [second version of unique jobs](https://github.com/riverqueue/river/pull/451) brought us _significantly_ improved performance using a database unique index on the hash of a computed unique key. However, it came with less flexibility: that code path would only be used when the user did not customize the list of unique states beyond our default set of states (everything but `cancelled` and `discarded`).

   River queries were hardcoded to clear out the `unique_key` value if the job transitioned to a designated state (ex. [1](https://github.com/riverqueue/river/blob/67c8a8af01e43672e11a910c33ac5734454ad8d0/riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql#L70-L72) [2](https://github.com/riverqueue/river/blob/67c8a8af01e43672e11a910c33ac5734454ad8d0/riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql#L476-L477)). If a user wanted to ensure that a job is never duplicated _even if_ there's already an identical job in either of those unsupported states, River would fall back to the slower/older advisory lock uniqueness path.
   
   This feature also was only available when inserting one job at a time, and because its unique index was prefixed with `kind` it couldn't be used across job kinds.

## An improved approach

In this PR, I've proofed out an improved design for unique jobs that I believe addresses all of the shortcomings of the previous designs, maintaining full flexibility while also maintaining the performance improvements of the database-level unique constraint design. The new design avoids design 2's need for a hardcoded list of allowed states by moving the list of unique states into the database on a per-row basis.

I initially looked at doing this with a `river_job_state[]` column, but that was both verbose and not particularly space-efficient. Instead I landed on a `bit(8)` bitmask column, which is included as part of the compound partial index for uniqueness. A custom SQL function allows that unique index to only apply to jobs when their state is within the allowed states as reflected in the bitmask, so when they are moved to another state they are also no longer considered in uniqueness guarantees. (Side note: GPT-o1 was actually quite helpful in ironing out some aspects of this design and helping me to reason about whether I had considered everything. It's super impressive ✨)

Importantly, this design does not rely on advisory locks in any way, and in this PR the new path has been expanded to also work on the bulk insertion paths. Additionally, if for some reason users wish to ensure uniqueness across different job kinds, they can do that here using the `ExcludeKind` field on `UniqueOpts`. `kind` is no longer part of the new unique index to allow for this flexibility.

Another nicety: this gets us very close to having a unified code path for all inserts without any caveats depending on how the job was inserted. 🔜 We should be able to get there as soon as we remove the old deprecated unique code paths.

### Restrictions on state

This approach _can_ be made to work for a totally custom list of states. However, there's some complexity and undefined edge cases with doing that, particularly for allowing people to remove `pending`, `scheduled`, `available`, or `running` from the list. To reason this through, I prepared a mermaid state transition diagram (included in PR) which I believe illustrates all possible job state transitions. On a printed version of this I was able to categorize all the state changes to note which queries are responsible for them.

The analysis of this is verbose, but if you're intrigued feel free to read it through.

<details>
<summary>
Detailed analysis of states list and reasons for customizing
</summary>

## Default unique states

Since we first introduced unique jobs, this has been the default list of states that are checked for uniqueness:

* Pending
* Available
* Scheduled
* Running
* Retryable
* Completed

These states ensure that uniqueness is enforced unless a job has been explicitly cancelled or has exhausted its errors.

The omission of `discarded` and `cancelled` makes sense as a default (those states indicate something has gone wrong and maybe you want to be able to retry with a new job). However they are not the only scenario.

## Reasons for customizing this list

For the various states, there are reasons someone might want to add or remove a state from the list. I’ll go through each state here and talk about why someone might want to add/remove it from the list, or what complications that would introduce.

### Pending

Included by default. I can't think of a use case that would benefit from removing this from the list.

If you remove it, we could have trouble moving jobs out of `pending` into `available` or `scheduled` (if they’re enabled). Would need to have those workflow/sequence queries be able to skip rows that have conflicts, but they would just keep conflicting and being skipped on every run.

### Scheduled

Included by default. I can't think of a use case that would benefit from removing this from the list.

Complicated to allow removal. If you allow multiple dupes to be in `scheduled` state, but block them from proceeding to `available`, then our scheduler maintenance process (or workflow scheduler) wouldn't be able to advance those jobs when it's time for them to run.

### Available

Included by default. The main use case I can think of to remove this and earlier states from the unique list is to facilitate a poor man's version of global concurrency limiting (limited to 1). 

Dangerous to remove, because we can get multiple jobs in the available state but then only one of them can move to `running` without a conflict. Allowing this that will put a lot of stress on the job fetch query, which is the absolute last place we want to deal with repeated conflicts and row rewrites.

### Running

Included by default.  I can't think of a use case that would benefit from removing this from the list.

It's dangerous to remove, because then you can insert and start working duplicate jobs, only to be unable to complete/retry/discad them (unless those other states are removed also).

### Retryable

Included by default. A user may wish to allow a job to try once and may want to enforce uniqueness up until the point at which it has errored once, at which case they could want to allow dupes.

If removed, the scheduler may have trouble moving back to `available`. That could be dealt with by having the scheduler query mark the row as discarded if it can’t be moved back to `available` due to a unique conflict.

### Completed

Included by default. Similar to `cancelled` and `discarded` it may be desirable to remove this if you want to ensure there is only one “live” job at a time in some non-finalized state.

No major downsides to removing it.

### Cancelled (opt in)

Use case: I want to ensure that there is only one job of a given kind for a timeframe or set of args, even if it gets cancelled.

If you allow earlier states to be removed from the list, the cancellation and completer queries would need to handle this situation somehow and I'm not sure there's a "right" answer for what to do in that case.

### Discarded (opt in)

Use case: I want to ensure that there is only one job of a given kind for a timeframe or set of args, even if it runs out of errors. This makes sure that job is never ever duplicated.

If you allow earlier states to be removed from the list, the cancellation and completer queries would need to handle this situation somehow and I'm not sure there's a "right" answer for what to do in that case.

## A solution?

Most problems are resolved if with one rule: don’t allow removing the `pending`, `available`, `scheduled`, and `running` states from the list. Here’s an analysis of the other states given this rule:

If one removes `retryable`, there can be dupes in the `retryable` state that _can’t_ all be moved back to another state. Manual `JobRetry` can just error in this case, whereas automated scheduling of retries can move the job to `discarded` if a conflict occurs.

If one adds `cancelled` to the default list, that does not have any adverse impact, because there can be no dupes in any of the default states _plus_ `cancelled`. You will always still be able to retry that cancelled job, because none of the states it will go into can be occupied by a duplicate job.

Likewise, adding `discarded` is not an issue for the same reasons.

Finally, there’s no issue with removing `completed`, other than somebody being blocked from manually retrying an already-completed job. In that case we can just bubble up the error to the user.
</details>

In short, I think it would make sense to restrict users from removing these 4 initial states from their `UniqueOpts.ByState`. I'm not sure of the best UX for that. We could simply error when you try to insert a job that breaks that rule (at least when using the multi insert / non-advisory code path). Or we could use a different way of customizing states besides a list.

## Bonus: opt in only specific args to uniqueness

Since it was first introduced, we've heard from users who want to consider only a subset of their job args when determining uniqueness. For example, I might want to make sure there is exactly one job of a particular type for a given `customer_id`, even though there may also be other args on the job. Once I had proofed out the new design, I scratched an itch to spike out this feature.

As implemented here, when a job has been opted into uniqueness via `UniqueOpts` with `ByArgs`, it can optionally add a `river:"unique"` struct tag to specific fields that the user wants to be considered in uniqueness. For example, consider this struct:

```go
type EmailJobArgs struct {
	EmailID string `json:"email_id" river:"unique"`
	Recipient string `json:"recipient" river:"unique"`
	Subject   string `json:"subject"`
	Body      string `json:"body"`
}
```

For unique jobs of this type, the uniqueness constraint will be applied on the following json:

```json
{"email_id": 12345,"recipient":"john@example.com"}
```

JSON field names are respected as well if customized in the struct tags, and fields that are omitted from the encoded JSON will also be omitted from the uniqueness JSON. These arg keys are sorted alphabetically before being built into the uniqueness JSON to ensure a consistent result.

## Transition Plan

While proofing out this design, I decided to avoid getting slowed down with legacy compatibility and stripped out the old uniqueness systems entirely. However I don't think we can do just merge it without considering how existing apps will transition between the systems, ideally without any downtime.

Also, the unique key calculation is necessarily different here than in the v2 design, because `kind` is now a part of the unique key instead of part of the index. So we can't easily migrate in-flight jobs from the v2 design to the v3 design. And v1 unique jobs don't even have a `unique_key` populated at all, so they can't be transitioned cleanly either.

I have some ideas:

* Maintain both the v1 and v2 code paths temporarily (say, for one final `v0.x.y` release series).
* Retain the bits of SQL queries which clear the `unique_key` value when cancelling or discarding jobs, but _only_ do so for rows with `unique_states IS NULL`. This will allow v2 unique jobs to continue being processed in the same way.
* For new jobs which specify a list of unique states that don't include all the required ones shown above (`pending`, `scheduled`, `available`, `running`), continue to fall back to the v1 design but log a deprecation warning. Make sure this legacy mode only works for the single-job insert paths and is explicitly blocked/errored when using `InsertMany`.
* New jobs with a compatible set of states will use the v3 path and have both `unique_key` and `unique_states` populated. No new jobs will be inserted with v2 mode because v3 is strictly a superset of it.

## Remaining issues

* [x] Decide how to implement the restricted list of unique states and refine its UX
* [x] Query changes to fix scheduler conflicts if `retryable` is removed from unique states (job won't be able to retry and must be discarded). Need tests for this too.
* [x] Implementation changes to account for migration plan (restore deleted code)
* [x] Rename `unique:"true"` struct tag to `river:"unique"` for cleaner, more forward-compatible API design
* [x] Tests to ensure that jobs using the old unique designs continue to function, for now.
* [x] Go documentation
* [ ] Website documentation
* [x] Consistent alphabetical sort order for all encoded args (or else reordering struct fields causes uniqueness changes)
* [x] Migration framework and tests don't play nicely with this latest migration. The reason is that migration 4 adds a `pending` state to the `river_job_state` enum, and this latest migration 6 attempts to use that value within an immutable SQL function. This results in the following error:
   
   ```
   ERROR: unsafe use of new value "pending" of enum type river_job_state (SQLSTATE 55P04)
   ```
   
   The cause of this is that our current migration system and its tests will attempt to run _all_ available migrations in a single transaction without committing between each one. I believe this might be part of why other migration systems (like Rails/ActiveRecord) use a separate transaction for each individual migration rather than running them all in one.

   We _may_ be able to solve this for compatible databases by opening a DDL sub transaction for each individual migration (as I've done in this PR to get tests passing and `testdbman` to work) but we should resolve that in a standalone PR before this one gets merged. In particular the one test still failing in this PR is one that tries to roll back and then re-run all migrations within a single test transaction. ~Maybe you want to take a crack at a standalone PR for migration framework changes @brandur?~ Fixed in #600.


rivermigrate/river_migrate.go
```diff
@@ -560,10 +559,20 @@ func (m *Migrator[TTx]) applyMigrations(ctx context.Context, exec riverdriver.Ex
...
+				if err != nil {
+					return fmt.Errorf("error applying version %03d [%s]: %w",
+						versionBundle.Version, strings.ToUpper(string(direction)), err)
+				}
+				return nil
+			})
 			if err != nil {
-				return nil, fmt.Errorf("error applying version %03d [%s]: %w",
-					versionBundle.Version, strings.ToUpper(string(direction)), err)
+				return nil, err
```
bgentry, line 575: The changes in this file were fairly brute force just to unblock me fixing the rest of the tests, so I haven't spent much time thinking about the right changes to make here. For now it's opening a sub-transaction for each individual migration. This works enough for nearly all our tests and for the `testdbman` tool to once again function, but it may not work with other DBs.

rivertype/river_type.go
```diff
@@ -17,6 +17,15 @@ var ErrNotFound = errors.New("not found")
...
 var ErrJobRunning = errors.New("running jobs cannot be deleted")
 
+// JobArgs is an interface that should be implemented by the arguments to a job.
+// This definition duplicates the JobArgs interface in the river package so that
+// it can be used in other packages without creating a circular dependency.
+type JobArgs interface {
+	// Kind returns a unique string that identifies the type of job. It's used to
+	// determine which worker should work the job.
+	Kind() string
+}
```
bgentry, line 27: Needed to use this from `dbunique` so I copied the interface into here. But then I ended up adding an unexported `jobArgs` within that package anyway—maybe that's better than exposing a duplicate in a user-facing package like this?

internal/dbunique/unique_fields.go
```diff
@@ -0,0 +1,123 @@
...
+var (
+	// uniqueFieldsCache caches the unique fields for each JobArgs type. These are
+	// global to ensure that each struct type's tags are only extracted once.
+	uniqueFieldsCache = make(map[reflect.Type][]string) //nolint:gochecknoglobals
+	cacheMutex        sync.RWMutex                      //nolint:gochecknoglobals
+)
+
+type jobArgs interface {
+	Kind() string
+}
```
bgentry, line 23: Oops, don't need both this _and_ the `rivertype.JobArgs` duplicate definition. Lmk if you think this unexported copy is the right way to go.

go.mod
```diff
@@ -15,6 +15,8 @@ require (
 	github.com/riverqueue/river/rivertype v0.11.4
 	github.com/robfig/cron/v3 v3.0.1
 	github.com/stretchr/testify v1.9.0
+	github.com/tidwall/gjson v1.17.3
+	github.com/tidwall/sjson v1.2.5
```
bgentry, line 19: `gjson` and `sjson` were super handy for being able to efficiently extract a subset of keys from a JSON payload and to reconstruct them into a new sorted subset payload. We could figure out a way to do this without them, but it would likely be slower (if using `encoding/json`) or a lot more verbose (if building up our own JSON bytes).

They're also super small dependencies with minimal indirects, so I feel like it's not a big deal to pull them in. Plus they're also used in riverpro.

client.go
```diff
@@ -53,24 +53,6 @@ const (
...
-	// the entire 64-bit advisory lock number space, which is large enough that
-	// conflicts are exceedingly unlikely. If callers don't strictly need this
-	// option then it's recommended to leave it unset because the prefix leaves
-	// only 32 bits of number space for advisory lock hashes, so it makes
-	// internally conflicting River-generated keys more likely.
-	//
-	// Advisory locks are currently only used for the fallback/slow path of
-	// unique job insertion where finalized states are included in a ByState
-	// configuration.
-	AdvisoryLockPrefix int32
```
bgentry, line 72: Once we complete the deprecation and eventual removal of the advisory lock based unique jobs, would you want to just nuke this config altogether? It's possible we could want to use advisory locks again at some point but it'd be easy enough to re-add.

internal/dbunique/db_unique.go
```diff
@@ -32,195 +28,127 @@ var defaultUniqueStates = []rivertype.JobState{ //nolint:gochecknoglobals
...
+var jobStateBitPositions = map[rivertype.JobState]uint{ //nolint:gochecknoglobals
+	rivertype.JobStateAvailable: 7,
+	rivertype.JobStateCancelled: 6,
+	rivertype.JobStateCompleted: 5,
+	rivertype.JobStateDiscarded: 4,
+	rivertype.JobStatePending:   3,
+	rivertype.JobStateRetryable: 2,
+	rivertype.JobStateRunning:   1,
+	rivertype.JobStateScheduled: 0,
+}
```
bgentry, line 40: These numbers correspond to the postgres bitshift numbers, which go from left to right. That's why the corresponding Go bit shifting code below shifts by `(7 - postgres_index)` when building up the bitmask.

internal/dbunique/db_unique.go
```diff
@@ -32,195 +28,127 @@ var defaultUniqueStates = []rivertype.JobState{ //nolint:gochecknoglobals
...
+		encodedArgsForUnique := params.EncodedArgs
+
+		{
+			// Get unique JSON keys from the JobArgs struct:
+			uniqueFields, err := getSortedUniqueFieldsCached(params.Args)
+			if err != nil {
+				// Instead of returning an error, we can just proceed using the entire
+				// encoded args.
+				//
+				// TODO: log or how will we find out about this?
```
bgentry, line 92: This is one of those "should never happen" errors. I believe it can only happen if you give us a non-struct `JobArgs`. Is that even possible or does it happen in the wild? 🤷‍♂️

internal/dbunique/db_unique.go
```diff
@@ -32,195 +28,127 @@ var defaultUniqueStates = []rivertype.JobState{ //nolint:gochecknoglobals
...
+			sortedJSONWithOnlyUniqueValues := make([]byte, 0, len(params.EncodedArgs))
+
+			sjsonOpts := &sjson.Options{ReplaceInPlace: true}
+			for i, key := range uniqueFields {
+				if uniqueValues[i] == "undefined" {
+					continue
+				}
+				sortedJSONWithOnlyUniqueValues, err = sjson.SetRawBytesOptions(sortedJSONWithOnlyUniqueValues, key, []byte(uniqueValues[i]), sjsonOpts)
+				if err != nil {
+					// Should not happen unless key was invalid
```
bgentry, line 116: The only documented reason for this error to be non-nil is if the json key was invalid.

internal/dbunique/unique_fields.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// Sort the uniqueFields alphabetically for consistent ordering
+	sort.Strings(uniqueFields)
+
+	return uniqueFields, nil
+}
+
+// getSortedUniqueFieldsCached retrieves unique fields with caching to avoid
+// extracting fields from the same struct type repeatedly.
+func getSortedUniqueFieldsCached(args rivertype.JobArgs) ([]string, error) {
```
bgentry, line 90: Similar to `encoding/json`, this cache allows a given struct type's tags to only need to be parsed once per program run instead of for every struct instance.

internal/riverinternaltest/riverdrivertest/riverdrivertest.go
```diff
@@ -1229,110 +1237,110 @@ func Exercise[TTx any](ctx context.Context, t *testing.T,
...
+
+	// 	t.Run("AllArgs", func(t *testing.T) {
+	// 		t.Parallel()
+
+	// 		exec, _ := setup(ctx, t)
+
+	// 		targetTime := time.Now().UTC().Add(-15 * time.Minute)
+
+	// 		insertRes, err := exec.JobInsertUnique(ctx, &riverdriver.JobInsertUniqueParams{
+	// 			JobInsertFastParams: &riverdriver.JobInsertFastParams{
```
bgentry, line 1287: Oops, unfinished conversion.

riverdriver/river_driver_interface.go
```diff
@@ -115,11 +115,10 @@ type Executor interface {
 	JobGetByKindAndUniqueProperties(ctx context.Context, params *JobGetByKindAndUniquePropertiesParams) (*rivertype.JobRow, error)
 	JobGetByKindMany(ctx context.Context, kind []string) ([]*rivertype.JobRow, error)
 	JobGetStuck(ctx context.Context, params *JobGetStuckParams) ([]*rivertype.JobRow, error)
-	JobInsertFast(ctx context.Context, params *JobInsertFastParams) (*rivertype.JobRow, error)
-	JobInsertFastMany(ctx context.Context, params []*JobInsertFastParams) ([]*rivertype.JobRow, error)
+	JobInsertFast(ctx context.Context, params *JobInsertFastParams) (*JobInsertFastResult, error)
+	JobInsertFastMany(ctx context.Context, params []*JobInsertFastParams) ([]*JobInsertFastResult, error)
```
bgentry, line 119: This relates to the API changes from #584 so we may want to align on the right change to make here.

riverdriver/river_driver_interface.go
```diff
@@ -249,44 +248,46 @@ type JobGetStuckParams struct {
...
-type JobInsertUniqueParams struct {
-	*JobInsertFastParams
-	UniqueKey []byte
-}
-
-type JobInsertUniqueResult struct {
+	// Args contains the raw underlying job arguments struct. It has already been
+	// encoded into EncodedArgs, but the original is kept here for to leverage its
+	// struct tags and interfaces, such as for use in unique key generation.
+	Args         rivertype.JobArgs
```
bgentry, line 254: Related to discussion from #584, this was added to facilitate tag reflection on the `JobArgs` struct, even though it's otherwise unused (`EncodedArgs` are already encoded at this point).

riverdriver/riverdatabasesql/go.mod
```diff
@@ -5,6 +5,7 @@ go 1.21
 toolchain go1.23.0
 
 require (
+	github.com/jackc/pgx/v5 v5.7.1
```
bgentry, line 8: Hmm, I guess this is what I get for using `pgtype.bits` for the bitmask column. I'd be shocked if pq properly supports that, thoughts on what to do?

riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
```diff
@@ -220,8 +223,17 @@ INSERT INTO river_job(
...
-    coalesce(@tags::varchar(255)[], '{}')
-) RETURNING *;
+    coalesce(@tags::varchar(255)[], '{}'),
+    @unique_key,
+    @unique_states
+)
+ON CONFLICT (unique_key)
+    WHERE unique_key IS NOT NULL
+      AND unique_states IS NOT NULL
+      AND river_job_state_in_bitmask(unique_states, state)
```
bgentry, line 233: I wish there were a better way but AFAICT this is the only way to indicate the conflict given the specific partial index in use here.

riverdriver/riverpgxv5/migration/main/006_bulk_unique.up.sql
```diff
@@ -0,0 +1,38 @@
...
+AS $$
+    SELECT CASE state
+        WHEN 'available' THEN get_bit(bitmask, 7)
+        WHEN 'cancelled' THEN get_bit(bitmask, 6)
+        WHEN 'completed' THEN get_bit(bitmask, 5)
+        WHEN 'discarded' THEN get_bit(bitmask, 4)
+        WHEN 'pending' THEN get_bit(bitmask, 3)
+        WHEN 'retryable' THEN get_bit(bitmask, 2)
+        WHEN 'running' THEN get_bit(bitmask, 1)
+        WHEN 'scheduled' THEN get_bit(bitmask, 0)
```
bgentry, line 15: Per the note on the Go bitmask code, Postgres bit positions are numbered left-to-right instead of a more conventional right-to-left.

rivermigrate/river_migrate.go
bgentry: Extracted to #600.

client_test.go
```diff
@@ -5167,40 +5179,150 @@ func TestInsertParamsFromJobArgsAndOptions(t *testing.T) {
...
+
+	t.Run("UniqueOptsWithPartialArgs", func(t *testing.T) {
+		t.Parallel()
+
+		uniqueOpts := UniqueOpts{ByArgs: true}
+
+		type PartialArgs struct {
+			JobArgsStaticKind
+			Included bool `json:"included" unique:"true"`
+			Excluded bool `json:"excluded"`
```
bgentry, line 5302: I had another design idea on this. Instead of `unique:"true"`, what if this was something like `river:"unique"`? There would be room to apply additional comma-separated options in that list, such as for batching or sequence keys.

I think I like that more than having a bunch of separate tags with vague names like `unique`, `batch`, etc. Those could be prefixed with `river_` of course but that'd be long and redundant, plus not a particularly efficient way to declare boolean attrs.

bgentry: Changed to `river:"unique"` so better & more cleanly support future options. Seems less vague/generic too which is nice.

internal/maintenance/job_scheduler_test.go
```diff
@@ -139,6 +148,50 @@ func TestJobScheduler(t *testing.T) {
 		requireJobStateUnchanged(t, bundle.exec, retryableJob3) // still retryable
 	})
 
+	t.Run("MovesUniqueKeyConflictingJobsToDiscarded", func(t *testing.T) {
```
bgentry, line 151: I realized I don't have corresponding test changes for the underlying query in the `riverdrivertest`, lmk if you think these details should be duplicated over there.

riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
```diff
@@ -396,30 +406,49 @@ FROM updated_job;
...
+    WHERE river_job.id = jts.id
+    RETURNING river_job.id, state = 'discarded'::river_job_state AS conflict_discarded
 )
-SELECT *
+SELECT
+    sqlc.embed(river_job),
+    updated_jobs.conflict_discarded
 FROM river_job
-WHERE id IN (SELECT id FROM river_job_scheduled);
+JOIN updated_jobs ON river_job.id = updated_jobs.id;
```
bgentry, line 451: Ok, so this is one way of handling the scenario where somebody removes `retryable` from their `ByStates` list, and then inserts a new job while there's still a would-be dupe in the `retryable` state. Essentially I tried to make this one query able to detect and handle that case, instead moving the conflicting jobs to be `discarded` instead of `retryable`.

Two questions:

1. Instead of doing this within a larger query like this, we _could_ attempt to detect this conflict in application code and make follow up queries to (a) clear out/update/discard the conflicting jobs separately, and (b) try the `JobSchedule` query again.
2. Should this set a metadata value on the discarded job so the transition is detectable?

Also open to other ways to handle this. If we wanted to allow other states to be removed from that list, it would involve similar handling and needing to define what the behavior would be when such a conflict arises.

bgentry: Added a metadata value of `"unique_key_confict": "scheduler_discarded"` so we can at least see when this case has been triggered.

bgentry:
I believe this is all set. I've updated the changelog with notes on how this PR impacts users and what to watch out for during migration. Going to merge it and aim for a `v0.12.0-rc.1` prerelease to shake out any issues.

PR #599, opened 2024-09-18
Improve error when `Client.Subscribe` called on a client that will not work

### LINKED ISSUE #596: Unclear panic when using Client.Subscribe() when client is not executing jobs
If I try to subscribe for job updates: 

    rc, err := river.NewClient(driver, &river.Config{})
    if err != nil {
        panic(err)
    }
    rc.Subscribe(river.EventKindJobCancelled, river.EventKindJobCompleted, river.EventKindJobFailed)

I get a panic:

    panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x1 addr=0x88 pc=0x201a3e0]
    
    goroutine 1 [running]:
    github.com/riverqueue/river.(*subscriptionManager).SubscribeConfig(0x0, 0xc00038d5a0)
            /home/martin/.cache/go/pkg/mod/github.com/riverqueue/river@v0.11.3/subscription_manager.go:223 +0x100
    github.com/riverqueue/river.(*Client[...]).SubscribeConfig(...)
            /home/martin/.cache/go/pkg/mod/github.com/riverqueue/river@v0.11.3/client.go:908
    github.com/riverqueue/river.(*Client[...]).Subscribe(0xc0001b6020?, {0xc00038d670?, 0x1?, 0x1?})
            /home/martin/.cache/go/pkg/mod/github.com/riverqueue/river@v0.11.3/client.go:881 +0x47

The reason for this is that in `river.NewClient()` it does (abdriged):

	// them):
	if config.willExecuteJobs() {
		// ...
		client.subscriptionManager = newSubscriptionManager(archetype, nil)
		// ...
	}

Which is reasonable and fair enough. I would expect a nicer message than a panic and having to read the source code to figure out why I get a panic. There isn't any error return, but just panic() with a nicer message is enough (this is clearly a programmer error, so it's probably an appropriate usage of panic).

brandur:
Related to #596. If `Subscribe` was called on a client that didn't have
a `Workers` bundle configure a nil pointer panic would occur because
`subscriptionManager` was never initialized.

Here, leave that as a panic since it makes sense to warn a user about an
API misuse that'd undoubtedly lead to more confusion/pain, but improve
the error message so that it's more obvious to the caller why this is a
problem.

Fixes #596.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed a panic that'd occur if `StopAndCancel` was invoked before a client was started. [PR #557](https://github.com/riverqueue/river/pull/557).
 - A `PeriodicJobConstructor` should be able to return `nil` `JobArgs` if it wishes to not have any job inserted. However, this was either never working or was broken at some point. It's now fixed. Thanks [@semanser](https://github.com/semanser)! [PR #572](https://github.com/riverqueue/river/pull/572).
+- Fixed a nil pointer exception if `Client.Subscribe` was called when the client had no configured workers (it still, panics with a more instructive error message now). [PR #599](https://github.com/riverqueue/river/pull/599).
 
 ## [0.11.4] - 2024-08-20
 
--- a/client.go
+++ b/client.go
@@ -906,6 +906,10 @@ type SubscribeConfig struct {
 
 // Special internal variant that lets us inject an overridden size.
 func (c *Client[TTx]) SubscribeConfig(config *SubscribeConfig) (<-chan *Event, func()) {
+	if c.subscriptionManager == nil {
+		panic("created a subscription on a client that will never work jobs (Workers not configured)")
+	}
+
 	return c.subscriptionManager.SubscribeConfig(config)
 }
 
--- a/client_test.go
+++ b/client_test.go
@@ -3893,6 +3893,19 @@ func Test_Client_Subscribe(t *testing.T) {
 
 		require.Empty(t, client.subscriptionManager.subscriptions)
 	})
+
+	// Just make sure this doesn't fail on a nil pointer exception.
+	t.Run("SubscribeOnClientWithoutWorkers", func(t *testing.T) {
+		t.Parallel()
+
+		dbPool := riverinternaltest.TestDB(ctx, t)
+
+		client := newTestClient(t, dbPool, &Config{})
+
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed a panic that'd occur if `StopAndCancel` was invoked before a client was started. [PR #557](https://github.com/riverqueue/river/pull/557).
 - A `PeriodicJobConstructor` should be able to return `nil` `JobArgs` if it wishes to not have any job inserted. However, this was either never working or was broken at some point. It's now fixed. Thanks [@semanser](https://github.com/semanser)! [PR #572](https://github.com/riverqueue/river/pull/572).
+- Fixed a nil pointer exception if `Client.Subscribe` was called when the client had no configured workers (it still, panics with a more instructive error message now). [PR #XXX](https://github.com/riverqueue/river/pull/XXX).
```
bgentry, line 43:
```suggestion
- Fixed a nil pointer exception if `Client.Subscribe` was called when the client had no configured workers (it still, panics with a more instructive error message now). [PR #599](https://github.com/riverqueue/river/pull/599).
```

client.go
```diff
@@ -906,6 +906,10 @@ type SubscribeConfig struct {
 
 // Special internal variant that lets us inject an overridden size.
 func (c *Client[TTx]) SubscribeConfig(config *SubscribeConfig) (<-chan *Event, func()) {
+	if c.subscriptionManager == nil {
+		panic("created a subscription on a client that will never work jobs (Workers not configure)")
```
bgentry, line 910:
```suggestion
		panic("created a subscription on a client that will never work jobs (Workers not configured)")
```

client_test.go
```diff
@@ -3893,6 +3893,19 @@ func Test_Client_Subscribe(t *testing.T) {
...
+
+	// Just make sure this doesn't fail on a nil pointer exception.
+	t.Run("SubscribeOnClientWithoutWorkers", func(t *testing.T) {
+		t.Parallel()
+
+		dbPool := riverinternaltest.TestDB(ctx, t)
+
+		client := newTestClient(t, dbPool, &Config{})
+
+		require.PanicsWithValue(t, "created a subscription on a client that will never work jobs (Workers not configure)", func() {
```
bgentry, line 3905:
```suggestion
		require.PanicsWithValue(t, "created a subscription on a client that will never work jobs (Workers not configured)", func() {
```

bgentry (approved):

brandur:
ty.

PR #609, opened 2024-09-23
InsertManyFast doc fixes for uniqueness

bgentry:
I missed this documentation change in #590. `InsertManyFast` / `InsertManyFastTx` _do_ support unique jobs now, just not as gracefully as the other `InsertMany` methods.


bgentry:
Minor doc fix I separated out for visibility, but I don't think it needs to block on review. Feel free to follow up with tweaks/suggestions/objections/etc :v:

PR #613, opened 2024-09-23
run make tidy

bgentry:
I'm not sure if there's something missing in the release instructions, but it doesn't seem like anything other than `cmd/river` ever gets `go mod tidy` run on it.


riverdriver/riverdatabasesql/go.mod
```diff
@@ -7,6 +7,7 @@ toolchain go1.23.0
 require (
 	github.com/jackc/pgx/v5 v5.7.1
 	github.com/lib/pq v1.10.9
+	github.com/riverqueue/river v0.12.0
```
bgentry, line 10: @brandur hmm, I think I've created a cycle here? Both drivers are now calling `dbunique.UniqueBitmaskToStates()` in `internal/dbunique`, which has made them both take a dependency on the main river package. And the main package depends on the drivers too due to testing:

https://github.com/riverqueue/river/blob/5e3533556d8e6fc77ae11ce36577113055a82434/go.mod#L12-L13

bgentry:
Still not quite sure if this is missing from one of the release steps (it's only listed under CLI release) but it seems necessary to get it merged so I'm doing that now. Lmk your thoughts :v:

PR #614, opened 2024-09-24
remove deprecated advisory lock uniqueness, consolidate insert logic

bgentry:
This removes the original unique jobs implementation in its entirety. It was already deprecated in the previous release.

🗒️ I'm not quite ready to merge this yet (I'd like to get the Ruby + Python clients to ship the new unique jobs changes first) but I still wanted to prepare it while the context is fresh. It's also needed to simplify some upcoming Pro changes.

All known use cases are better supported with the new unique jobs implementation which is also dramatically faster and supports batch insertion.

As part of this change, single insertions now inherit the behavior of batch insertions as far as always setting a `scheduled_at` time in the job args prior to hitting the database. This is due to the difficulty of trying to pass an array of nullable timestamps for `scheduled_at` to the database using sqlc. One side effect of this is that some tests needed to be updated because they run in a transaction, which locks in a particular `now()` time used in `JobGetAvailble` by default. Jobs inserted _after_ the start of that transaction would pick up a scheduled timestamp from Go code that is _later_ than the database transaction's timestamp, and so those jobs would never run. This was fixed in some cases by allowing `now` to be overridden by lower level callers of that query, whereas other tests were updated to simply insert jobs with a past `scheduled_at` to ensure their visibility.

### TODO

* [x] Changelog entry

Currently based on #613.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+⚠️ Version 0.13.0 removes the original advisory lock based unique jobs implementation that was deprecated in v0.12.0. See details in the note below or the v0.12.0 release notes.
+
+### Changed
+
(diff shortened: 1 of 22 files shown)


bgentry:
I haven't yet found the time to get the Ruby and Python clients upgraded to support this newer unique jobs path and remove the advisory lock path. However with all the stuff in flight and the much smaller user base of those libraries, I don't think I can justify holding up these other improvements any longer.

I'm going to go ahead and ship this and #627 in order to move forward on #584 and the other stuff that's blocked on them.

PR #627, opened 2024-09-29
Add internal "pilot" concept

bgentry:
This builds out a new layer to sit between the Client and Driver, focused more on higher level operations than individual database queries. This allows those operations to be implemented with a combination of multiple database queries in a transaction, rather than just a single one.

Naming here was tough, I really didn't like any of the options I came up with and ultimately landed on ["pilot"](https://en.wikipedia.org/wiki/Maritime_pilot), but I'm pretty indifferent to it if a truly good one can be found. I spent enough time trying to find a good one that I truly don't care at this point 😆 

For now, the only operations implemented in the pilot are `InsertMany` and `JobSetStateIfRunningMany` (from the completer). I also made some changes within River to use these at the correct places and to propagate the pilot as needed. `JobCompleteTx` was updated to pull a `Client` out of the context in order to access this and use the same underlying query (via the pilot) as the `BatchCompleter`.

Based on #614 (including #624).


internal/jobcompleter/job_completer.go
```diff
@@ -359,7 +354,21 @@ func (c *BatchCompleter) handleBatch(ctx context.Context) error {
...
 
 		return withRetries(ctx, &c.BaseService, c.disableSleep, func(ctx context.Context) ([]*rivertype.JobRow, error) {
-			return c.exec.JobSetStateIfRunningMany(ctx, batchParams)
+			tx, err := c.exec.Begin(ctx)
+			if err != nil {
+				return nil, err
+			}
+			defer tx.Rollback(ctx)
+
+			rows, err := c.pilot.JobSetStateIfRunningMany(ctx, tx, batchParams)
```
bgentry, line 363: The choice here was to either wrap this with a transaction _before_ calling into the pilot method, or to do so _within_ it. I opted for the former primarily because it seemed to simplify the interfaces related to the driver system and avoid having to make the pilot type generic with `TTx`.

Open to changing any of this of course as long as it lets us do what we need to do!

internal/jobcompleter/job_completer_test.go
```diff
@@ -25,6 +26,7 @@ import (
 )
 
 type partialExecutorMock struct {
+	riverdriver.Executor
```
bgentry, line 29: This is now a full executor, but one that allows specific methods to be overridden. A tx variant of it was also needed because of the txn change in the completer.

internal/maintenance/queue_maintainer.go
```diff
@@ -99,7 +99,12 @@ func (m *QueueMaintainer) Start(ctx context.Context) error {
 // reflection and potential for panics.
 func GetService[T startstop.Service](maintainer *QueueMaintainer) T {
 	var kindPtr T
-	return maintainer.servicesByName[reflect.TypeOf(kindPtr).Elem().Name()].(T) //nolint:forcetypeassert
+	return maintainer.servicesByName[serviceName(kindPtr)].(T) //nolint:forcetypeassert
+}
+
+func serviceName(service startstop.Service) string {
+	elem := reflect.TypeOf(service).Elem()
+	return elem.PkgPath() + "." + elem.Name()
```
bgentry, line 107: I made this change to the make the internal names more unique because I had originally had two pro maintenance processes with the same name in different packages, and I lost a fair bit of time trying to figure out why one of them wasn't working. I don't think there are downsides/risks here other than just the internal name being the longer & more complicated path+type combo since it's only used internally.

rivershared/riverpilot/standard.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"context"
+
+	"github.com/riverqueue/river/riverdriver"
+	"github.com/riverqueue/river/rivershared/baseservice"
+	"github.com/riverqueue/river/rivertype"
+)
+
+type StandardPilot struct {
```
bgentry, line 11: This one is just a passthrough to the driver queries as of now.

nexovec:
Would you mind telling me what this is even for? The implementation is basically empty now, what's the pilot supposed to be doing?

bgentry:
@nexovec essentially we need a way to be able to override or extend certain functions with more complex implementations to power upcoming River Pro features. Our "driver" concept mainly exists to power low level queries across different database engines with query or type differences, so abusing/reusing that concept for something like a higher level grouping of functionality didn't feel appropriate, hence the added layer. As of now the "pilot" in the OSS version is just a passthrough to the driver, I'm not sure whether that will change or not in the future, though it's unlikely to be anything users need to care about.

nexovec:
So it's something for workflows?

bgentry:
Not currently being used for workflows, no. But yes for the next Pro feature and some of the ones after that 🚀

PR #632, opened 2024-10-04
Add middleware system for jobs

bgentry:
This is a rebased version of #584 that's been adapted for recent changes including consolidated insert logic and the newer unique jobs implementation.

Here, implement a middleware system that adds middleware functions to job lifecycles, which results in them being invoked during specific phases of a job like as it's being inserted or worked.

The most obvious unlock for this is telemetry (e.g. logging, metrics), but it also acts as a building block for features like encrypted jobs.

There are two middleware interfaces added: `JobInsertMiddleware` and `WorkerMiddleware`. A user could implement these on the same struct type for something like telemetry where you want to inject a trace ID at job insertion, and then resurrect it into the context at execution time. In other cases, such as execution-specific logic, it may only make sense to implement one of these.

The `Worker[T]` type has been extended with a `Middleware()` method that allows each individual worker type to define its own added lists of middleware which will be run _after_ the global `WorkerMiddleware` at the client config level. There are also global `JobInsertMiddleware` on the client config.

### DIFF
--- a/client.go
+++ b/client.go
@@ -139,6 +139,10 @@ type Config struct {
 	// deployments.
 	JobCleanerTimeout time.Duration
 
+	// JobInsertMiddleware are optional functions that can be called around job
+	// insertion.
+	JobInsertMiddleware []rivertype.JobInsertMiddleware
+
 	// JobTimeout is the maximum amount of time a job is allowed to run before its
 	// context is cancelled. A timeout of zero means JobTimeoutDefault will be
 	// used, whereas a value of -1 means the job's context will not be cancelled
@@ -234,6 +238,10 @@ type Config struct {
 	// ahead of time that a worker is properly registered for an inserted job.
 	// (i.e.  That it wasn't forgotten by accident.)
 	Workers *Workers
+
+	// WorkerMiddleware are optional functions that can be called around
+	// all job executions.
+	WorkerMiddleware []rivertype.WorkerMiddleware
 
 	// Scheduler run interval. Shared between the scheduler and producer/job
 	// executors, but not currently exposed for configuration.
@@ -467,6 +475,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		FetchCooldown:               valutil.ValOrDefault(config.FetchCooldown, FetchCooldownDefault),
 		FetchPollInterval:           valutil.ValOrDefault(config.FetchPollInterval, FetchPollIntervalDefault),
 		ID:                          valutil.ValOrDefaultFunc(config.ID, func() string { return defaultClientID(time.Now().UTC()) }),
+		JobInsertMiddleware:         config.JobInsertMiddleware,
 		JobTimeout:                  valutil.ValOrDefault(config.JobTimeout, JobTimeoutDefault),
 		Logger:                      logger,
 		MaxAttempts:                 valutil.ValOrDefault(config.MaxAttempts, MaxAttemptsDefault),
@@ -478,6 +487,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		RetryPolicy:                 retryPolicy,
 		TestOnly:                    config.TestOnly,
 		Workers:                     config.Workers,
+		WorkerMiddleware:            config.WorkerMiddleware,
 		schedulerInterval:           valutil.ValOrDefault(config.schedulerInterval, maintenance.JobSchedulerIntervalDefault),
 		time:                        config.time,
 	}
@@ -1165,7 +1175,7 @@ func (c *Client[TTx]) ID() string {
 	return c.config.ID
 }
 
-func insertParamsFromConfigArgsAndOptions(archetype *baseservice.Archetype, config *Config, args JobArgs, insertOpts *InsertOpts) (*riverdriver.JobInsertFastParams, error) {
+func insertParamsFromConfigArgsAndOptions(archetype *baseservice.Archetype, config *Config, args JobArgs, insertOpts *InsertOpts) (*rivertype.JobInsertParams, error) {
 	encodedArgs, err := json.Marshal(args)
 	if err != nil {
 		return nil, fmt.Errorf("error marshaling args to JSON: %w", err)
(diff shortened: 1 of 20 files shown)


rivertype/river_type.go
```diff
@@ -229,6 +230,47 @@ type AttemptError struct {
...
+// implementations for phases where no custom code is needed, and for forward
+// compatibility in case new functions are added to this interface.
+type JobMiddleware interface {
+	// InsertMany is invoked around a batch insert operation. Implementations
+	// must always include a call to doInner to call down the middleware stack
+	// and perfom the batch insertion, and may run custom code before and after.
+	//
+	// Returning an error from this function will fail the overarching insert
+	// operation, even if the inner insertion originally succeeded.
+	InsertMany(ctx context.Context, manyParams []*JobInsertParams, doInner func(ctx context.Context) ([]*JobInsertResult, error)) ([]*JobInsertResult, error)
```
bgentry, line 262: I think it could still be possible to have a single-job method here for simplicity, but we would need to have some adapter code to make that work. It seemed easier to get this out with a more minimal version that does just what we need for now. Plus `JobMiddlewareDefaults` helps enough with forward compatibility if we update the interface.

client.go
```diff
@@ -1436,39 +1441,58 @@ func (c *Client[TTx]) insertMany(ctx context.Context, tx riverdriver.ExecutorTx,
...
 	}
-	if err := c.maybeNotifyInsertForQueues(ctx, tx, queues); err != nil {
-		return nil, err
+
+	if len(c.config.JobMiddleware) > 0 {
+		// Wrap middlewares in reverse order so the one defined first is wrapped
+		// as the outermost function and is first to receive the operation.
+		for i := len(c.config.JobMiddleware) - 1; i >= 0; i-- {
+			middlewareItem := c.config.JobMiddleware[i] // capture the current middleware item
+			previousDoInner := doInner                  // Capture the current doInner function
```
bgentry, line 1478: In the course of actually using these, it turns out these were both shadowing on subsequent iterations. With just a single middleware it was trying to grab the one at index -1. Realized both of these need to be captured. Probably should figure out some tests to exercise this!

job_executor.go
```diff
@@ -210,8 +211,26 @@ func (e *jobExecutor) execute(ctx context.Context) (res *jobExecutorResult) {
...
+		// Wrap middlewares in reverse order so the one defined first is wrapped
+		// as the outermost function and is first to receive the operation.
+		for i := len(e.JobMiddleware) - 1; i >= 0; i-- {
+			middlewareItem := e.JobMiddleware[i] // capture the current middleware item
+			previousDoInner := doInner           // Capture the current doInner function
+			doInner = func(ctx context.Context) error {
+				return middlewareItem.Work(ctx, e.JobRow, previousDoInner)
+			}
+		}
+	}
```
bgentry, line 231: This can come in a subsequent PR since it's an optimization, but I'm realizing we probably don't want to construct this middleware pipeline on every job execution. I would guess it wouldn't be too hard to do this once at the time the client is started or constructed and then just pass jobs into the existing pipeline.

The same is probably true with the insert middleware—we should be able to do that init just once.

rivertype/river_type.go
```diff
@@ -229,6 +230,49 @@ type AttemptError struct {
...
+
+type WorkerMiddleware interface {
+	// Work is invoked after a job's JSON args being unmarshaled and before the
+	// job is worked. Implementations must always include a call to doInner to
+	// call down the middleware stack and perfom the batch insertion, and may run
+	// custom code before and after.
+	//
+	// Returning an error from this function will fail the overarching work
+	// operation, even if the inner work originally succeeded.
+	Work(ctx context.Context, job *JobRow, doInner func(context.Context) error) error
```
bgentry, line 273: I just added a commit which splits the middleware interface in two. I'm not 100% settled on the design, but I was led here by trying to allow for middleware to be defined per-worker and related to this comment: https://github.com/riverqueue/river/pull/584#discussion_r1750869562

At insert time, we only have a `JobArgs` type and have no guaranteed access to the `Worker`. This inherently means that per-job-type job insertion middleware needs to be derived from the `JobArgs` and not connected to the `Worker`.

However when looking at `Work` middleware and its conceptual similarity to a worker's `Work` function, it can really only make sense for that to be dynamically defined on the stateful `Worker[T]` type. And if one were returning a general purpose `[]JobMiddleware` from their `JobArgs`, they would probably expect that the `Work()` functions on those would actually be used—but again they don't really make sense there rather than on the actual `Worker`.

As such, I've split the interface. While I don't think this split is ideal, I think it might be an unavoidable consequence of where we ended up on the ancient `JobArgs` vs `Worker` discussion from like a year ago 😃

On the plus side, along with this change, I added dynamic worker-level middleware defined by a `Middleware(job *Job[T]) []rivertype.WorkerMiddleware`. The executor calls into this via the work unit to extract the middleware that should be used for this specific job, and it applies those in order after the globally-defined `WorkerMiddleware` on the client config.

To make this work, I did have to move the arg deserialization to happen _before_ the middleware are called, but I think this is probably ok. This also adds some flexibility if there's a use case out there which demands access to the decoded args struct, because you can use that and the worker's other stateful properties when determining which middleware should be used. Not sure that's realistic or not 🤷‍♂️

rivertype/river_type.go
```diff
@@ -229,6 +230,49 @@ type AttemptError struct {
...
+// implementations for phases where no custom code is needed, and for forward
+// compatibility in case new functions are added to this interface.
+type JobInsertMiddleware interface {
+	// InsertMany is invoked around a batch insert operation. Implementations
+	// must always include a call to doInner to call down the middleware stack
+	// and perfom the batch insertion, and may run custom code before and after.
+	//
+	// Returning an error from this function will fail the overarching insert
+	// operation, even if the inner insertion originally succeeded.
+	InsertMany(ctx context.Context, manyParams []*JobInsertParams, doInner func(context.Context) ([]*JobInsertResult, error)) ([]*JobInsertResult, error)
```
bgentry, line 262: @brandur I don't think we ever discussed why these middleware functions don't pass through the `manyParams` or `job` as arguments to `doInner`. I'm guessing it's partly to do with making it easier & cleaner for callers?

I don't think there's a use case for changing out the job struct on the `Work` variant, but for insertion it's conceivable that one could want to add or remove rows to the insert batch as part of a middleware call (in addition to merely manipulating them before insertion).

Something to consider adjusting as this bakes in an RC release.

bgentry:
Moving forward on shipping this into an RC release to unblock other work :v: We can still make adjustments to it after.

PR #647, opened 2024-10-16
docs: update MaxWorkers behaviour as it propagates down to the producer

kamikazechaser:
refs:

*https://github.com/riverqueue/river/blob/master/producer.go#L598

I think this makes it clear on how the max workers value should be tuned based on the job type.

### DIFF
--- a/client.go
+++ b/client.go
@@ -309,7 +309,8 @@ func (c *Config) willExecuteJobs() bool {
 // QueueConfig contains queue-specific configuration.
 type QueueConfig struct {
 	// MaxWorkers is the maximum number of workers to run for the queue, or put
-	// otherwise, the maximum parallelism to run.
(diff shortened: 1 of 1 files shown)


bgentry:
Hi @kamikazechaser, from looking at this I'm not quite sure what's missing from the original explanation that is made more clear with this change.  The statement that `MaxWorkers` corresponds to max parallelism for the client seems concise and correct to me, without getting into internal details like the concept of an executor.

Can you maybe elaborate on what you think is missing from the original explanation? Would still like to improve this further if possible. Thanks.

kamikazechaser:
> without getting into internal details like the concept of an executor.

It was related to some question in one of the issues. But looking back at it now, it makes sense not to get into the internals of how the MaxWorkers option is utilized.

Issue #643, opened 2024-10-13
ClientFromContext always return client not found in context, can only be used in a Worker


tommy31:
Hello,

I try to use `river.ClientFromContextSafely` or `river.ClientFromContext` in a worker. But functions always return an error `client not found in context, can only be used in a Worker`. 

I think i'm missing something, but documentation is pretty straighforward: https://riverqueue.com/docs/context-client 
I found few example is the issues and they all seems to work this way:

- https://github.com/riverqueue/river/issues/317#issuecomment-2071169536
- https://github.com/riverqueue/river/issues/495

Here is the complete code, it's a modified copy of https://github.com/riverqueue/river/blob/master/example_periodic_job_test.go

```yaml
//docker-compose.yml
version: '3.8'
services:
  db:
    image: postgres
    restart: always
    # set shared memory limit when using docker-compose
    shm_size: 128mb
    ports:
      - 5433:5432 ## i change the port !!
    environment:
      - POSTGRES_PASSWORD=test
      - POSTGRES_USER=admin
      - POSTGRES_DB=test_db

  river_ui:
    image: ghcr.io/riverqueue/riverui:latest
    ports:
     - 8080:8080
    environment:
      - DATABASE_URL=postgres://admin:test@db:5432/test_db
```

```go
// main.go
package main

import (
	"context"
	"database/sql"
	"fmt"
	"github.com/jackc/pgx/v5"
	"github.com/riverqueue/river/riverdriver/riverdatabasesql"
	"github.com/riverqueue/river/rivershared/riversharedtest"
	"gorm.io/driver/postgres"
	"gorm.io/gorm"
	"log/slog"
	"time"

	"github.com/riverqueue/river"
	"github.com/riverqueue/river/rivershared/util/slogutil"
)

//
// This file used as a holding place for test helpers for examples so that the
// helpers aren't included in Godoc and keep each example more succinct.
//

// Wait on the given subscription channel for numJobs. Times out with a panic if
// jobs take too long to be received.
func waitForNJobs(subscribeChan <-chan *river.Event, numJobs int) {
	var (
		timeout  = riversharedtest.WaitTimeout()
		deadline = time.Now().Add(10 * time.Second)
		events   = make([]*river.Event, 0, numJobs)
	)

	for {
		select {
		case event := <-subscribeChan:
			events = append(events, event)

			if len(events) >= numJobs {
				return
			}

		case <-time.After(time.Until(deadline)):
[...]

PR #656, opened 2024-10-24
add ClientFromContext examples for both pgx + dbsql

bgentry:
While researching #643, I decided to write examples to illustrate whether or not `ClientFromContext` was actually working as intended. Turns out it is, but that there's some nuanced differences between pgx and database/sql drivers here due to the fact that `pgx.Tx` is an interface, whereas `sql.Tx` is a struct which needs to be referenced by pointer with generics.


example_client_from_context_dbsql_test.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	fmt.Printf("client found in context, id=%s\n", client.ID())
+	return nil
+}
+
+// ExampleClientFromContext_databaseSQL demonstrates how to extract the River
+// client from the worker context when using the [database/sql] driver.
+// ([github.com/riverqueue/river/riverdriver/riverdatabasesql])
+func ExampleClientFromContext_databaseSQL() {
```
bgentry, line 41: These examples are named so that they can be correctly associated with the `ClientFromContext` function. The pkgsite docs will group them together underneath the function's documentation.

https://pkg.go.dev/github.com/fluhus/godoc-tricks#Examples

PR #663, opened 2024-11-01
Immediately fetch another batch when a full set of job is returned

gaffneyc:
Jobs can often be stuck on queues for a long while when jobs are sporadically enqueued in large batches. Prior to this change jobs would be fetched every FetchPollInterval (default 1s) unless new jobs were enqueued. This change reduces that to FetchCooldown (default 100ms) when a full set of jobs is returned implying that there may be more in the queue.

See https://github.com/riverqueue/river/discussions/652 for benchmarks and further discussion.

### DIFF
--- a/producer.go
+++ b/producer.go
@@ -447,7 +447,7 @@ func (p *producer) fetchAndRunLoop(fetchCtx, workCtx context.Context, fetchLimit
 			if p.paused {
 				continue
 			}
-			p.innerFetchLoop(workCtx, fetchResultCh)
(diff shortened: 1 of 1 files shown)


gaffneyc:
https://github.com/riverqueue/rivercla/pull/7

bgentry:
@gaffneyc awesome, thank you! I'm gonna see if I can get some test coverage around this change before merging it in.

bgentry:
An alternative version of this was merged in #664, thank you again! :pray:

PR #678, opened 2024-11-14
Fix running job cancellation when not fetching

bgentry:
Cancellation of running jobs relied on a channel that was only being
received when in the job fetch routine, meaning that jobs which were
cancelled would not be cancelled until the next scheduled fetch.

This was fixed by also receiving from the job cancellation channel when
in the main producer loop, even if no fetches are happening.


client_test.go
```diff
@@ -492,76 +496,87 @@ func Test_Client(t *testing.T) {
...
+			return client.JobCancel(ctx, jobID)
 		})
 	})
 
-	t.Run("CancelScheduledJob", func(t *testing.T) {
-		t.Parallel()
+	t.Run("AlternateSchema", func(t *testing.T) {
 
-		client, _ := setup(t)
+		t.Run("CancelRunningJobInTx", func(t *testing.T) {
```
bgentry, line 516: Not sure what happened here 🤦‍♂️ fixing

PR #679, opened 2024-11-16
unified insert path for periodic jobs

### LINKED ISSUE #675: Middleware for Periodic Jobs (Tracing)
While adding tracing, I've noticed that periodic jobs are inserted without going through middleware configured on the Client - it looks like [PeriodicJobEnqueuer's insertBatch](https://github.com/riverqueue/river/blob/cef224550afab3ec9dc20b6f793e1d62964e6137/internal/maintenance/periodic_job_enqueuer.go#L332) goes directly to the River driver instead of going through the client.

For completeness of the trace (job may be scheduled earlier than actually run if workers are busy), it would be nice to be able to intercept periodic job insertion in middleware.

bgentry:
Job insert middleware were not being utilized for periodic jobs, resulting in behavioral differences (especially Pro features). This periodic job insertion path has been refactored to utilize the unified insertion path from the client, which also makes these inserts pass through the Pilot instead of going directly to the driver.

Fixes #675.

### DIFF
--- a/client.go
+++ b/client.go
@@ -611,7 +611,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		{
 			periodicJobEnqueuer := maintenance.NewPeriodicJobEnqueuer(archetype, &maintenance.PeriodicJobEnqueuerConfig{
 				AdvisoryLockPrefix: config.AdvisoryLockPrefix,
-				NotifyInsert:       client.maybeNotifyInsertForQueues,
+				Insert:             client.insertMany,
 			}, driver.GetExecutor())
 			maintenanceServices = append(maintenanceServices, periodicJobEnqueuer)
 			client.testSignals.periodicJobEnqueuer = &periodicJobEnqueuer.TestSignals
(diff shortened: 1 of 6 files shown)


client.go
```diff
@@ -1421,11 +1421,26 @@ func (c *Client[TTx]) InsertMany(ctx context.Context, params []InsertManyParams)
...
+		return nil, err
+	}
+
+	return c.insertMany(ctx, tx, insertParams)
+}
+
+// insertMany is a shared code path for InsertMany and InsertManyTx, also used
+// by the PeriodicJobEnqueuer.
+func (c *Client[TTx]) insertMany(ctx context.Context, tx riverdriver.ExecutorTx, insertParams []*rivertype.JobInsertParams) ([]*rivertype.JobInsertResult, error) {
+	return c.insertManyShared(ctx, tx, insertParams, func(ctx context.Context, insertParams []*riverdriver.JobInsertFastParams) ([]*rivertype.JobInsertResult, error) {
```
bgentry, line 1443: The nesting in here is getting a bit unwieldy and is probably worth spending some more time to refactor. For now though I want to get this bug fixed quickly due to its impact on pro features.

internal/maintenance/periodic_job_enqueuer.go
```diff
@@ -341,28 +338,13 @@ func (s *PeriodicJobEnqueuer) insertBatch(ctx context.Context, insertParamsMany
...
-			}
-		}
-	}
-
-	if len(queues) > 0 {
-		if err := s.Config.NotifyInsert(ctx, tx, queues); err != nil {
-			s.Logger.ErrorContext(ctx, s.Name+": Error notifying insert", "error", err.Error())
-			return
-		}
-		s.TestSignals.NotifiedQueues.Signal(queues)
```
bgentry, line 365: I like that this change ended up simplifying some things, like not needing to deal with insert notifications in the `PeriodicJobEnqueuer` anymore. Now it's all the same insert logic as the `Client`.

PR #690, opened 2024-12-12
Tolerate existing `unique_states` column in 006 migration

### LINKED ISSUE #688: Problems with migrations for 0.12.0
[CHANGELOG upgrade guide](https://github.com/riverqueue/river/blob/master/CHANGELOG.md#0120---2024-09-23) recommends running index migration manually for large tables. After that, "normal" river migration should be run to finalize the process.

The problem is that the existing index is [tolerated](https://github.com/riverqueue/river/blob/master/riverdriver/riverdatabasesql/migration/main/006_bulk_unique.up.sql#L28), but the existing column [is not](https://github.com/riverqueue/river/blob/master/riverdriver/riverdatabasesql/migration/main/006_bulk_unique.up.sql#L23), which results in the following error:

```
column "unique_states" of relation "river_job" already exists
```

brandur:
As described in #688, if a user has followed changelog instructions to
create the new index concurrently, they will encounter an error as they
try to run the 006 migration to finalize it.

Here, tolerate the column already existing to fix the problem. An
existing index is already tolerated.

Fixes #688.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Dropped internal random generators in favor of `math/rand/v2`, which will have the effect of making code fully incompatible with Go 1.21 (`go.mod` has specified a minimum of 1.22 for some time already though). [PR #691](https://github.com/riverqueue/river/pull/691).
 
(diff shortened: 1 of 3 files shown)


bgentry (approved):

brandur:
Thx. Added changelog.

PR #691, opened 2024-12-12
Drop internalized random number generators in favor of `math/rand/v2`

brandur:
The service archetype and a number of functions internalized a random
source because in Go, that used to be the only way that you could
guarantee that a non-crypto generated was actually seeded.

Go 1.22 introduced `math/rand/v2` which fixed that problem. Now that Go
1.21 support's been dropped, it's a nice clean up to move to the V2
package and get rid of random sources being passed around everywhere.

### DIFF
--- a/rivershared/util/randutil/rand_util.go
+++ b/rivershared/util/randutil/rand_util.go
@@ -2,31 +2,14 @@ package randutil
 
 import (
 	cryptorand "crypto/rand"
-	"encoding/binary"
 	"encoding/hex"
-	mathrand "math/rand"
-	"sync"
+	"math/rand/v2"
 	"time"
 )
 
-// NewCryptoSeededConcurrentSafeSource generates a new pseudo-random source
-// that's been created with a cryptographically secure seed to ensure reasonable
-// distribution of randomness between nodes and services, and wrapped so that
-// access to it is concurrent safe.
-//
-// This project uses this technique instead of falling back on crypto/rand
-// because uses of randomness don't need to be cryptographically secure, and the
-// non-crypto variant is about twenty times faster.
-func NewCryptoSeededConcurrentSafeRand() *mathrand.Rand {
-	return mathrand.New(newCryptoSeededConcurrentSafeSource())
-}
-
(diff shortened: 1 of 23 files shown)


rivershared/util/randutil/rand_util.go
```diff
@@ -2,31 +2,14 @@ package randutil
...
-	return mathrand.New(newCryptoSeededConcurrentSafeSource())
-}
-
 // DurationBetween generates a random duration in the range of [lowerLimit, upperLimit).
-//
-// TODO: When we drop Go 1.21 support, switch to `math/rand/v2` and kill the
-// `rand.Rand` argument.
-func DurationBetween(rand *mathrand.Rand, lowerLimit, upperLimit time.Duration) time.Duration {
-	return time.Duration(IntBetween(rand, int(lowerLimit), int(upperLimit)))
+func DurationBetween(lowerLimit, upperLimit time.Duration) time.Duration {
```
bgentry, line 11: riverpro will need a bump to correspond with these internal API changes

bgentry (approved):

rivershared/util/randutil/rand_util.go
brandur: Makes sense. I'll send one over.

brandur:
Also added a changelog just because use of `math/rand/v2` might be kind of breaking even if the `go.mod` has read 1.22 for a while in case code was being vendored in or something.

bgentry:
@sallam-dev sorry about that, fix & new release for riverui coming shortly!

brandur:
@sallam-dev Just so I understand the breakage correctly ... River UI is still pointed to v0.14.2. Are you saying that the same app is pointing to both River and River UI, thereby getting upgraded to v0.14.3, and thus the breakage?

sallam-dev:
Exactly..

Our usecase involves embedding RiverUI while also using River in the same binary

When upgrading River itself, the shard package got upgraded as well and I had to downgrade back to 0.14.2 to compile the project.

brandur:
@sallam-dev Okay should be fixed as of v0.7.0.

sallam-dev:
Great!

Thanks a ton! 🙏

PR #702, opened 2024-12-24
Alternatively accept PG* env vars for database configuration

### LINKED ISSUE #676: Password in database URL should url encode special characters
Hi, if `DATABASE_URL` is used with password that is randomly generated with special characters such as `{` and `}` (like from AWS) an error occurs `failed: error parsing database URL: cannot parse 'postgres://root:xxxxxx@localhost:5432/dbname': failed to parse as URL
(net/url: invalid userinfo)`

Errors: `river migrate-up --database-url "postgres://root:a{a}@localhost:5432/dbname"`
Works: `river migrate-up --database-url "postgres://root:a%7Ba%7D@localhost:5432/dbname"`

brandur:
This one's related to #676, in which it was reported that database URLs
work somewhat suboptimally when dealing with passwords that contain
characters that traditionally require encoding in URLs (e.g. various
types of brackets).

I was originally going down a path wherein we'd offer an alternative
that'd involve adding additional CLI parameters that could work as an
alternative to `--database-url` based on these ones from psql:

    Connection options:
      -h, --host=HOSTNAME      database server host or socket directory (default: "local socket")
      -p, --port=PORT          database server port (default: "5432")
      -U, --username=USERNAME  database user name (default: "brandur")
      -w, --no-password        never prompt for password
      -W, --password           force password prompt (should happen automatically)

However, as I implemented it, I found that pgx somewhat surprisingly
doesn't allow you to instantiate your own config struct. If not parsed
from a URL, it requires you to assemble a more traditional connection
string like `user=XX pass=YY database=ZZ` and parse that, which
introduces its own special character encoding challenges when using
something like a space, equal, or quotation mark.

Digging further I realized that pgx automatically supports the whole
set of PG* environmental variables like `PGHOST`, `PGPORT`, `PGUSER`,
and supporting these would add a solid way of using the River CLI
without a database URL using a common interface that's also supported
by psql and a variety of other pieces of related software.

Even better, since we expect most people to be using a database URL most
of the time, env vars give us a way to provide an alternative, but one
which doesn't add any complication to the existing CLI interface. In
general fewer parameters is better because it keeps the help docs easy
to understand.

Another advantage is that this will technically let us support more
complex connection setups involving things like TLS certificates. All of
these vars are supported, and none of which had a great answer with a
database URL:

* `PGSSLCERT`
* `PGSSLKEY`
* `PGSSLROOTCERT`
* `PGSSLPASSWORD`

Fixes #676.

### DIFF
--- a/cmd/river/rivercli/command.go
+++ b/cmd/river/rivercli/command.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/riverqueue/river/cmd/river/riverbench"
 	"github.com/riverqueue/river/rivermigrate"
+	"github.com/riverqueue/river/rivershared/util/ptrutil"
 )
 
 const (
@@ -77,7 +78,7 @@ type RunCommandBundle struct {
 }
 
 // RunCommand bootstraps and runs a River CLI subcommand.
-func RunCommand[TOpts CommandOpts](ctx context.Context, bundle *RunCommandBundle, command Command[TOpts], opts TOpts) {
+func RunCommand[TOpts CommandOpts](ctx context.Context, bundle *RunCommandBundle, command Command[TOpts], opts TOpts) error {
 	procureAndRun := func() (bool, error) {
 		if err := opts.Validate(); err != nil {
 			return false, err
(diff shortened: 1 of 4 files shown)


cmd/river/rivercli/command.go
```diff
@@ -164,3 +175,9 @@ func openPgxV5DBPool(ctx context.Context, databaseURL string) (*pgxpool.Pool, er
 
 	return dbPool, nil
 }
+
+func pgEnvConfigured() bool {
+	// Consider these two vars the bare minimum for an env-based configuration.
+	// return os.Getenv("PGHOST") != "" && os.Getenv("PGDATABASE") != ""
+	return os.Getenv("PGDATABASE") != ""
+}
```
bgentry, line 183: Some commented code in here and a comment which only lines up with the commented variant (not the active one)

bgentry (approved):
Nice! :shipit: aside from the one commented code nit

bgentry:
Oh, needs changelog too :pray:

cmd/river/rivercli/command.go
brandur: Ah good catch. I changed my mind on this one. `PGHOST` defaults nicely to `localhost` so it seems okay to assume that `PGDATABASE` could be considered a bare minimum viable configuration.

brandur:
Added changelog.

PR #703, opened 2024-12-24
add line name to migration comment

bgentry:
When printing the pro sequence migration, I realized its comment was still:

```
-- River migration 001 [up]
```

Which is to say that it's the same no matter which line you're migrating, pro or non-pro. I thought this could be improved by adding the line name so it's instead:

```
-- River sequence migration 001 [up]
```

Still doesn't explicitly make pro vs non-pro clear, but it's a small improvement IMO.

Will add a changelog if you agree this is a good change.


brandur (approved):
Yep, seems good. Maybe add changelog too for completeness.

PR #705, opened 2024-12-25
Reduce slice allocations during unique insertions

brandur:
Just reduce some really easy to avoid slice allocations for every unique
insertion:

* No reason to have `requiredV3states` allocated every time a new unique
  opts is validated. It never changes.

* Leave `missingStates` unallocated by default. Most of the time a
  unique insertion will have have no missing unique states, so this
  skips a unique allocation completely the overwhelmingly majority of
  the time.


bgentry (approved):
Good fix! Thanks

PR #707, opened 2024-12-27
Add missing doc for `JobRow.UniqueStates` + reveal `rivertype.UniqueOptsByStateDefault()`

brandur:
Two small ones related to unique job insertion:

* Adding a missing doc comment on `JobRow.UniqueStates`.

* As discussed in #704, reveal a way for end users to get the default
  set of unique job states via `rivertype.UniqueOptsByStateDefault()`.
  Similar to `rivertype.JobStates()`, this is revealed as a function so
  that it's not possible to accidentally mutate a global slice. (Go:
  readonly variables would sure be pretty nice.)

### DIFF
--- a/rivertype/river_type.go
+++ b/rivertype/river_type.go
@@ -129,6 +129,9 @@ type JobRow struct {
 	// opts configuration.
 	UniqueKey []byte
 
+	// UniqueStates is the set of states where uniqueness is enforced for this
+	// job. Equivalent to the default set of unique states unless
+	// UniqueOpts.ByState was assigned a custom value.
 	UniqueStates []JobState
 }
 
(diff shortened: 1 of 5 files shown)


rivertype/river_type.go
```diff
@@ -303,3 +306,19 @@ type Queue struct {
 	// deleted from the table by a maintenance process.
 	UpdatedAt time.Time
 }
+
+// UniqueOptsByStateDefault is the set of job states that are used to determine
+// uniqueness unless unique job states have been overridden with
+// UniqueOpts.ByState. So for example, with this default set a new unique job
+// may be inserted even if another job already exists, as long as that other job
+// is set `cancelled` or `discarded`.
+func UniqueOptsByStateDefault() []JobState {
```
brandur, line 315: Naming on this one seems a little awkward, but I think it might be the best way to go: it's basically `UniqueOpts` + `ByState` + `Default` so it copies the taxonomy of the field it's being set to.

bgentry (approved):

Issue #59, opened 2023-11-21
Autovacuum Starvation


mikehale:
I think we implemented this after your time at Heroku. We finally realized that sometimes the 2 PG auto vacuum processes would both end up working on tables that were taking hours to complete. That kept them from vacuuming the que_jobs table, which would lead to poor performance. We ended up adding a clock process task to manually vacuum our high churn tables at regular intervals, and we haven't seen poor job selection performance since.

It might be worthwhile to add something similar to what [que](https://github.com/que-rb/que/blob/master/docs/README.md#vacuuming) has to your docs to mitigate this potential issue for river users.

bgentry:
Oh hey @mikehale 👋  Thanks for raising this concern. Out of curiosity, how frequently are you manually vacuuming, how long does it take to complete, and are there any related settings you would want to share?

mikehale:
When I left we were vacuuming once every 5 minutes, and IIRC it took less than a second to complete. We had tuned the auto vacuum settings, but obviously those don't come into play with a manual vacuum. Our main concern was about affecting the performance of the rest of the database, through either a lock or increased I/O load. In practice neither of those manifested as issues, and I believe that is partly because of the consistently low overhead due to the regular and frequent vacumming. The trade-off of potentially higher I/O every 5 minutes vs randomly not having auto-vacuum run for many hours turned out to be a good one for us.

dyeje:
We got bit by this. Was pretty frustrating to debug.

bgentry:
@dyeje can you add any context around what you were seeing, what your workload is like, jobs/sec or day, etc?

brandur:
Postgres major version too please. I would've also expected that many of the B-tree optimizations would've helped slow down degenerate table bloat compared to those old days at Heroku.

mikehale:
It was admittedly an extreme case, but I think it had been over 24 hours that the auto vacuum processes were working on very large tables, that coupled with a very thrashed jobs table caused a noticeable though not fatal degradation in job acquisition performance.

PR #718, opened 2025-01-13
Enable the reindexer by default

bgentry:
Following the questions and discussion in https://github.com/riverqueue/river/discussions/717, this PR enables the reindexer by default. It was originally disabled in #34 in the lead up to release because we felt we may want to spend more time addressing potential downsides (like multiple simultaneous reindexings).

Just from looking at the demo app, it's clear that the GIN indexes are particularly prone to long-term bloat. Pulling from what I posted in that discussion, here's the relevant index sizes before and after a manual concurrent reindex call:

```
riverdemo=# SELECT
    indexrelname AS index_name,
    pg_size_pretty(pg_relation_size(indexrelid)) AS index_size
FROM pg_stat_user_indexes;
               index_name               | index_size 
----------------------------------------+------------
 river_job_prioritized_fetching_index   | 80 MB
 river_job_args_index                   | 4013 MB
 river_job_metadata_index               | 645 MB
...
```

```
riverdemo=# SELECT                                    
    indexrelname AS index_name,
    pg_size_pretty(pg_relation_size(indexrelid)) AS index_size
FROM pg_stat_user_indexes;
               index_name               | index_size 
----------------------------------------+------------
 river_job_prioritized_fetching_index   | 38 MB
 river_job_args_index                   | 35 MB
 river_job_metadata_index               | 5656 kB
...
```

This is likely the reason the Fly deployment where the demo is running keeps running out of disk space on its Postgres instance. There might be situations where `river_job_prioritized_fetching_index` would also benefit from reindexing, but it's not clear that it _needs_ it to the same extent. Perhaps with a higher throughput use case (i.e. Heroku's API per https://github.com/riverqueue/river/issues/59) it might matter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `NeverSchedule` returns a `PeriodicSchedule` that never runs. This can be used to effectively disable the reindexer or any other maintenance service. [PR #718](https://github.com/riverqueue/river/pull/718).
+
 ### Changed
 
+- The reindexer maintenance process has been enabled. As of now, it will reindex only the `river_job_args_index` and `river_jobs_metadata_index` `GIN` indexes, which are more prone to bloat than b-tree indexes. By default it runs daily at midnight UTC, but can be customized on the `river.Config` type via `ReindexerSchedule`. Most installations will benefit from this process, but it can be disabled altogether using `NeverSchedule`. [PR #718](https://github.com/riverqueue/river/pull/718).
 - Periodic jobs now have a `"periodic": true` attribute set in their metadata to make them more easily distinguishable from other types of jobs. [PR #728](https://github.com/riverqueue/river/pull/728).
 
 ## [0.15.0] - 2024-12-26
--- a/client_test.go
+++ b/client_test.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -3640,10 +3641,9 @@ func Test_Client_Maintenance(t *testing.T) {
 
 	t.Run("Reindexer", func(t *testing.T) {
 		t.Parallel()
-		t.Skip("Reindexer is disabled for further development")
 
 		config := newTestConfig(t, nil)
-		config.ReindexerSchedule = cron.Every(time.Second)
+		config.ReindexerSchedule = &runOnceSchedule{}
 
 		client, _ := setup(t, config)
 
@@ -3654,6 +3654,18 @@ func Test_Client_Maintenance(t *testing.T) {
 		svc.TestSignals.Reindexed.WaitOrTimeout()
 		svc.TestSignals.Reindexed.WaitOrTimeout()
 	})
+}
+
+type runOnceSchedule struct {
+	ran atomic.Bool
+}
+
+func (s *runOnceSchedule) Next(time.Time) time.Time {
+	if !s.ran.Swap(true) {
+		return time.Now()
+	}
+	// Return the maximum future time so that the schedule doesn't run again.
+	return time.Unix(1<<63-1, 0)
 }
 
 func Test_Client_QueueGet(t *testing.T) {
@@ -4935,6 +4947,12 @@ func Test_NewClient_Defaults(t *testing.T) {
 	require.Zero(t, enqueuer.Config.AdvisoryLockPrefix)
 	require.False(t, enqueuer.StaggerStartupIsDisabled())
 
+	reindexer := maintenance.GetService[*maintenance.Reindexer](client.queueMaintainer)
+	require.Equal(t, []string{"river_job_args_index", "river_job_metadata_index"}, reindexer.Config.IndexNames)
+	now := time.Now().UTC()
+	nextMidnight := time.Date(now.Year(), now.Month(), now.Day(), 0, 0, 0, 0, time.UTC).AddDate(0, 0, 1)
+	require.Equal(t, nextMidnight, reindexer.Config.ScheduleFunc(now))
+
 	require.Nil(t, client.config.ErrorHandler)
 	require.Equal(t, FetchCooldownDefault, client.config.FetchCooldown)
 	require.Equal(t, FetchPollIntervalDefault, client.config.FetchPollInterval)
@@ -4979,6 +4997,7 @@ func Test_NewClient_Overrides(t *testing.T) {
 		Logger:                      logger,
 		MaxAttempts:                 5,
 		Queues:                      map[string]QueueConfig{QueueDefault: {MaxWorkers: 1}},
+		ReindexerSchedule:           &periodicIntervalSchedule{interval: time.Hour},
 		RetryPolicy:                 retryPolicy,
 		TestOnly:                    true, // disables staggered start in maintenance services
 		Workers:                     workers,
@@ -4997,6 +5016,10 @@ func Test_NewClient_Overrides(t *testing.T) {
 	enqueuer := maintenance.GetService[*maintenance.PeriodicJobEnqueuer](client.queueMaintainer)
 	require.Equal(t, int32(123_456), enqueuer.Config.AdvisoryLockPrefix)
 	require.True(t, enqueuer.StaggerStartupIsDisabled())
+
+	reindexer := maintenance.GetService[*maintenance.Reindexer](client.queueMaintainer)
+	now := time.Now().UTC()
+	require.Equal(t, now.Add(time.Hour), reindexer.Config.ScheduleFunc(now))
 
 	require.Equal(t, errorHandler, client.config.ErrorHandler)
 	require.Equal(t, 123*time.Millisecond, client.config.FetchCooldown)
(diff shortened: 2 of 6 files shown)


client.go
```diff
@@ -191,6 +191,11 @@ type Config struct {
 	// than working them. If it's specified, then Workers must also be given.
 	Queues map[string]QueueConfig
 
+	// ReindexerIndexes is a list of indexes to reindex on each run of the
+	// reindexer. If empty, defaults to only the args and metadata GIN indexes
+	// (river_job_args_index and river_job_metadata_index).
+	ReindexerIndexes []string
```
brandur, line 197: Especially given that no one's asking for this, IMO better to leave internal for now. Just improves future flexibility and I can't think of a good reason to have a non-default.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- The reindexer has been enabled. By default, it will reindex only the `river_job_args_index` and `river_jobs_metadata_index` `GIN` indexes, which are more prone to bloat. It will also run daily at midnight UTC by default. Each of these settings can be customized on the `river.Config` type via `ReindexerIndexes` and `ReindexerSchedule`. [PR #XXX](https://github.com/riverqueue/river/pull/XXX).
```
brandur, line 12:
```suggestion
- The reindexer has been enabled. By default, it will reindex only the `river_job_args_index` and `river_jobs_metadata_index` `GIN` indexes, which are more prone to bloat. It will also run daily at midnight UTC by default. Each of these settings can be customized on the `river.Config` type via `ReindexerIndexes` and `ReindexerSchedule`. [PR #718](https://github.com/riverqueue/river/pull/718).
```

client_test.go
```diff
@@ -3656,6 +3656,17 @@ func Test_Client_Maintenance(t *testing.T) {
 	})
 }
 
+type runOnceScheduler struct {
```
brandur, line 3659: Maybe call this a "schedule" instead of a "scheduler" given the existing naming of types like `PeriodicSchedule` and `DefaultReindexerSchedule`.

client_test.go
```diff
@@ -3656,6 +3656,17 @@ func Test_Client_Maintenance(t *testing.T) {
...
+	ran atomic.Bool
+}
+
+func (s *runOnceScheduler) Next(time.Time) time.Time {
+	if !s.ran.Swap(true) {
+		return time.Now()
+	}
+	// Return the maximum future time so that the scheduler doesn't run again.
+	return time.Unix(1<<63-1, 0)
+}
```
brandur, line 3669: Fix spacing here (no newline after this function).

brandur (commented):
A couple thoughts:

* Two data points on index size tell you something, but when evaluating the value of something like this, it'd be better to know about index growth over time instead (even if it were three data points instead of two). Indexes tend to degrade from their pristine state fairly quickly, but it's not a big problem unless they continue to degrade. In the demo app's case, I wonder if it might grow to hit 80 MB, but then basically become stable there.
* What about the original concerns about possible slow reindexer calls? Without getting too fancy, it's probably a good idea to put in a check to verify that a reindex for a specific index isn't already running before starting a new one. This could be done potentially with [`pg_stat_progress_create_index `](https://www.postgresql.org/docs/current/progress-reporting.html#CREATE-INDEX-PROGRESS-REPORTING).
    * Along similar lines, looking at the implementation, it would definitely be better if each reindex was kicked off in sequence rather than all simultaneously. This would require some of the code to be augmented, but it'd be clearly better if it were that way.
* This is definitely in the territory of non-trivial risk, so we should probably need to have a way of turning it off. I wonder if the package should export a `NeverSchedule` or something like that which you could pass to `ReindexerSchedule` to disable it.

bgentry:
> * This is definitely in the territory of non-trivial risk, so we should probably need to have a way of turning it off. I wonder if the package should export a NeverSchedule or something like that which you could pass to ReindexerSchedule to disable it.

Hah, I actually meant to add a “never schedule” to this PR but forgot that item before I pushed it. That’s exactly what I was thinking for disabling it. The other option could be to pass a non nil empty list of indexes, but that seemed a bit more error prone. The other benefit of the never schedule design is we should be able to apply that same design to other maintenance services to disable them.

Will get that added.

> In the demo app's case, I wonder if it might grow to hit 80 MB, but then basically become stable there.

This may not have been clear enough from my original PR but I wanted to make sure you saw that I did _not_ include the ~80 MB `river_job_prioritized_fetching_index` in the default list here, specifically because I believe the b-tree indexes are generally quite well maintained by auto-vacuum.  The ones I did include here are the args and metadata GIN indexes, which shrank from 4013 MB to 35 MB, and 645 MB to 5 KB respectively after a manual reindex. From what I’ve read it seems like GIN indexes are much more prone to this issue.

After a few days, you can see these indexes growing steadily while the main btree fetch index is pretty stable:

```
               index_name               | index_size 
----------------------------------------+------------
 river_job_state_and_finalized_at_index | 26 MB
 river_job_args_index                   | 121 MB
 river_job_metadata_index               | 24 MB
```

> * What about the original concerns about possible slow reindexer calls? Without getting too fancy, it's probably a good idea to put in a check to verify that a reindex for a specific index isn't already running before starting a new one. This could be done potentially with [pg_stat_progress_create_index ](https://www.postgresql.org/docs/current/progress-reporting.html#CREATE-INDEX-PROGRESS-REPORTING).
>
>   * Along similar lines, looking at the implementation, it would definitely be better if each reindex was kicked off in sequence rather than all simultaneously. This would require some of the code to be augmented, but it'd be clearly better if it were that way.

I agree these would be improvements, although I'm not certain I'd delay shipping this until we have these optimizations in place. Is that what you'd prefer to do? My thinking is with the default setting of daily reindexes, you'd need to have some serious issues for the reindex to not complete in that timeframe. If you changed them to run every few minutes or seconds, then yeah, it's a lot more risky.

If you think it's important to do something to address these issues before we enable the feature at all, I think we could mitigate most of the risk with just the first option. We could check for ongoing reindexing attempts each time the reindexer is triggered, and exclude those from the run. Staggering the reindex calls (while still maintaining a schedule) sounds a fair bit more complex.

Lmk what you think is appropriate for this stage vs later improvements.

periodic_job.go
```diff
@@ -87,6 +87,18 @@ func (s *periodicIntervalSchedule) Next(t time.Time) time.Time {
...
+
+func (s *neverSchedule) Next(t time.Time) time.Time {
+	// Return the maximum future time so that the schedule never runs.
+	return time.Unix(1<<63-1, 0)
+}
+
+// ScheduleNever returns a PeriodicSchedule that never runs.
+func ScheduleNever() PeriodicSchedule {
+	return &neverSchedule{}
+}
```
bgentry, line 100: Naming is a little tough here. `neverSchedule` is consistent with the internal name `periodicIntervalSchedule` above, but for the external API name I thought `ScheduleNever()` makes more sense. If we're going to provide one-off schedule options, they should probably be grouped with a common prefix `ScheduleXX` for discoverability.

None are consistent with `PeriodicInterval`'s name though.

Let me know what you think.

client.go
bgentry: I think the reason would be if you want to add an additional index to the default list, like if you were doing enough throughput that `river_job_prioritized_fetching_index` was seeing a lot of bloat. But it can definitely go away for now and be re-added later when requested.

bgentry:
@brandur any thoughts on the remaining questions here? It should be ready to go other than the question on naming.

periodic_job.go
brandur: Hrmm, I'm a little mixed on this one. Given that `PeriodicSchedule` is already an external naming scheme, it may be that `NeverSchedule` fits better into that, don't you think? "Never schedule" also flows off the tongue a bit easier.

brandur:
> After a few days, you can see these indexes growing steadily while the main btree fetch index is pretty stable:
> 
> ```
>                index_name               | index_size 
> ----------------------------------------+------------
>  river_job_state_and_finalized_at_index | 26 MB
>  river_job_args_index                   | 121 MB
>  river_job_metadata_index               | 24 MB
> ```

Just since I don't think I have access and it's been about a week, can you post these numbers one more time (without reindexing)? I'm curious if the Gist indexes stabilize.

> > * What about the original concerns about possible slow reindexer calls? Without getting too fancy, it's probably a good idea to put in a check to verify that a reindex for a specific index isn't already running before starting a new one. This could be done potentially with [pg_stat_progress_create_index ](https://www.postgresql.org/docs/current/progress-reporting.html#CREATE-INDEX-PROGRESS-REPORTING).
> >   
> >   * Along similar lines, looking at the implementation, it would definitely be better if each reindex was kicked off in sequence rather than all simultaneously. This would require some of the code to be augmented, but it'd be clearly better if it were that way.
> 
> I agree these would be improvements, although I'm not certain I'd delay shipping this until we have these optimizations in place. Is that what you'd prefer to do? My thinking is with the default setting of daily reindexes, you'd need to have some serious issues for the reindex to not complete in that timeframe. If you changed them to run every few minutes or seconds, then yeah, it's a lot more risky.

Alright. Yeah, I guess they can be pushed.

brandur (approved):

bgentry:
Now, one week after the last manual run:

```
 river_job_args_index                   | 258 MB
 river_job_metadata_index               | 55 MB
 river_job_prioritized_fetching_index   | 74 MB
```

periodic_job.go
bgentry: Fair point, renamed to `NeverSchedule` 👍

bgentry:
@chaporgin the reindexer was not running regularly in the above statistics queries—those are from the demo app running v0.15.0 before enabling the reindexer.

PR #728, opened 2025-01-23
Set periodic attribute in periodic job metadata

bgentry:
This is a minor improvement coming out of https://github.com/riverqueue/river/discussions/727#discussioncomment-11923628. Periodic jobs are currently indistinguishable from jobs inserted normally, which can make debugging more difficult. This PR changes the periodic job enqueuer to set a `"periodic": true` attribute on all periodic jobs' metadata to make them identifiable.


brandur (approved):
Nice! Yep, seems worthwhile.

PR #731, opened 2025-01-24
feat(config): add SkipUnknownJobCheck client config option to skip job arg worker validation

magaldima:


client.go
```diff
@@ -90,6 +90,17 @@ type Config struct {
...
+	// SkipUnknownJobCheck is a flag to control whether the client should skip
+	// checking to see if a registered worker exists in the client's worker bundle
+	// for a job arg prior to insertion.
+	//
+	// This can be set to true to allow a client to insert jobs which are
+	// intended to be worked by a different client which effectively makes
+	// the client's insertion behavior mimic that of an insert-only client.
+	//
+	// Defaults to false.
+	SkipUnknownJobCheck bool
```
bgentry, line 102: We tend to keep struct fields sorted alphabetically unless there's a compelling reason not to. Could you put this down above `TestOnly`?

bgentry (commented):
Thanks for opening this! Could you please add yourself to [the CLA](https://github.com/riverqueue/rivercla) since this is your first contribution?

Could you also add test coverage for the new config option? For a simple bool field like this, just adding it to the tests `Test_NewClient_Defaults` and `Test_NewClient_Overrides` [here](https://github.com/riverqueue/river/blob/e29646108736768c2247001ae03ad9e78ce688d6/client_test.go#L4922-L5033) should be sufficient. This is essentially just making sure it's propagated through to the client's inner config when the copy is done.

Finally, we'll want some proper test coverage for this as well. We have cases for each insert method named `AllowsUnknownJobKindWithoutWorkers` (6 in total, we should really dedupe the logic now that the insert path and behavior are mostly shared). I think we'll want to add a case below each of those for `AllowsUnknownJobKindWith SkipUnknownJobCheck` to ensure that an unknown job can be inserted even if workers are registered if that config is true.

magaldima:
Signed CLA here: https://github.com/riverqueue/rivercla/pull/8

client.go
magaldima: updated! thanks for the feedback!

brandur:
Great work. Thank you!

Would you mind adding this line in `CHANGELOG.md` under the unreleased "added" section? 

``` md
- Add `SkipUnknownJobCheck` client config option to skip job arg worker validation. [PR #731](https://github.com/riverqueue/river/pull/731).
```

bgentry (approved):
Thanks so much for taking the lead on this feature PR! Hoping to get a release out today with this and other changes. 🚀

PR #733, opened 2025-01-25
Update golangci-lint to v1.63.4

bgentry:
This is a regular upgrade of golangci-lint along with required fixes. The new version of vet seems to detect potentially leaked context around `workCtx` / `workCancel`. This may or may not matter in practice, but rather than ignore the lint error I fixed it by ensuring it's always called after a failed start or on shutdown after the producers have shut down, which should make it generally a no-op.


internal/riverinternaltest/riverdrivertest/riverdrivertest.go
```diff
@@ -905,12 +905,12 @@ func Exercise[TTx any](ctx context.Context, t *testing.T,
 				require.Nil(t, job.AttemptedAt)
 				require.Empty(t, job.AttemptedBy)
 				require.WithinDuration(t, now, job.CreatedAt, 2*time.Second)
-				require.Equal(t, []byte(`{"encoded": "args"}`), job.EncodedArgs)
+				require.JSONEq(t, `{"encoded": "args"}`, string(job.EncodedArgs))
```
brandur, line 908: I can already tell it's going to drive me crazy that they didn't called this `JSONEqual` to match all the other existing `Equal` helpers.

brandur (approved):

PR #734, opened 2025-01-25
Allow by period uniqueness to be based off scheduled time

### LINKED ISSUE #715: Scheduled jobs with unique
I'm dynamically inserting jobs into river with ScheduledAt attribute. 

It seems from my tests that when I call:
```
res, err := riverClient.Insert(ctx,
	SomeArgs{}, &river.InsertOpts{
		ScheduledAt: scheduleTime,
		UniqueOpts: river.UniqueOpts{
			ByPeriod: 12 * time.Minute,
		},
	},
)
```
twice in a loop, second job is treated as violating uniqueness condition and not created. `scheduleTime` is (of course) different by more than 12 minutes.

Does "Uniqueness" feature check only created time, and dosen't work well with scheduled?

brandur:
This one attempts to resolve #715. Currently, by period uniqueness
always bases the period off the current time, but there's a good
argument that if the job has been scheduled for a particular time in the
future, it should be based off that time instead.

This is one that could nominally be considered a small breaking change
in a Hyrum's Law sort of way, even though it's really patching what
could be considered a bug. Even though it was sort of broken before,
some apps may have come to depend on the broken behavior of the unique
code ignoring `ScheduledAt`. I'm not sure that it's a big enough problem
to be worth calling out though, so I didn't.

Fixes #715.


bgentry (approved):
Looks good, I agree with your analysis and think it's unlikely to impact anybody in a negative way. This is probably something we'll want to pull over to the Ruby + Python libs too, right?

brandur:
Thanks.

Yes on Ruby and Python. I'll take a crack at those.

PR #739, opened 2025-02-02
Fix riverdatabasesql nil values in bytea

### LINKED ISSUE #650: riverdatabasesql driver inserts empty unique key arrays instead of null
River migration 6 down fails in river 0.13.0, at least when using the `riverdatabasesql` driver, when there are periodic jobs in the job table. It fails with the following message:

    ERROR: could not create unique index "river_job_kind_unique_key_idx" (SQLSTATE 23505))

The relevant line from the migration is here: https://github.com/riverqueue/river/blob/56ddf095ab3b7ca183425641e84022a194d1111e/riverdriver/riverdatabasesql/migration/main/006_bulk_unique.down.sql#L11

It appears that the periodic jobs, rather than have their unique key set to `NULL`, instead just have an empty byte array (`\x`). As a result, they violate the unique constraint. Perhaps this was different in past versions of river?

bgentry:
As #650 illustrates, the riverdatabasesql driver has issues with encoding arrays of byte arrays (`bytea[]`) and just byte arrays in general because the underlying pq implementation used by sqlc doesn't differentiate between nil vs empty slices. Fix this with a custom type.

The previously written test coverage (first commit) was failing, and gets fixed by this change.

Fixes #650.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- `riverdatabasesql` driver: properly handle `nil` values in `bytea[]` inputs. This fixes the driver's handling of empty unique keys on insert for non-unique jobs with the newer unique jobs implementation. [PR #739](https://github.com/riverqueue/river/pull/739).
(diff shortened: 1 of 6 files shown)


brandur (approved):
Nice one! Thx.

Issue #749, opened 2025-02-10
Change in recent snooze behaviour (v0.16.0)


danieljosephchambers:
Release 0.16.0 introduces new behaviour for the `job.Attempt`:
https://github.com/riverqueue/river/releases/tag/v0.16.0

> Snoozing a job now causes its attempt to be decremented, whereas previously the max_attempts would be incremented. In either case, this avoids allowing a snooze to exhaust a job's retries; however the new behavior also avoids potential issues with wrapping the max_attempts value, and makes it simpler to implement a RetryPolicy based on either attempt or max_attempts. The number of snoozes is also tracked in the job's metadata as snoozes for debugging purposes.
The implementation of the builtin RetryPolicy implementations is not changed, so this change should not cause any user-facing breakage unless you're relying on attempt - len(errors) for some reason. https://github.com/riverqueue/river/pull/730.

One of the ways we were using snoozes was something like:

```
// Inside the Work function
if someCondition {
   if job.Attempt <= maxSnoozeRetries {
    snooze := 5 * time.Minute
    return river.JobSnooze(snooze)
   }
}

```

Ie: in certain circumstances we want to stop the Work function early, and do a limited number of retries (maxSnoozeRetries) which is much less than `max_attempts`: eg 3, before carrying on attempting the job properly.

If I understand correctly this will no longer work since `job.Attempt` doesn't increment when snoozing. Is there a more idiomatic way to do what we're doing here that will work with 0.16.0 onwards?

As I see it there are a few options:

1. rather than relying on counting `job.Attempt`, we independently calculate a time to abandon these retries. And do something like `if now.Before(abandonTime)` to return the snooze. 
2. use the new  `snoozes` attribute in the job's `job.Metadata` instead of the `job.Attempt`. There doesn't appear to be an API to extract key/values from this json blob (which would be nice) so we'd have to write some simple thing to get the `snoozes` or use something like gjson
3. Don't use snoozing for this use case, and instead do something like setting `job.MaxAttempts` to eg 3 - but this doesn't feel right to me.

Thank you for any input

bgentry:
Oof, sorry for the breakage on this @danieljosephchambers! Hyrum's law and all, and I failed to predict the ways in which people could be relying on this 😬 

[...]

PR #750, opened 2025-02-10
feat(river_type): add `GetMetadata` and `MustGetMetadata` methods to `river_type` types

the-potato-church:
Helper methods for interacting with the `Metadata` fields returning a `map[string]interface{}` without introducing breaking changes.

Tangentially related to: https://github.com/riverqueue/river/issues/749

### DIFF
--- a/rivertype/river_type.go
+++ b/rivertype/river_type.go
@@ -306,11 +306,7 @@ type JobInsertMiddleware interface {
 	//
 	// Returning an error from this function will fail the overarching insert
 	// operation, even if the inner insertion originally succeeded.
-	InsertMany(
-		ctx context.Context,
-		manyParams []*JobInsertParams,
-		doInner func(context.Context) ([]*JobInsertResult, error),
-	) ([]*JobInsertResult, error)
+	InsertMany(ctx context.Context, manyParams []*JobInsertParams, doInner func(context.Context) ([]*JobInsertResult, error)) ([]*JobInsertResult, error)
 }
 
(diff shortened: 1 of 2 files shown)


bgentry:
Hmm, I appreciate the thought but I’m not sure if this is the right approach. I’ve had a lot of success developing features using sjson and gjson in cases where the exact structure of the JSON isn’t entirely known or where there are a lot of things which may not be present. The ergonomics of it compared to a `map[string]interface` also feel much better for this style of access, especially when wanting to cast to specific types.

I think I’m inclined to hold off on committing ourselves to a prominent interface like this just yet until we have a better feel for the best way to expose things in here. Meanwhile gjson is already a dependency so it’s easy to use that, or you can have a helper to unmarshal using stdlib if desired.

@brandur your thoughts?

brandur:
@bgentry Yeah, +1 on that. Personally, I'd discourage the use of `map[string]any` because you lose so much type safety. I'd also highly discourage the use of gjson/sjson for the same reason. IMO the right way to use metadata is to make your own type-safe metadata struct and go through `encoding/json`.

I bring that up to make the point that amongst three different people, we have three different opinions on what to do, so we should maybe just leave it up to each River user to decide how to use it. It's quite easy to write your own version of this to put in a utility function if you'd like to have it.

bgentry:
@the-potato-church Once again, thank you for the initiative on this :pray: For now however we'd like to hold off with adding any official APIs on this front, at least until we have a better idea how we'd like it to work.

Issue #605, opened 2024-09-20
Any advice on testing worker code that calls: JobCompleteTx


advdv:
I'm very excited about this project, thank you and hats off! 

I have work function that calls: JobCompleteTx to atomicaly complete the job as well as perform other work on the tx. It seems that this is pretty hard to unit test right now, the JobCompleteTx can only complete if the job actually exist in the database. 

Ofcourse the JobCompleteTx can be mocked/replaced in the tests. Or i could insert actualy jobs, but that is a bit harder to do in my specific case because it's periodic job. Or maybe i'm overlooking something?

brandur:
@advdv Presumably you've tried inserting a job, but aren't able to target it because `JobCompleteTx` requires that it be running. Is that right?

Hmm, we might be missing a helper that allows you to transition a job from from available to running for testing purposes.

advdv:
I did not try to insert the job, I ended up going down a unit testing setup. I would image that for an e2e setup it would definitely be usefull to have a "PrepareJobForTesting" helper function. I now use special tag that i can inspect to not call "JobCompleteTx" during testing. Something like this (NOTE that the tx is real though):

```
// SkipAtomicCompleteTag can be added to a job's tags to indicate that it should not be completed in the same
// tx as the work. This is mainly useful for testing.
const SkipAtomicCompleteTag = "__no_atomic_complete"

// work utility that runs fn within a transaction that is committed with the job completion, or rolled back
// when an error occurred.
func work[TArgs river.JobArgs](
	ctx context.Context,
	rw *pgxpool.Pool,
	job *river.Job[TArgs],
	fn func(pgx.Tx, *river.Job[TArgs]) error,
) error {
	tx, err := rw.Begin(ctx)
	if err != nil {
		return fmt.Errorf("failed to start tx: %w", err)
	}
	defer tx.Rollback(ctx)

	if err := fn(tx, job); err != nil {
		return err
	}

	if !slices.Contains(job.Tags, SkipAtomicCompleteTag) {
		// in tests, we don't usually have an actual job in the database, so we need to be able to skip this:
		// https://github.com/riverqueue/river/issues/605
		if _, err = river.JobCompleteTx[*riverpgxv5.Driver](ctx, tx, job); err != nil {
			return fmt.Errorf("failed to complete job: %w", err)
		}
	}

	if err = tx.Commit(ctx); err != nil {
		return fmt.Errorf("failed to commit: %w", err)
	}

	return nil
}
```

[...]

Issue #751, opened 2025-02-10
Feature Request: Expose Time Control for Testing


mastercactapus:
**Description:**

Our app manages scheduling, and we're migrating some workloads to River, including updating rotations (e.g., advancing to the next person at a set time) which we're looking to leverage the `ScheduledAt` option. In app, everything is based on the `now()` value to avoid clock-drift issues. In our behavioral tests, we currently manipulate PostgreSQL's `now()` function (via proxying) to control time, but we’ve noticed that scheduled jobs don’t run when DB time advances. Looking deeper, it appears River’s scheduling relies on system time instead of database time.

**Feature Request:**

Would it be possible to expose a way to control time? The `river.Config` has a `time` property that takes a `baseservice.TimeGenerator`—would making it public (e.g., `Time`) be enough? This would help with:

- **Testing:** Simulating different time scenarios without modifying system time.
- **Production:** Keeping `ScheduledAt` jobs aligned with database time to mitigate clock drift issues.

**Potential Approaches:**

1. **Make `Time` Public:** Allowing a custom time source to be set.
2. **Sync to DB Time:** Providing an option for River to rely on PostgreSQL's `now()`.

**Benefits:**

- More reliable scheduling in both tests and production.
- Prevents early job execution due to clock drift.
- Simplifies simulating past and future time scenarios in tests.

I would love to hear your thoughts on this; thanks for considering it!

brandur:
Yeah, definitely not against this.

My fear at the moment is that it might not work that well without some substantial refactoring. Per your two options above:

1. (Make time public) This might work fairly well, but I think we're still using a mix of the time generator and database `now()`, so it'd work in some places, but not always reliably.
2. (Sync to DB time) I think we've generally tried to do this wherever it's easily doable, but there are some places where it'd require an extra round trip to the DB, which isn't great. e.g. When generating a unique key using `ByPeriod` in job uniqueness. Also, stubbing Postgres' `now()` is pretty freaking hard. You guys have done it, but it's a big lift if you haven't built this out already.

@bgentry Any thoughts here? (1) seems the most doable to me, but I'm a bit afraid we'd find out a whole bunch of new bugs with it as soon as we implement it. Probably we'd need to audit every use of `now()` in the code base and then have it be overridable with a stubbed time if provided. Might be possible given a few hours of work.

mastercactapus:
[...]

PR #753, opened 2025-02-11
rivertest: implement a Worker type to simplify testing

bgentry:
Since the early days of the project, we've talked about potentially building some APIs to make it easier to test River workers in a realistic execution environment. This is an attempt to dramatically simplify this process, partly derived from testing implementations I wrote for River Pro's API backend code.

The `rivertest.Worker` type makes it much simpler to test River workers with a realistic execution environment without needing to touch the database at all. The type leverages a `river.Config` to allocate an inner `Client` and to enable jobs to be easily created with the normal default values, along with full access to features like middleweare. The execution context also contains all the normal stuff that a real context in River might, such as the client and a timeout if configured.

Here's a simple example of how to use this:

```go
worker := river.WorkFunc(func(ctx context.Context, job *river.Job[testArgs]) error {
	return nil
})
testWorker := rivertest.NewWorker(t, driver, config, worker)

// Now that the `rivertest.Worker` has been initialized, it can be used to run as many different jobs as you'd like:
require.NoError(t, testWorker.Work(context.Background(), t, testArgs{Value: "test"}))

// An opts variant allows customizing insert opts:
require.NoError(t, testWorker.Work(context.Background(), t, testArgs{Value: "test"}), &river.InsertOpts{Metadata: `{"some": "value"}`))

// Want to totally customize the input job to test it in any desired state? See `WorkJob` or `WorkJobOpts`.
```

This is missing a lot of test coverage and there are some open TODOs like extracting the `insertParamsFromConfigArgsAndOptions` from the main `river` package so `rivertest` can also use it, but this should hopefully illustrate what I think is a pretty smooth API for easier testing of workers than what it otherwise takes today.

Related issue: #605

Currently based on #754.


rivertest/worker.go
```diff
@@ -0,0 +1,166 @@
...
+		// TODO: is this timeout being applied in the wrong place? Should it apply
+		// on the *outside* of all middleware?
+		if jobTimeout > 0 {
+			var cancel context.CancelFunc
+			ctx, cancel = context.WithTimeout(ctx, jobTimeout)
+			defer cancel()
+		}
+
+		return w.worker.Work(ctx, job)
+	}
```
brandur, line 90: Would definitely be nice to be able to share all this with the main package a little better. I could see it being an easy mistake to make to add something in one place and then forget to do the other.

brandur (commented):
Cool. Yep, idea looks good to me. Maybe finish it up and add some niceties like docs on the public-facing functions and maybe an example test to demonstrate usage.

rivertest/worker.go
```diff
@@ -0,0 +1,166 @@
...
+	encodedArgs, err := json.Marshal(args)
+	require.NoError(t, err)
+
+	// Round trip the args to validate JSON marshalling/unmarshalling on the type.
+	var decodedArgs T
+	if err = json.Unmarshal(encodedArgs, &decodedArgs); err != nil {
+		return nil, err
+	}
+
+	// TODO: use insertParamsFromConfigArgsAndOptions instead of reimplementing it here?
```
bgentry, line 135: @brandur looked at this one a bit, and `insertParamsFromConfigArgsAndOptions` can't really be extracted from the main package because there are a number of dependencies on its own types, including interfaces whose methods return structs in that package. There's no way to avoid circular dependencies.

Options I see:

1. Live with the ugliness of exporting this API even though it's not useful or intended for public usage
2. Essentially reimplement its logic here to create realistic test jobs
3. Instead use `rivershared/testfactory` for building test jobs (wrap `testfactory.Job_Build` to return a `*Job[T]` instead of `*riverdriver.JobInsertFullParams`.

rivertest/worker.go
```diff
@@ -0,0 +1,186 @@
...
+		EncodedArgs: encodedArgs,
+		Kind:        args.Kind(),
+		MaxAttempts: valutil.FirstNonZero(opts.MaxAttempts, jobInsertOpts.MaxAttempts, config.MaxAttempts, river.MaxAttemptsDefault),
+		Metadata:    sliceutil.FirstNonEmpty(opts.Metadata, jobInsertOpts.Metadata, []byte(`{}`)),
+		Priority:    valutil.FirstNonZero(opts.Priority, jobInsertOpts.Priority, river.PriorityDefault),
+		Queue:       valutil.FirstNonZero(opts.Queue, jobInsertOpts.Queue, river.QueueDefault),
+		State:       rivertype.JobStateAvailable,
+		Tags:        sliceutil.FirstNonEmpty(opts.Tags, jobInsertOpts.Tags, []string{}),
+	}
+	// TODO: use exposed client config Time when added
```
bgentry, line 119: Timely and relevant to #751: here's a place within River where I'm wanting to set the time generator explicitly for testing purposes (though I need to reach into the internal dbunique package as well).

bgentry:
@brandur this is still missing documentation and an example but I think the implementation is essentially there. Want to take a look?

job_complete_tx.go
```diff
@@ -50,6 +50,9 @@ func JobCompleteTx[TDriver riverdriver.Driver[TTx], TTx any, TArgs JobArgs](ctx
 	if err != nil {
 		return nil, err
 	}
+	if len(rows) == 0 {
+		return nil, rivertype.ErrNotFound
+	}
```
bgentry, line 55: Uncovered this bug while thinking about how `JobCompleteTx` would function if called from the new test helpers. Will add it to the changelog.

rivertest/worker.go
```diff
@@ -0,0 +1,218 @@
...
+	jobTimeout := valutil.FirstNonZero(w.worker.Timeout(job), w.config.JobTimeout)
+	ctx, cancel := execution.MaybeApplyTimeout(ctx, jobTimeout)
+	defer cancel()
+
+	return doInner(ctx)
+}
+
+// WorkOpts allocates a synthetic job with the provided arguments and insert
+// options, then works it.
+func (w *Worker[T, TTx]) WorkOpts(ctx context.Context, tb testing.TB, args T, opts river.InsertOpts) error {
```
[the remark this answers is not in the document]
brandur: > I opted to make a separate helper method for working a job with custom insert opts because (1) I expect the vast majority of use cases won't need to care about this and will only care about worker + args, and (2) the much more common use case will benefit from not needing to specify `, nil` at the end of every call.

Works for me I suppose, but just a note that you could state that exact same rationale as to why `Insert` should've been broken into `Insert` + `InsertOpts`.

rivertest/worker.go
```diff
@@ -0,0 +1,218 @@
...
+//	}
+//
+//	if err := worker.WorkOpts(ctx, t, args, river.InsertOpts{Queue: "custom_queue"}); err != nil {
+//		t.Fatalf("failed to work job: %s", err)
+//	}
+//
+// In all cases the underlying [river.Worker] will be called with the synthetic
+// job. The execution environment has a realistic River context with access to
+// all River features, including [river.ClientFromContext] and worker
+// middleware.
```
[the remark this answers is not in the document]
brandur: Discussed a bit on Slack. Option (2) seems to be the most plausible, although stubbing out real functionality that'd otherwise hit a DB or do so something else in the real world is always a little icky. Hopefully though this is a special case where `river.JobCompleteTx` is already tested so well that it's not as necessary to do it here.

brandur (approved):

rivertest/worker.go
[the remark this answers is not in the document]
bgentry: This turned out to be more complicated so we decided to leverage the context value to panic if `JobCompleteTx` is called within a test worker environment and when the job doesn't exist in the database.

rivertest/worker.go
```diff
@@ -0,0 +1,223 @@
...
+//
+// When relying on features that require a database record (such as JobCompleteTx),
+// the job must be inserted into the database first and then executed with
+// WorkJob.
```
brandur, line 41: We can probably punt on this, but it'd probably make sense to have a test or something that'd make sure users have a sane way to do this.

`JobCompleteTx` will probably refuse to complete the job unless it's `running`, and there might not be that easy of a way to manipulate a job from `available` to `running` right now. They could `Client.Insert`, but then it'd get stuck there.

A couple possibilities:

* We could provide a test helper for inserting jobs in any state similar to our internal factories.
* We could have the test helpers here check to see if a job exists in the DB when running. If so, transition the job to `running`.

rivertest/worker.go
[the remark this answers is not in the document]
bgentry: You're right and I had a similar thought. Consolidated to just `Work` (mirroring client `Insert` method) + `WorkJob` for a full job struct.

brandur (approved):

rivertest/worker.go
bgentry: Yeah, I just ran into that while writing the test for this! 😆 I think it would make sense as a fast follow so we can refine the API a little more.

PR #754, opened 2025-02-13
expose `Config.Test` with `Time` setting

### LINKED ISSUE #751: Feature Request: Expose Time Control for Testing

bgentry:
The new `TestConfig` type will contain test-specific client configs. For now, its only setting is a `Time` field which contains a `TimeGenerator` type that can be used to set a synthetic clock for testing.

As part of this, the `TimeGenerator` interface was moved from the semi-internal `rivershared` package to `rivertype` for the sake of API guarantees. This will require updates in dependent packages (riverpro).

Fixes #751.

### DIFF
--- a/rivertest/time_stub.go
+++ b/rivertest/time_stub.go
@@ -1,0 +1,48 @@
+package rivertest
+
+import (
+	"sync"
+	"time"
+)
+
+// TimeStub implements rivertype.TimeGenerator to allow time to be stubbed in
+// tests. It is implemented in a thread-safe manner with a mutex, allowing the
+// current time to be stubbed at any time with StubNowUTC.
+type TimeStub struct {
+	mu     sync.RWMutex
+	nowUTC *time.Time
+}
+
+// NowUTC returns the current time. This may be a stubbed time if the time has
+// been actively stubbed in a test.
+func (t *TimeStub) NowUTC() time.Time {
+	t.mu.RLock()
+	defer t.mu.RUnlock()
+
+	if t.nowUTC == nil {
+		return time.Now().UTC()
+	}
+
+	return *t.nowUTC
+}
+
+// NowUTCOrNil returns if the currently stubbed time _if_ the current time
+// is stubbed, and returns nil otherwise. This is generally useful in cases
+// where a component may want to use a stubbed time if the time is stubbed,
+// but to fall back to a database time default otherwise.
+func (t *TimeStub) NowUTCOrNil() *time.Time {
+	t.mu.RLock()
+	defer t.mu.RUnlock()
+
+	return t.nowUTC
+}
+
+// StubNowUTC stubs the current time. It will panic if invoked outside of tests.
+// Returns the same time passed as parameter for convenience.
+func (t *TimeStub) StubNowUTC(nowUTC time.Time) time.Time {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	t.nowUTC = &nowUTC
+	return nowUTC
+}
--- a/rivertype/time_generator.go
+++ b/rivertype/time_generator.go
@@ -1,0 +1,19 @@
+package rivertype
+
+import "time"
+
+// TimeGenerator generates a current time in UTC. In test environments it's
+// implemented by riverinternaltest.timeStub which lets the current time be
+// stubbed. Otherwise, it's implemented as UnStubbableTimeGenerator which
+// doesn't allow stubbing.
+type TimeGenerator interface {
+	// NowUTC returns the current time. This may be a stubbed time if the time
+	// has been actively stubbed in a test.
+	NowUTC() time.Time
+
+	// NowUTCOrNil returns if the currently stubbed time _if_ the current time
+	// is stubbed, and returns nil otherwise. This is generally useful in cases
+	// where a component may want to use a stubbed time if the time is stubbed,
+	// but to fall back to a database time default otherwise.
+	NowUTCOrNil() *time.Time
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Exposed `TestConfig` struct on `Config` under the `Test` field for configuration that is specific to test environments. For now, the only field on this type is `Time`, which can be used to set a synthetic `TimeGenerator` for tests. [PR #754](https://github.com/riverqueue/river/pull/754).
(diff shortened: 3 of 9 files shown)


rivertype/time_generator.go
```diff
@@ -0,0 +1,33 @@
...
+	NowUTCOrNil() *time.Time
+
+	// StubNowUTC stubs the current time. It will panic if invoked outside of
+	// tests. Returns the same time passed as parameter for convenience.
+	StubNowUTC(nowUTC time.Time) time.Time
+}
+
+// UnStubbableTimeGenerator is a TimeGenerator implementation that can't be
+// stubbed. It's always the generator used outside of tests.
+type UnStubbableTimeGenerator struct{}
```
brandur, line 27: I don't think users will ever need to access `UnStubbableTimeGenerator` since it's set by default and they'll only want to ever set a mock right? Maybe move this back to `rivershared`.

rivertype/time_generator.go
```diff
@@ -0,0 +1,33 @@
...
+
+	// NowUTCOrNil returns if the currently stubbed time _if_ the current time
+	// is stubbed, and returns nil otherwise. This is generally useful in cases
+	// where a component may want to use a stubbed time if the time is stubbed,
+	// but to fall back to a database time default otherwise.
+	NowUTCOrNil() *time.Time
+
+	// StubNowUTC stubs the current time. It will panic if invoked outside of
+	// tests. Returns the same time passed as parameter for convenience.
+	StubNowUTC(nowUTC time.Time) time.Time
```
brandur, line 22: It feels a little off to require this as part of the interface — it's basically there so we can do shortcuts in tests like this:

``` go
		archetype := riversharedtest.BaseServiceArchetype(t)
		archetype.Time.StubNowUTC(time.Now().UTC())
```

But unlike our test code, users won't be able to do that anyway because they won't have access to the `TimeGenerator` on a River client after they've set it (archetype is not exported).

We could potentially fix this by breaking `TimeGenerator` into two related interfaces along with a wrapper like:

``` go
index 74643ad..06c9533 100644
--- a/client.go
+++ b/client.go
@@ -510,7 +510,11 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 
 	archetype := baseservice.NewArchetype(config.Logger)
 	if config.time != nil {
-		archetype.Time = config.time
+		if withStub, ok := config.time.(baseservice.TimeGeneratorWithStub); ok {
+			archetype.Time = withStub
+		} else {
+			archetype.Time = &baseservice.TimeGeneratorWithStubWrapper{TimeGenerator: config.time}
+		}
 	}
 
 	client := &Client[TTx]{
diff --git a/rivershared/baseservice/base_service.go b/rivershared/baseservice/base_service.go
index 764c0f3..d5b1fe2 100644
--- a/rivershared/baseservice/base_service.go
+++ b/rivershared/baseservice/base_service.go
@@ -25,7 +25,7 @@ type Archetype struct {
 	// `time.Now().UTC()`, but is riverinternaltest.timeStub in tests to allow
 	// the current time to be stubbed.  Services should try to use this function
 	// instead of the vanilla ones from the `time` package for testing purposes.
-	Time TimeGenerator
+	Time TimeGeneratorWithStub
 }
 
 // NewArchetype returns a new archetype. This function is most suitable for
@@ -90,12 +90,28 @@ type TimeGenerator interface {
 	// where a component may want to use a stubbed time if the time is stubbed,
 	// but to fall back to a database time default otherwise.
 	NowUTCOrNil() *time.Time
+}
+
+type TimeGeneratorWithStub interface {
+	TimeGenerator
 
 	// StubNowUTC stubs the current time. It will panic if invoked outside of
 	// tests. Returns the same time passed as parameter for convenience.
 	StubNowUTC(nowUTC time.Time) time.Time
 }
 
+// TimeGeneratorWithStubWrapper provides a wrapper around TimeGenerator that
+// implements missing TimeGeneratorWithStub functions. This is used so that we
+// only need to expose the minimal TimeGenerator interface publicly, but can
+// keep a stubbable version of widely available for internal use.
+type TimeGeneratorWithStubWrapper struct {
+	TimeGenerator
+}
+
+func (g *TimeGeneratorWithStubWrapper) StubNowUTC(nowUTC time.Time) time.Time {
+	panic("time not stubbable outside tests")
+}
+
 // UnStubbableTimeGenerator is a TimeGenerator implementation that can't be
 // stubbed. It's always the generator used outside of tests.
 type UnStubbableTimeGenerator struct{}
```

Thoughts?

brandur (commented):
The only other thing that comes to mind is that it feels a little weird to make people write their own `TimeGenerator` implementations every time they used the project. Should we move `TimeStub` into `rivertest` do you think?

rivertype/time_generator.go
bgentry: All great ideas, thank you! Applied them all in the latest revision.

brandur (approved):
lgtm.

rivertest/time_stub.go
```diff
@@ -0,0 +1,48 @@
...
+func (t *TimeStub) NowUTCOrNil() *time.Time {
+	t.mu.RLock()
+	defer t.mu.RUnlock()
+
+	return t.nowUTC
+}
+
+// StubNowUTC stubs the current time. It will panic if invoked outside of tests.
+// Returns the same time passed as parameter for convenience.
+func (t *TimeStub) StubNowUTC(nowUTC time.Time) time.Time {
```
bgentry, line 42: We could just call this `Stub`? It doesn't even need to match the internal `baseservice.TImeGeneratorWithStub` because there's a circular dependency and that one has to exist separately. Your call!

brandur: Hm, unless we're going to rename the other two functions, I'd probably leave it. Just seems more consistent.

brandur (approved):

PR #755, opened 2025-02-13
Use Go 1.24 in CI, golangci-lint v1.64.4, remove lint rules from pre-1.22

bgentry:
This bumps CI to test against Go v1.24 and v1.23. It also upgrades golangci-lint to v1.64.4 and fixes lint issues from that upgrade.

In doing so, I noticed we had some rules still enabled that were specific to Go 1.21 before it became unnecessary to capture range vars, and before the simplified range syntax was available. We could have already turned these off as of #667 (if not sooner) but didn't until now. I've removed the disabled rules and allowed the auto-fixer to correct them.


cmd/river/rivercli/river_cli_test.go
```diff
@@ -148,7 +148,7 @@ func TestBaseCommandSetIntegration(t *testing.T) {
 		buildInfo, _ := debug.ReadBuildInfo()
 
 		require.Equal(t, strings.TrimSpace(fmt.Sprintf(`
-River version (unknown)
+River version (devel)
```
bgentry, line 151: huh, for some reason this value (from `debug.ReadBuildInfo().Main.Version`) is coming back as `"devel"` on v1.24, but `""` on v1.23. I wonder if they messed up a flag in the Go release somehow?

brandur: Oh weird. Yeah not sure — it might also just be a minor breaking change I suppose. They released a change to `go vet` (which runs with other Go commands) that bans use of variable strings in printf-like format directives that initially broke our build at work when I upgraded. It wasn't too hard to fix, but I was surprised it wasn't considered breaking TBH.

brandur (approved):

cmd/river/rivercli/river_cli_test.go
bgentry: I added an override for this specifically for `go1.24.` and when the reported version is `(devel)` to ensure we get identical output between versions. This will keep working whether or not the Go behavior is tweaked in i.e. 1.24.1.

PR #758, opened 2025-02-15
Recorded job output, allow metadata updates upon job completion

bgentry:
This adds a feature we've had several requests for: the ability to record some output from a job that can be stashed alongside the job itself, rather than needing to create your own separate table for it. (Ex https://github.com/riverqueue/river/discussions/500) This will be particularly useful for workflows where it's common to have a task that depends on something an earlier task did, and in these cases there's not always another existing table where it's convenient to store that result (such as an upload URL) for the next task to make use of.

This PR implements this functionality via a `river.RecordOutput` helper. The helper takes an `any` as input and serializes it to JSON in the `"output"` metadata key on the job row. There are two possible mechanisms explored in this PR's commits as of now:

1. Stash the pending metadata changes into an unexported field on `rivertype.JobRow`, along with helpers to set and retrieve these values (since they're in another package from the executor).
2. Stash the pending metadata changes in the context so they can be retrieved following the job's execution (or in `JobCompleteTx` if used).

Each of these has tradeoffs, with (1) adding extra methods to `JobRow` that aren't usable outside of an executing job, and (2) relying on stuffing data in the context (which again won't do anything useful outside an executing job). Interested to hear your thoughts on these @brandur, I will save mine for after you form your own opinions 😄 Easy to flip it back to option (1) as I had in earlier commits, or stick with (2) if preferable.

Additionally, this output feature is implemented using a lower level feature that enables any metadata key & value to be "put" / overwritten as part of the job's completion. For now, I've leveraged this feature to replace the specialized snooze increment argument, as well as for recorded output. It could also potentially be used in middleware to i.e. record job timing details into a metadata key, but without requiring additional SQL updates.

The completion queries were updated to facilitate this, particularly by making sure this metadata merge process still happens even if the job is somehow already completed. This would cover an edge case where a middleware puts a metadata key _after_ the job has finished executing and completing itself via `JobCompleteTx`; information such as the timing example above would still need to be saved with the job post-execution, even though it'd happen via a separate SQL `UPDATE`.


rivertype/river_type.go
```diff
@@ -136,6 +140,14 @@ type JobRow struct {
 	UniqueStates []JobState
 }
 
+// Output returns the previously recorded output for the job, if any.
+func (j *JobRow) Output() []byte {
+	if !gjson.GetBytes(j.Metadata, MetadataKeyOutput).Exists() {
+		return nil
+	}
+	return []byte(gjson.GetBytes(j.Metadata, MetadataKeyOutput).Raw)
```
brandur, line 148: You're repeating the same work here twice by invoking into `GetBytes()`.

recorded_output.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+
+	"github.com/riverqueue/river/rivertype"
+)
+
+func RecordOutput(ctx context.Context, output any) error {
```
brandur, line 11: I don't think I'm quite wrapping my head around this API — is the expectation is that users keep some sort of persistent object around that they continue to append to as the job goes, then call `RecordOutput` at the end of each of their job functions?

If so, a couple things:

* Isn't that pretty inconvenient? You'd have to remember to do that for every worker you made.
* You presumably want to get as much output as possible, so you'd do `RecordOutput` at the bottom of the work function, but doesn't that mean you lose everything if you return early?

I probably would've taken a completely approach for this — given that the project is pretty tied into slog already, I would've added a slog handler that could persist log output to context, and then maybe an easy way to get a per-worker invocation slog logger that'd persist everything the user logged into it after the worker finished up.

Perhaps the difference is that I don't 100% understand what you mean by "output". What kind of data is output do you expect to go in here exactly?

bgentry: The idea is to give you a place that you can store essentially any job output that you want without it having a permanent home in another table. For example, if you have a workflow where a later step depends on the result of previous step, you could put the output in here and pick it up from the subsequent task. An example would be a step that creates an external resource and needs to make the new identifier available to the next step.

In addition to a couple of existing Pro customers that have asked for something like this, there’s this comment:
https://github.com/riverqueue/river/discussions/111#discussioncomment-7862870

Oban for example has a [recorded jobs](https://hexdocs.pm/oban/2.11.0/pro_worker.html#recorded-jobs) feature.

I think it’s a common part of workflow systems but also useful for individual jobs. I believe Cloudflare Workflows has something similar as well.

Obviously a difference in our case is there’s no return value to use as the output. We _could_ implement this with another special error return type I suppose 🤔

brandur: Oh I see! Okay I didn't realize this was mainly intended for workflow use. It'd be useful to put some of this info into the docstring here so people can get some context around what this is useful for.

Okay one more on the workflow thing: how are downstream jobs meant to pull output from their upstream dependencies? Workflow tasks don't receive their input jobs when they run do they? And I don't think you'd even have job IDs that you could use to make a look up?

bgentry: > Okay one more on the workflow thing: how are downstream jobs meant to pull output from their upstream dependencies? Workflow tasks don't receive their input jobs when they run do they? And I don't think you'd even have job IDs that you could use to make a look up?

As of now there's nothing for this. We have [`WorkflowFromExisting`](https://riverqueue.com/pkg/riverpro/latest/riverpro#WorkflowFromExisting) to allow you to append jobs to an existing workflow from within it, but the deps are loaded lazily and are not exposed in any way. Once we have a way to stash the output I want to iterate on this aspect of it for sure.

brandur: Ah gotcha. Okay, good to know!

brandur: Going to resolve this one in favor of updated comment.

rivertype/river_type.go
brandur: Not 100% sure about this, but figured I'd call it out.

IMO, GJSON's APIs are very wrong — it doesn't return an error for covenience's sake, but there's no way it can guarantee that the bytes it's looking at are valid JSON. It optimizes for some user convenience at the cost of behaving degenerately under any edge condition encountered (not sure what it does, but it'd have to be either (1) panic, or (2) return the wrong thing).

By using JSON, we're bubbling up that API. I think in our case it's a little more defensible because we have a strong guarantee of well formed JSON because Postgres will guarantee it at the row level, but you can still imagine edge cases like if someone manually initialize a `JobRow` with bad JSON by accident, invokes `Output` and can't figure out why they're getting the totally wrong thing.

So not sure which way I'd go, but there's a possibility the return type of this function should fit the rest of what you see in Go and be a `([]byte, error)`.

Larger convo, but IMO we should remove GJSON/SJSON back out of the project. They give you a little syntatic sugar, but at the cost of two more dependencies, poor hygiene around perf (e.g. the doubled up work we see on this line), and poor hygiene around considering edges (i.e. degenerate cases and error conditions).

brandur (commented):
Awesome dude! Looking good to me.

riverdriver/river_driver_interface_test.go
```diff
@@ -1 +1,266 @@
...
+		require.NotNil(t, result.ScheduledAt)
+		require.True(t, result.ScheduledAt.Equal(scheduledAt))
+		require.Equal(t, rivertype.JobStateScheduled, result.State)
+	})
+}
+
+func TestJobSetStateSnoozedAvailable(t *testing.T) { //nolint:dupl
+	t.Parallel()
+
+	t.Run("empty metadata", func(t *testing.T) {
```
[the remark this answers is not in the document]
bgentry: Whoops, not sure how that snuck through, thanks!

riverdriver/river_driver_interface.go
```diff
@@ -312,54 +312,111 @@ type JobScheduleResult struct {
...
+	ErrData         []byte
+	FinalizedAt     *time.Time
+	MetadataDoMerge bool
+	MetadataUpdates []byte
+	ScheduledAt     *time.Time
+	State           rivertype.JobState
+}
+
+func JobSetStateCancelled(id int64, finalizedAt time.Time, errData []byte, metadataUpdates []byte) *JobSetStateIfRunningParams {
+	metadataDoUpdate := len(metadataUpdates) > 0
```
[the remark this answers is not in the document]
bgentry: Ah yep thanks, there was an earlier version where this made more sense but that justification was lost after refactoring 👍

rivertype/river_type.go
bgentry: It's easy enough to remove the usage here, though the options have tradeoffs:

1. Use a struct and duplicate the output key name in a struct tag
2. Decode and allocate memory for all the other keys and values I don't care about, only to throw them away
3. Use lower level `json.Decoder` primitives to extract only what I need, with a lot more code than I have now.

Given your desire to nuke this dependency I'll choose (1) for now. Maybe `encoding/json/v2` will make this easy to do without these drawbacks.

recorded_output.go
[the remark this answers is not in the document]
brandur: Great comment. Thanks.

brandur (approved):
Good stuff man! Ty.

Issue #759, opened 2025-02-15
NeverSchedule: invalid max time time causes periodic job to run with no interval


krhubert:
When NeverSchedule is used, a periodic job is run immediately with zero interval between each job.

Links:
- playground example - https://go.dev/play/p/6deycEfuQ_u
- stackoverflow with more details - https://stackoverflow.com/questions/25065055/what-is-the-maximum-time-time-in-go

PR #760, opened 2025-02-15
update max time in NeverSchedule.Next method

krhubert:
closes https://github.com/riverqueue/river/issues/759

### DIFF
--- a/periodic_job_test.go
+++ b/periodic_job_test.go
@@ -18,7 +18,7 @@ func TestNeverSchedule(t *testing.T) {
 		t.Parallel()
 
 		schedule := NeverSchedule()
-		require.Equal(t, time.Unix(1<<63-1, 0), schedule.Next(time.Now()))
+		require.Equal(t, time.Unix(1<<63-62135596801, 999999999), schedule.Next(time.Now()))
 	})
 }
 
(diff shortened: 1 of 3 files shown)


periodic_job_test.go
```diff
@@ -18,7 +18,7 @@ func TestNeverSchedule(t *testing.T) {
 		t.Parallel()
 
 		schedule := NeverSchedule()
-		require.Equal(t, time.Unix(1<<63-1, 0), schedule.Next(time.Now()))
+		require.Equal(t, time.Unix(1<<63-62135596801, 999999999), schedule.Next(time.Now()))
```
bgentry, line 21: Maybe we should add one more test here to verify that it’s more than 1000 years in the future or something like that?

bgentry (approved):
Awesome, thank you for the fix and welcome back @krhubert 🙏🏻 could you please add a “fixed” changelog entry for this? We plan to ship a release today and can get this included.

periodic_job_test.go
krhubert: I'll add that plus check if the next schedule is  not before time.Now

krhubert: Also I think there might be the same issue with PeriodicInterval with a large duration. I'll take a look at this

krhubert: Done

krhubert:
Good to see you @bgentry ! I added a note to changelog's fixed section

bgentry:
One final request @krhubert, could you please add yourself to the CLA when you have a minute? I was thinking you were already on there but apparently not! https://github.com/riverqueue/rivercla

PR #762, opened 2025-02-15
add a test/race target to run all tests w/ race detector

bgentry:


Makefile
```diff
@@ -57,6 +57,14 @@ define test-target
 endef
 $(foreach mod,$(submodules),$(eval $(call test-target,$(mod))))
 
+.PHONY: test/race
+test/race:: ## Run test suite for all submodules with race detector
+define test-race-target
+    test/race:: ; cd $1 && go test ./... -p 1 -race
+endef
```
brandur, line 64: What's going on here?

bgentry: Not sure I follow your question. It's the same as the above `test` target, except for running all targets with the race detector (as CI does). Unless I messed something up in the copy/pasta?

brandur: Ah my bad — I didn't realize that this was all in there already. Damn, I can't read any of this make stuff.

brandur (approved):

Issue #761, opened 2025-02-15
Recent CI flakiness with DB pool acquire timeouts


bgentry:
I've seen a bunch of these in the last few days: https://github.com/riverqueue/river/actions/runs/13346849278/job/37278355065?pr=760

```
--- FAIL: TestDriverRiverPgxV5 (33.34s)
    --- FAIL: TestDriverRiverPgxV5/GetMigrationLines ([10](https://github.com/riverqueue/river/actions/runs/13346849278/job/37278355065?pr=760#step:7:11).00s)
        riverdrivertest.go:55: Failed to acquire pool for test DB: context deadline exceeded
FAIL
```

At a glance I don't see how this could have been caused by any specific recent changes but it's probably in part a sign that we need to refocus on speeding up the test suite soon.

brandur:
The schema-based testing should resolve this one I think. I've been running `-race` at high iteration counts and I haven't seen this problem in a really long time.

PR #763, opened 2025-02-15
test a broader CI matrix to isolate potential v1.24 issue

bgentry:
I'm seeing if #761 happens to be triggered by Go v1.24. Testing against 1.24 + 1.23 in the matrix here to see if it reveals any differences.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,6 +24,7 @@ jobs:
       # This ensures that we're really running the Go version in the CI matrix
       # rather than one that the Go command has upgraded to automatically.
       GOTOOLCHAIN: local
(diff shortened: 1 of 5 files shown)


bgentry:
@brandur ok so I think we have confirmation that #761 is specifically showing up with Go v1.24. I did 5 full runs and 4 of them had a v1.24 failure, 0 of them had a v1.23 failure.

brandur:
Weird ...

Alright, something may have changed in Go 1.24, but at the end of the day it's still a River thing I think ... I've already run our (much larger) test suite at work hundreds of times on Go 1.24 and it's fine. Probably a byproduct of the test suite's novel approach to database management again.

PR #766, opened 2025-02-17
rivertest.Worker: Use job executor, always insert jobs, require tx

bgentry:
This updates the `rivertest.Worker` type (#753) to leverage the extracted job executor logic from #768 and to avoid some of the downsides with the first pass of that design (partly outlined in #765), particularly the fact that certain database-dependent operations like `JobCompleteTx` couldn't be accurately simulated without hitting the database.

As part of these changes, test jobs are always inserted into the database and the test worker methods always require a transaction to be used for additional operations on that job (such as completions). The return types were also altered so that an extensible result struct is returned in addition to any legitimate worker errors that occur. The struct contains the kind of event that occurred from execution (completion, snoozing, cancellation, etc) so that it's easy to assert against the result, even if a special not-really-an-error like snooze was returned.

To make the tests line up for this, I had to make corrections to the job insert queries to allow them to properly set the created time in a similar fashion to how they were already setting the scheduled time. This allows stub clocks to be used in tests and ensures they're properly respected throughout the execution process.

### DIFF
--- a/rivertest/worker.go
+++ b/rivertest/worker.go
@@ -58,6 +58,8 @@ func NewWorker[T river.JobArgs, TTx any](tb testing.TB, driver riverdriver.Drive
 	tb.Helper()
 
 	config = config.WithDefaults()
+	config.Test.DisableUniqueEnforcement = true
+
 	client, err := river.NewClient(driver, config)
 	if err != nil {
 		tb.Fatalf("failed to create client: %s", err)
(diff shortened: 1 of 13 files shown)


rivertest/worker.go
```diff
@@ -70,86 +74,129 @@ func NewWorker[T river.JobArgs, TTx any](tb testing.TB, driver riverdriver.Drive
...
 
-func (w *Worker[T, TTx]) workJobExec(ctx context.Context, tb testing.TB, exec riverdriver.Executor, job *river.Job[T]) error {
+func (w *Worker[T, TTx]) workJob(ctx context.Context, tb testing.TB, tx TTx, job *rivertype.JobRow) error {
 	tb.Helper()
 
+	exec := w.client.Driver().UnwrapExecutor(tx)
+	subscribeCh := make(chan []jobcompleter.CompleterJobUpdated, 1)
+	archetype := riversharedtest.BaseServiceArchetype(tb)
+	completer := jobcompleter.NewInlineCompleter(archetype, exec, w.client.Pilot(), subscribeCh)
+	tb.Cleanup(completer.Stop)
```
brandur, line 129: This job completer is technically started, so I wonder if we should just skip stopping it too to save the cleanup hook.

rivertest/worker.go
```diff
@@ -70,86 +74,129 @@ func NewWorker[T river.JobArgs, TTx any](tb testing.TB, driver riverdriver.Drive
...
 }
 
-func (w *Worker[T, TTx]) workJobExec(ctx context.Context, tb testing.TB, exec riverdriver.Executor, job *river.Job[T]) error {
+func (w *Worker[T, TTx]) workJob(ctx context.Context, tb testing.TB, tx TTx, job *rivertype.JobRow) error {
 	tb.Helper()
 
+	exec := w.client.Driver().UnwrapExecutor(tx)
+	subscribeCh := make(chan []jobcompleter.CompleterJobUpdated, 1)
+	archetype := riversharedtest.BaseServiceArchetype(tb)
+	completer := jobcompleter.NewInlineCompleter(archetype, exec, w.client.Pilot(), subscribeCh)
```
brandur, line 128: Would be a very minor saving, but I wonder if we should let this tolerate a `nil` subscribe channel so we can skip the allocation and the frontmatter code.

brandur (commented):
@bgentry Yep, this looks great man! I've been wanting to packag-ify job executor forever anyway, and this is a good excuse to get it done.

You mentioned potentially just removing the non-tx work variants — IMO that's a good idea given that it'd be kind of weird to leave side effects like a new row on a raw pool.

And looking at the API again, returning like a result struct like you mentioned might not be a bad idea regardless of what we put in it for future compatibility if nothing else.

rivertest/worker.go
bgentry: I actually ended up leveraging this in the updated version in order to get the post-completion job row for the new return struct. We might consider refactoring it as there's a handful of channels at play, lmk once you get to review the new version (coming shortly).

bgentry:
@brandur aside from missing a changelog entry, this should be ready to go! Flaky test DB issues are still on my radar and I think are probably fixable even with the current design.

rivertest/worker_test.go
```diff
@@ -58,48 +80,43 @@ func TestWorker_Work(t *testing.T) {
 
 			return nil
 		})
-		tw := NewWorker(t, driver, config, worker)
-		require.NoError(t, tw.Work(context.Background(), t, testArgs{Value: "test"}, nil))
+		tw := NewWorker(t, bundle.driver, bundle.config, worker)
+		res, err := tw.Work(context.Background(), t, bundle.tx, testArgs{Value: "test"}, nil)
```
brandur, line 84: Maybe replace this `context.Background()` with `ctx` since it's broadly defined now, and it cleans up a lot of visual noise.

brandur (approved):
Great. LGTM.

rivertest/worker_test.go
bgentry: Done, did that in the whole file 👍

PR #768, opened 2025-02-17
Extract job executor internal package

bgentry:
This will make it possible to build more advanced rivertest harnesses that simulate real job execution.

One of the main changes in here is to extract some of the top-level error types like `JobCancelError` into `rivertype` so that they can be referenced both from the `internal/jobexecutor` package and the main `river` package. The only way to make this work well and continue to have the errors be considered "equal" by `errors.Is` / `errors.As` is for them to be the same underlying type. As such error types like `river.JobCancelError` are now aliases to those `rivertype.JobCancelError`. I don't think this should cause problems, but thought it should be called out.

I also pulled in a standalone commit to add `AttemptedBy` to the `JobUpdate` query that's used by tests. Not because it's related to the above, but because it's self-contained and this will help keep the next PR smaller.

Extracted from #766.


brandur (approved):
Great. WFM.

PR #769, opened 2025-02-18
Test DB pool fixes

bgentry:
Tested the heck out of this on #763 until I figured out a few issues:

1. We were allowing `NUM_CPU` conns to be created in _each_ of the pools for the `testdb.Manager`, meaning it was actually `NUM_CPU * NUM_CPU` connections. That's on top of the `NUM_CPU` connections being allocated for the global `TestTx` pool. I reduced this to a fixed limit of 4.
2. We didn't have a minimum set on the pools. There's no reason for any of these pooled test DBs to drop below 1, so I set that as the minimum.
3. We were always closing pools after every use, which is wasteful and inefficient. After digging, we only had a single test where we were intentionally closing the pool (one for the notifier). Rather than always closing the pool after each test, do a ping operation to make sure it's not closed. If it is closed, then it can be recreated.

The combination of these seems to have caused the flakiness issues that got worse with 1.24. Ultimately we were putting a ton of churn on Postgres conns and creating a lot of them at once. With these changes, not only do the tests pass consistently but the time for the `river` package on Actions seems to drop from ~40 seconds to ~31 seconds.

### DIFF
--- a/internal/cmd/testdbman/main.go
+++ b/internal/cmd/testdbman/main.go
@@ -139,11 +139,7 @@ func createTestDatabases(ctx context.Context, out io.Writer) error {
 		return nil
 	}
 
-	// This is the same default as pgxpool's maximum number of connections
-	// when not specified -- either 4 or the number of CPUs, whichever is
-	// greater. If changing this number, also change the similar value in
-	// `riverinternaltest` where it's duplicated.
-	dbNames := generateTestDBNames(max(4, runtime.NumCPU()))
+	dbNames := generateTestDBNames(runtime.GOMAXPROCS(0))
 
 	for _, dbName := range dbNames {
 		if err := createDBAndMigrate(dbName); err != nil {
--- a/internal/riverinternaltest/riverinternaltest.go
+++ b/internal/riverinternaltest/riverinternaltest.go
@@ -287,6 +287,8 @@ func WrapTestMain(m *testing.M) {
 	// NUM_CPU pools, each with NUM_CPU connections, and that's a lot of
 	// connections if there are many CPUs.
 	poolConfig.MaxConns = 4
+	// Pre-initialize 1 connection per pool.
+	poolConfig.MinConns = 1
 
 	var err error
 	dbManager, err = testdb.NewManager(poolConfig, int32(runtime.GOMAXPROCS(0)), nil, TruncateRiverTables) //nolint:gosec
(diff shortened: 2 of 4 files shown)


internal/cmd/testdbman/main.go
```diff
@@ -139,11 +139,7 @@ func createTestDatabases(ctx context.Context, out io.Writer) error {
 		return nil
 	}
 
-	// This is the same default as pgxpool's maximum number of connections
-	// when not specified -- either 4 or the number of CPUs, whichever is
-	// greater. If changing this number, also change the similar value in
-	// `riverinternaltest` where it's duplicated.
-	dbNames := generateTestDBNames(max(4, runtime.NumCPU()))
+	dbNames := generateTestDBNames(runtime.GOMAXPROCS(0))
```
bgentry, line 142: I'm not sure we actually want to rely on this setting since it could be removed someday. However it's at least somewhat dynamic and can be overridden, whereas `runtime.NumCPU()` is fixed. My thought was this option is just slightly more reliable if somebody tweaks the `GOMAXPROCS` and still wants things to work here.

Not set on it though.

internal/riverinternaltest/riverinternaltest.go
```diff
@@ -282,8 +280,16 @@ func TruncateRiverTables(ctx context.Context, pool *pgxpool.Pool) error {
 // amongst all packages. e.g. Configures a manager for test databases on setup,
 // and checks for no goroutine leaks on teardown.
 func WrapTestMain(m *testing.M) {
+	poolConfig := DatabaseConfig("river_test")
+	// Use a smaller number of conns per pool, because otherwise we could have
+	// NUM_CPU pools, each with NUM_CPU connections, and that's a lot of
+	// connections if there are many CPUs.
```
bgentry, line 286: Actually realized this issue because in some of my tests I was hitting the 100 conn limit on my local Postgres, at least once I stopped closing every pool after every test. 14 cores * 14 conns per pool = 196 potential conns.

I don't think any given test ever needs to use more than 4 in the pool.

brandur (approved):
Noice!

bgentry:
Unfortunately I don't think this is a total fix, I've still been seeing occasional errors (though less frequently). But still I think each of these changes is good, so it gets us closer. Will keep chasing down the remaining cause.

PR #774, opened 2025-02-19
remove irrelevant frames from panic traces

bgentry:
While testing the `rivertest.Worker`, I realized the panic traces we're sending to the panic handler have initial frames that reflect the code that actually generates the trace, rather than starting with the user's code that actually panicked:

```
--- FAIL: TestChargeCreditCard (0.00s)
    --- FAIL: TestChargeCreditCard/Success (7.33s)
        logger.go:256: time=2025-02-18T20:27:37.080-06:00 level=ERROR msg="JobExecutor: panic recovery; possible bug with Worker" job_id=8374 kind=ChargeCreditCard panic_val="runtime error: invalid memory address or nil pointer dereference"
        logger.go:256: time=2025-02-18T20:27:37.081-06:00 level=ERROR msg="JobExecutor: Job panicked" error="runtime error: invalid memory address or nil pointer dereference" job_id=8374 job_kind=ChargeCreditCard
        worker.go:184: panic: runtime error: invalid memory address or nil pointer dereference
            goroutine 36 [running]:
            runtime/debug.Stack()
            	/opt/homebrew/Cellar/go/1.24.0/libexec/src/runtime/debug/stack.go:26 +0x64
            github.com/riverqueue/river/internal/jobexecutor.(*JobExecutor).execute.func1()
            	/Users/bgentry/River/river/internal/jobexecutor/job_executor.go:134 +0x380
            panic({0x1035d3c40?, 0x103c0f530?})
            	/opt/homebrew/Cellar/go/1.24.0/libexec/src/runtime/panic.go:787 +0xf0
            github.com/riverqueue/riverdemo/worker.(*SyntheticBaseWorker).syntheticSleepOrError(0x0, {0x1036c44c0, 0x14000056380}, {0x1036bb5e0, 0x140000b0320})
            	/Users/bgentry/River/riverdemo/worker/synthetic_base_job.go:22 +0x50
            github.com/riverqueue/riverdemo/worker.(*ChargeCreditCard).Work(0x1400000c090, {0x1036c44c0, 0x14000056380}, 0x1400004a310)
            	/Users/bgentry/River/riverdemo/worker/charge_credit_card.go:33 +0x19c
            github.com/riverqueue/river/rivertest.(*wrapperWorkUnit[...]).Work(0x1036c4320, {0x1036c44c0, 0x140000562a0})
            	/Users/bgentry/River/river/rivertest/worker.go:281 +0x60
[...]

PR #775, opened 2025-02-19
rivertest.Worker fixes: handle nil config, rename Kind/EventKind, log panic trace

bgentry:
I wrote some tests for the demo app using the test worker (#766) and found a few issues:

1. It panics with a `nil` `*river.Config`. I fixed this by making `Config.WithDefaults` handle a nil pointer so no change was necessary in the `rivertest.NewWorker` caller.
2. The `Kind` field in `rivertest.WorkResult` felt unnatural as I was writing assertions for it like `require.Equal(t, river.EventKindJobCompleted, result.Kind)`. I renamed this to `EventKind` which I still don't love but it feels at least a little more appropriate to me.
3. When I accidentally made my worker panic during the course of developing tests, I realized the current output (just the panic value) was not very helpful for debugging. I adjusted this to also log the full trace, which is also how I realized #774 was necessary.

Side note on (3) above: I realized I couldn't actually use `rivertest.Worker` to write a test for a worker that _intentionally_ panics, because its panic handler calls `tb.Fatal`. This is probably not a very likely issue for real users, and it's probably a good _default_ behavior, but I won't be surprised if we feel compelled to add an option to the test worker type so that it avoid calling fatal and instead surface an error or call `tb.Error`. Wanted to run it by you in case you had other thoughts on this. For example, we could just switch it to `tb.Error` because `Fatal` is quite presumptive?

### DIFF
--- a/rivertest/worker_test.go
+++ b/rivertest/worker_test.go
@@ -117,7 +117,7 @@ func TestWorker_Work(t *testing.T) {
 		tw := NewWorker(t, bundle.driver, bundle.config, worker)
 		res, err := tw.Work(ctx, t, bundle.tx, testArgs{Value: "test"}, nil)
 		require.NoError(t, err)
-		require.Equal(t, river.EventKindJobCompleted, res.Kind)
+		require.Equal(t, river.EventKindJobCompleted, res.EventKind)
 	})
 
 	t.Run("Reusable", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


brandur:
> Wanted to run it by you in case you had other thoughts on this. For example, we could just switch it to `tb.Error` because `Fatal` is quite presumptive?

Yeah, honestly I didn't see this part on my review, but IMO the panic shouldn't be swallowed. Might even be best not to have any error/panic handlers and just let users do what they want with those values.

rivertest/worker_test.go
```diff
@@ -22,6 +22,40 @@ type testArgs struct {
...
+		t.Helper()
+
+		return &testBundle{
+			config: &river.Config{ID: "rivertest-worker"},
+			driver: riverpgxv5.New(nil),
+			tx:     riverinternaltest.TestTx(ctx, t),
+		}
+	}
+
+	t.Run("HandlesANilRiverConfig", func(t *testing.T) {
```
brandur, line 46: Nit, but found the casing a bit awkward to read. Really doesn't need the "A" anyway:

```suggestion
	t.Run("HandlesNilRiverConfig", func(t *testing.T) {
```

brandur (approved):

bgentry:
> Yeah, honestly I didn't see this part on my review, but IMO the panic shouldn't be swallowed. Might even be best not to have any error/panic handlers and just let users do what they want with those values.

An error handler is necessary in order to capture and return the error via the test helper—otherwise it'd be lost and only recorded in the DB.

I've restructured this so there's a `PanicError` type that gets surfaced as the return `error` from `Work` / `WorkJob`. We're no longer using `tb.Fatal` when working a test job except for a couple of internal error cases that shouldn't ever happen.

PR #777, opened 2025-02-19
improve driver test coverage for InsertFastMany, fix databasesql unique key issue

### LINKED ISSUE #779: riverdatabasesql: invalid UTF8 byte sequence for unique_key
> @bgentry I appear to now be getting errors like this on insertion of unique jobs:
> 
> `ERROR: invalid byte sequence for encoding "UTF8": 0xf4 0x8d 0x61 0x24 (SQLSTATE 22021)` 

 _Originally posted by @wydengyre in [#650](https://github.com/riverqueue/river/issues/650#issuecomment-2669095897)_

While fixing #650, a different issue was introduced. This one escaped the new test coverage because those tests were using real UTF8 byte strings instead of generic bytes.

bgentry:
When the unique key was a binary non-UTF8 []byte, it was generating an encoding error in the `database/sql` driver. The fix is to encode it, similar to what pq does. I couldn't find a way to do this via pq or pgx without reimplementing this code block, but would love a way to do that if possible.

Also improve test coverage to make sure we are explicitly testing both intentionally set unique key + unique states, as well as intentionally omitted.

Fixes #779.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fix error message on unsuccessful client subscribe that erroneously referred to "Workers" not configured. [PR #771](https://github.com/riverqueue/river/pull/771).
+- Fix an issue with encoding unique keys in riverdatabasesql driver. [PR #777](https://github.com/riverqueue/river/pull/777).
 
 ## [0.17.0] - 2025-02-16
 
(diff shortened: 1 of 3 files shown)


bgentry:
Updated to fix #779 and hopefully finally get unique jobs working again in database/sql 🤞

bgentry:
@brandur I should mention on this one. I tried to solve it using sqlc overrides, but unfortunately I don't think there's any way to target only the `[]bytea` types that these inputs use, without also targeting the non array bytea columns coming back from other queries. Obviously I'm open to anything else that works and ideally doesn't require maintaining this extra code.

brandur (approved):

PR #782, opened 2025-02-20
Limit recorded output to 32MB in size

bgentry:
It'd be a bad idea to get anywhere close to Postgres' jsonb limit of 255 MB. We choose 32MB as an arbitrary cap, though you really shouldn't get anywhere close to this.


brandur (approved):
Ah doh, sorry must've archived this one by accident. Too many issues/PRs flying around. LGTM.

bgentry:
@brandur np, I have been on a bit of a tear the last week or two 😄

Issue #752, opened 2025-02-10
Job timeout applied after middleware


bgentry:
While reading through the executor for another change, I realized that as of #632 / #584 we're applying the job timeout _after_ stepping through all the middleware:

https://github.com/riverqueue/river/blob/fc667eb33b2427f28ca335f5287610edaa2b7e93/job_executor.go#L209-L227

I'm not sure if this was an intentional decision and I couldn't find any conversations where it was mentioned, so I thought we should at least talk about it. Do you think this is the right place for it, or should the timeout happen _before_ the middleware stack gets called? I was thinking of a worst case scenario where a middleware got stuck or took awhile on a blocking operation and the job ended up running far longer than either the job's timeout or the client's default timeout.

brandur:
Hmm, so I didn't consider it one way or the other super deeply admittedly, but including the middleware in the timeout doesn't seem super obviously more or less correct to me.

An argument could be made that that when setting work timeouts, users are trying to set a timeout specifically on _their_ code as opposed to River's internal code, which to their eye would be an implementation detail. So similarly how the timeout doesn't include the time it takes to complete a job, it doesn't include middleware, which is largely made up of other internal stuff.

bgentry:
The risk I see is that as it currently is, the middleware have no timeout whatsoever applied to them. Beyond that I don't think it matters a lot either way, but I do worry about the risk of a middleware getting stuck and the job just running forever despite whatever timeouts are configured.

brandur:
Yeah I suppose that could be a danger depending on what you're trying to do in the middleware. Probably a separate timeout like you'd find in [`http.Transport`](https://pkg.go.dev/net/http#Transport) (each phase of a request can be configured separately) would be as defensible as mixing in the same timeout being used for the job's work phase.

PR #783, opened 2025-02-21
Job executor: Unmarshal job args late, after middlewares have run

brandur:
Here, modify the job executor so that instead of exclusively invoking a
work unit's `UnmarshalJSON` implementation before starting to execute
the stack of work middleware, invoke it much later from within the
executor's `doInner` implementation. This gives middlewares the
opportunity to modify args before unmarshaling, thereby enabling use
cases like args encryption.

This does have the downside of having to modify the worker interface's
`Middleware` function from taking a generic job:

    type Worker[T JobArgs] interface {
        Middleware(job *Job[T]) []rivertype.WorkerMiddleware

To taking a `JobRow` instead:

    type Worker[T JobArgs] interface {
        Middleware(job *rivertype.JobRow) []rivertype.WorkerMiddleware

Because the full generic `Job[T]` is not yet available by the time we
need to extract a middleware stack.

### DIFF
--- a/internal/jobexecutor/job_executor.go
+++ b/internal/jobexecutor/job_executor.go
@@ -183,16 +183,21 @@ func (e *JobExecutor) execute(ctx context.Context) (res *jobExecutorResult) {
 		return &jobExecutorResult{Err: &rivertype.UnknownJobKindError{Kind: e.JobRow.Kind}, MetadataUpdates: metadataUpdates}
 	}
 
-	if err := e.WorkUnit.UnmarshalJob(); err != nil {
-		return &jobExecutorResult{Err: err, MetadataUpdates: metadataUpdates}
-	}
+	doInner := execution.Func(func(ctx context.Context) error {
+		if err := e.WorkUnit.UnmarshalJob(); err != nil {
+			return err
+		}
 
-	doInner := execution.MiddlewareChain(e.GlobalMiddleware, e.WorkUnit.Middleware(), e.WorkUnit.Work, e.JobRow)
-	jobTimeout := valutil.FirstNonZero(e.WorkUnit.Timeout(), e.ClientJobTimeout)
-	ctx, cancel := execution.MaybeApplyTimeout(ctx, jobTimeout)
-	defer cancel()
+		jobTimeout := valutil.FirstNonZero(e.WorkUnit.Timeout(), e.ClientJobTimeout)
+		ctx, cancel := execution.MaybeApplyTimeout(ctx, jobTimeout)
+		defer cancel()
 
-	return &jobExecutorResult{Err: doInner(ctx), MetadataUpdates: metadataUpdates}
+		return e.WorkUnit.Work(ctx)
+	})
+
+	executeFunc := execution.MiddlewareChain(e.GlobalMiddleware, e.WorkUnit.Middleware(), doInner, e.JobRow)
+
+	return &jobExecutorResult{Err: executeFunc(ctx), MetadataUpdates: metadataUpdates}
 }
 
 func (e *JobExecutor) invokeErrorHandler(ctx context.Context, res *jobExecutorResult) bool {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+⚠️ Version 0.19.0 has minor breaking changes for the `Worker.Middleware`, introduced fairly recently in 0.17.0. We tried not to make this change, but found the existing middleware interface insufficient to provide the necessary range of functionality we wanted, and this is a secondary middleware facility that won't be in use for many users, so it seemed worthwhile.
+
+### Changed
+
 - The `river.RecordOutput` function now returns an error if the output is too large. The output is limited to 32MB in size. [PR #782](https://github.com/riverqueue/river/pull/782).
+- **Breaking change:** The `Worker` interface's `Middleware` function now takes a `JobRow` parameter instead of a generic `Job[T]`. This was necessary to expand the potential of what middleware can do: by letting the executor extract a middleware stack from a worker before a job is fully unmarshaled, the middleware can also participate in the unmarshaling process. [PR #783](https://github.com/riverqueue/river/pull/783).
 
 ## [0.18.0] - 2025-02-20
 
--- a/client_test.go
+++ b/client_test.go
@@ -661,8 +661,8 @@ func Test_Client(t *testing.T) {
 				require.Equal(t, "called", ctx.Value(privateKey("middleware")))
 				return nil
 			},
-			middlewareFunc: func(job *Job[callbackArgs]) []rivertype.WorkerMiddleware {
-				require.Equal(t, "middleware_test", job.Args.Name, "JSON should be decoded before middleware is called")
+			middlewareFunc: func(job *rivertype.JobRow) []rivertype.WorkerMiddleware {
+				require.Equal(t, "callback", job.Kind)
 
 				return []rivertype.WorkerMiddleware{
 					&overridableJobMiddleware{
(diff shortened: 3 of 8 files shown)


internal/jobexecutor/job_executor.go
```diff
@@ -183,16 +184,31 @@ func (e *JobExecutor) execute(ctx context.Context) (res *jobExecutorResult) {
...
+		return e.WorkUnit.Work(ctx)
+	})
+
+	executeFunc := execution.MiddlewareChain(e.GlobalMiddleware, e.WorkUnit.Middleware(), doInner, e.JobRow)
 	jobTimeout := valutil.FirstNonZero(e.WorkUnit.Timeout(), e.ClientJobTimeout)
 	ctx, cancel := execution.MaybeApplyTimeout(ctx, jobTimeout)
 	defer cancel()
 
-	return &jobExecutorResult{Err: doInner(ctx), MetadataUpdates: metadataUpdates}
+	return &jobExecutorResult{Err: executeFunc(ctx), MetadataUpdates: metadataUpdates}
```
bgentry, line 211: Huh, I think I fixed #752 as part of 0132e537ab49246431f4b021f4287ab53ee75f48 / #753 without explicitly calling it out. Looks like it's now applying the timeout to the entire chain.

internal/jobexecutor/job_executor.go
```diff
@@ -183,16 +184,31 @@ func (e *JobExecutor) execute(ctx context.Context) (res *jobExecutorResult) {
...
+	doInner := execution.Func(func(ctx context.Context) error {
+		// The middleware stack is allowed an opportunity to modify EncodedArgs.
+		// If we do detect a modification, unmarshal args again.
+		if bytes.Compare(encodedArgs, e.JobRow.EncodedArgs) != 0 {
+			if err := e.WorkUnit.UnmarshalJob(); err != nil {
+				return err
+			}
+		}
+
+		return e.WorkUnit.Work(ctx)
```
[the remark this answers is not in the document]
brandur: Yeah fair points. I went back over this again, and yeah, this approach seems clearly wrong to me now.

I switched to a new version that I think is much more clearly right — basically, all unmarshaling migrates into `doInner`, and this tends to be okay because the middleware only have access to the `JobRow` anyway, so they really don't need the unmarshaled `Job` so early.

The downside is that the `Worker` middleware function has to take a `JobRow` instead of a `Job`:

``` diff
-	Middleware(job *Job[T]) []rivertype.WorkerMiddleware
+	Middleware(job *rivertype.JobRow) []rivertype.WorkerMiddleware
```

TBH though, that seems more consistent with the middleware API anyway.

It is less consistent with the other `Worker` interface functions though (they tend to take `Job[T]`). We could either leave it like that, or maybe it'd be better to separate the `Middleware` function out into its own interface given that it's probably far less used in a general sense anyway.

Want to take a look at my latest commit and LMK what you think?

It might be tempting not to change any of this stuff, but having played around with the encryption side quite a bit, I don't think there's any way to implement it as middleware given the current API. Given that an encryption middleware is about as basic of a middleware use as one can imagine, it seems to me that something's got to be changed here.

internal/jobexecutor/job_executor.go
brandur: True. Latest push though does change this back to extracting a timeout from within `doInner`, heh.

internal/jobexecutor/job_executor.go
```diff
@@ -183,16 +183,21 @@ func (e *JobExecutor) execute(ctx context.Context) (res *jobExecutorResult) {
...
+		ctx, cancel := execution.MaybeApplyTimeout(ctx, jobTimeout)
+		defer cancel()
+
+		return e.WorkUnit.Work(ctx)
+	})
 
-	doInner := execution.MiddlewareChain(e.GlobalMiddleware, e.WorkUnit.Middleware(), e.WorkUnit.Work, e.JobRow)
-	jobTimeout := valutil.FirstNonZero(e.WorkUnit.Timeout(), e.ClientJobTimeout)
-	ctx, cancel := execution.MaybeApplyTimeout(ctx, jobTimeout)
-	defer cancel()
```
bgentry, line 193: A notable and maybe unintentional change here: the context passed to middleware is no longer being cancelled at any point, because you're only canceling the innermost context given to the worker's `Work()`. Maybe this should be resolved by adding a `context.WithCancel()` and `defer cancel()` up at the top of `execute()` so ensure that the context is 100% always canceled for all layers once execution is done?

bgentry (approved):
I think this makes sense. Obviously it's not great to break this interface but IMO we've found a worthy reason to do so given the use cases that can't be adequately solved without the change.

Needs a changelog callout and I think a fix for context cancellation (should that have a test?), but otherwise LGTM :v:

internal/jobexecutor/job_executor.go
brandur: Hmm, okay, I think this might be a big difference in how you and I understand contexts/cancellations, but I still don't feel great about littering random `context.WithCancel()`s all over the place where the `cancel()` is never intended for use except on `defer` return.

Think about it this way: the only time `cancel()` is called is if we've _successfully_ returned from middleware functions. If any middleware gets stuck somewhere waiting, the `cancel` will never be called anyway because it's only called on `defer`.

So the only thing we're protecting against by adding more cancels is cases where a middleware was able to return successfully, but then improperly spawned off some extraneous goroutines but which will _properly_ respond to a cancelled context.

IMO, it shouldn't really be a framework's job to cancel that kind of stuff. I'd even go so far as to say that it'd be better if it didn't, because that'd potentially let the problem be found more easily and fixed, whereas context cancellations from River might paper over the problem for a long time in this one place, but letting it maybe still occur somewhere else.

I re-read the docs for [WithCancel](https://pkg.go.dev/context?utm_source=godoc#WithCancel) in case I've been missing something big all these years, but I can't find anything in there suggesting use of `WithCancel` in places where `cancel()` is not intended to be used (e.g. see the example there).

I looked into putting in some sort global context cancellation of another kind of middlewares, but I couldn't find any easy resolution for this one right now because no cancellation should be applied to `doInner`, and `doInner` is intrinsically linked to the rest of the middleware invocation.

IMO again, it might be better not to have one anyway, and instead encourage each middleware to select appropriate timeouts _if_ they need it. i.e. Maybe it makes sense for a concurrency limiter middleware to have one, but an encryption middleware should have no context timeout at all (one will be ignored it it's set anyway).

Thoughts on all that? Okay leaving this out for the time being and continuing to refine this as we go?

bgentry: I guess I see it as being conceptually the same as the context on an incoming `http.Request`. From https://pkg.go.dev/net/http#Request.Context :

> For incoming server requests, the context is canceled when the client's connection closes, the request is canceled (with HTTP/2), or when the ServeHTTP method returns.

The idea being that as soon as user-controlled code is completed, the context is canceled to prevent any potential leaked resources. I see the same pattern is followed for gRPC.

To me it makes sense given all the ways in which a context is scoped to a specific job being executed, and it feels like a low cost nicety that might save users some pain vs having to remember to do it themselves when spawning concurrent goroutines in a job. But it’s not a hill I’ll die on, so your call.

brandur: > I guess I see it as being conceptually the same as the context on an incoming `http.Request`. From https://pkg.go.dev/net/http#Request.Context :

K, I view this as quite a different situation because in the case of an HTTP handler, something might be happening out of band (i.e. the original request that the handler is trying to serve's connection is closed) which makes work being done by a handler superfluous, and therefore it's a very useful feature for the context to be cancelled, because you're preventing additional work that no longer needs to be done.

In this case, the `cancel()` would only happen on `return`, so nothing would ever happen out-of-band. Unless I'm  missing something here, there's no circumstances under which this would save additional work (as mentioned above, it would necessitate the middleware stack to return before `cancel()` could ever be invoked).

Going to merge as is for now, but we've got to retool how timeouts work on middleware stacks anyway, so let's get into that next.

brandur:
Added changelog entry and retooled the PR title/description to match changes since I opened this.

Thanks!

4almrasla:
- - [ ]

PR #787, opened 2025-03-01
Minor middleware improvements: Reduce allocations + docs

brandur:
Just a couple small middleware improvements:

* Retool `MiddlewareChain` so that in case of no worker-specific
  middleware (which should be the overwhelmingly common case), don't
  bother with a new slice allocation. Just use the global middleware
  slice that we already have.

* Add missing docs for `WorkerInsertMiddlewareDefaults`.


bgentry (approved):
:shipit:

brandur:
thx.

Issue #788, opened 2025-03-01
Bifurcated middleware kind of ugly + potentially fragile


brandur:
Just one I noticed while trying to implement a middleware that does encryption. Here's some lines from the test suite that show roughly what it's like to use this middleware:

    middleware := riverencrypt.NewEncryptMiddleware(riversecretbox.NewSecretboxEncryptor(key))

    config := &river.Config{
        JobInsertMiddleware: []rivertype.JobInsertMiddleware{middleware},
        WorkerMiddleware:    []rivertype.WorkerMiddleware{middleware},
    }

The middleware has to go into both `JobInsertMiddleware` and `WorkerMiddleware`. This is pretty unsightly, but also makes things fragile -- if you added the middleware to one and forgot the other, it won't work.

Another related, but separate problem is that while workers can implement work-specific middleware, job args don't have their own variant of that, so something like a worker-specific encryption middleware is currently not possible.

I think there's a chance that maybe what we should have done here was to do something like `river.WorkerDefaults` like `MiddlewareDefaults` that lets a middleware struct implement run on insertion, run on work, or both, and which would allow for a more streamlined configuration like:

    config := &river.Config{
        Middleware: []rivertype.Middleware{middleware},
    }

And then maybe have a single `Middleware()` function on either `JobArgs` or `Worker` that would job/worker-specific middleware, so that once again, it wouldn't be possible to accidentally forget middleware from one location or the other.

Category-specific middleware does have some advantages like shallower stack traces in case of a problem (you might otherwise have 2x the middleware in your trace as everything got mixed together), but it kind of feels like what we have here isn't quite right ergonomically.

brandur:
@bgentry I wrote this up a few days ago, but mainly wanted to submit it now so I can reference it from another PR informationally.

brandur:
Middleware now supports a non-bifurcated version.

PR #789, opened 2025-03-01
Add "hooks" interface in addition to middleware

brandur:
Here, try an experiment that adds "hooks", a middleware-like concept, but one
which differs in subtle ways, and if done right will unlock functionality that
middleware doesn't make possible, like job-specific actions like encryption.

While experimenting anyway, we also try a variation on middleware configuration
that tries to improve the experience of using hooks to address #788.

The middleware interface is very generic, containing just one sentinel function
so that it's distinct from `interface{}` (and which would make any type
accidentally settable to it):

    type Hook interface {
        // IsAnyHook is a sentinel function to check that a type is implementing
        // Hook on purpose and not by accident (Hook would otherwise be an empty
        // interface). Hooks should embed river.HookDefaults to pick up an
        // implementation for this function automatically.
        IsAnyHook() bool
    }

But from there, we get into specific hook interfaces that run at various phases
in a job's lifecycle. These ones should look familiar from middleware:

    // HookInsertBegin is an interface to a hook that runs before job insertion.
    type HookInsertBegin interface {
        Hook

        InsertBegin(ctx context.Context, params *JobInsertParams) error
    }

    // HookWorkBegin is an interface to a hook that runs after a job has been locked
    // for work and before it's worked.
    type HookWorkBegin interface {
        Hook

        WorkBegin(ctx context.Context, job *JobRow) error
    }

The primary advantage of hooks is that they don't get layered onto the stack,
and especially when inserting rows, don't all need to share a single set of
hooks, letting each job in question run different hooks if desired.

By extension, this also means that unlike middleware, hooks cannot set anything
to context. This is the main practical distinction between the two concepts

By consolidating different phases into a single interface, we let logically
related hooks be configured much more cleanly. Whereas before we needed to
something like this to install a middleware to different lifecycle phases:

    middleware := riverencrypt.NewEncryptMiddleware(riversecretbox.NewSecretboxEncryptor(key))

    config := &river.Config{
        JobInsertMiddleware: []rivertype.JobInsertMiddleware{middleware},
        WorkerMiddleware:    []rivertype.WorkerMiddleware{middleware},
    }

We now get to do this instead:

    config := Config: river.Config{
        Hooks: []rivertype.Hook{
            riverencrypt.NewEncryptHook(riversecretbox.NewSecretboxEncryptor(key)),
        },

This is cleaner code-wise, but it also prevents accidental misuse where say an
encryptor was installed for insertion, but then forgotten for when jobs are
worked.

Similar to `JobArgsWithInsertOpts`, job args can implement `JobArgsWithHooks`
to add hooks for specific job kinds, and which will also tak effect for when
jobs are worked:

    type JobArgsWithHooks interface {
        // Hooks returns specific hooks to run for this job type. These will run
        // after the global hooks configured on the client.
        //
        // Warning: Hooks returned should be based on the job type only and be
        // invariant of the specific contents of a job. Hooks are extract by
        // instantiating a generic instance of the job even when a specific instance
        // is available, so any conditional logic within will be ignored. This is
        // done because although specific job information may be available in some
        // hook contexts like on InsertBegin, it won't be in others like WorkBegin.
        Hooks() []rivertype.Hook
    }

### DIFF
--- a/client.go
+++ b/client.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/riverqueue/river/internal/dblist"
 	"github.com/riverqueue/river/internal/dbunique"
+	"github.com/riverqueue/river/internal/hooklookup"
 	"github.com/riverqueue/river/internal/jobcompleter"
 	"github.com/riverqueue/river/internal/leadership"
 	"github.com/riverqueue/river/internal/maintenance"
@@ -166,6 +167,11 @@ type Config struct {
 	// Defaults to 1 minute.
 	JobTimeout time.Duration
 
+	// Hooks are functions that may activate at certain points during a job's
+	// lifecycle (see rivertype.Hook), installed globally. Jobs may have their
+	// own specific hooks by implementing the JobArgsWithHooks interface.
+	Hooks []rivertype.Hook
+
 	// Logger is the structured logger to use for logging purposes. If none is
 	// specified, logs will be emitted to STDOUT with messages at warn level
 	// or higher.
@@ -314,6 +320,7 @@ func (c *Config) WithDefaults() *Config {
 		FetchCooldown:               valutil.ValOrDefault(c.FetchCooldown, FetchCooldownDefault),
 		FetchPollInterval:           valutil.ValOrDefault(c.FetchPollInterval, FetchPollIntervalDefault),
 		ID:                          valutil.ValOrDefaultFunc(c.ID, func() string { return defaultClientID(time.Now().UTC()) }),
+		Hooks:                       c.Hooks,
 		JobInsertMiddleware:         c.JobInsertMiddleware,
 		JobTimeout:                  valutil.ValOrDefault(c.JobTimeout, JobTimeoutDefault),
 		Logger:                      logger,
@@ -427,6 +434,8 @@ type Client[TTx any] struct {
 	config               *Config
 	driver               riverdriver.Driver[TTx]
 	elector              *leadership.Elector
+	hookLookupByJob      *hooklookup.JobHookLookup
+	hookLookupGlobal     hooklookup.HookLookupInterface
 	insertNotifyLimiter  *notifylimiter.Limiter
 	notifier             *notifier.Notifier // may be nil in poll-only mode
 	periodicJobs         *PeriodicJobBundle
@@ -543,6 +552,8 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 	client := &Client[TTx]{
 		config:               config,
 		driver:               driver,
+		hookLookupByJob:      hooklookup.NewJobHookLookup(),
+		hookLookupGlobal:     hooklookup.NewHookLookup(config.Hooks),
 		producersByQueueName: make(map[string]*producer),
 		testSignals:          clientTestSignals{},
 		workCancel:           func(cause error) {}, // replaced on start, but here in case StopAndCancel is called before start up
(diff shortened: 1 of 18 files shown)


rivertype/river_type.go
```diff
@@ -270,6 +270,55 @@ type JobInsertParams struct {
...
+	InsertBegin(ctx context.Context, params *JobInsertParams) error
+}
+
+// HookWorkBegin is an interface to a hook that runs after a job has been locked
+// for work and before it's worked.
+type HookWorkBegin interface {
+	Hook
+
+	WorkBegin(ctx context.Context, job *JobRow) error
+}
```
bgentry, line 320: I'm not really vibing with the naming here. I assume the intention is to keep them sortable by event type (insert vs work) but the names feel unintuitive and not as descriptive as I'd hope. For example `BeforeInsert` feels obvious as to its timing (right before an insert), where as `InsertBegin` is not as obvious to me (it's not really the "beginning" of the operation").

IMO this is enough of a setback from the alternative names that it warrants breaking the convention of focusing on sort order & autocompletion as the top priority over legibility & intuitiveness.

bgentry (commented):
I definitely think this can work well as a pattern and it solves some different use cases than middleware, or at least solves some of them much better than middleware. I think it's worth pursuing for sure :v: Thanks for proofing it out.

rivertype/river_type.go
brandur: Hm, dunno about `BeforeInsert` — not only does it sort badly, but IMO it breaks naming convention with the middleware's `InsertMany`. Same if we did `BeforeWork` alongside the existing `Work`.

Generally I really like the words "begin" and "end" because they organize so nicely, and this gets important once you start introducing a lot of things. e.g.:

* `InsertBegin`
* `InsertEnd`
* `WorkBegin`
* `WorkEnd`

Before/after would feel a tad more natural, but are strictly worse for this kind of thing, so IMO the slightly better name isn't worth it enough:

* `InsertAfter`
* `InsertBefore`
* `WorkAfter`
* `WorkBefore`

Or even worse, using convention like `InsertBefore` would actually order things randomly compared to how they actually run (order would be 3, 1, 4, 2):

* `AfterInsert`
* `AfterWork`
* `BeforeInsert`
* `BeforeWork`

I'm open to alternatives, but don't want to throw the baby out with the bathwater — especially when you're implementing a hook with multiple implementations, it's a really great feature if the functions on it sort naturally in the same order that they run. Elegant, organized by default, and avoids ambiguity around what good convention should look like.

So if we're going to change it, it'd be nice to find a pair of words with similar properties. e.g. enter/exit, start/stop. (I think begin/end are better than those ones, but that sort of thing.)

brandur:
@bgentry Added a bunch of new tests for this one along with a memoization layer. I think I have some ideas on how to do something to help converge middleware a little more to improve the UX there somewhat too. Can I get your thoughts again on this one.

client.go
```diff
@@ -1510,6 +1521,24 @@ func (c *Client[TTx]) insertManyShared(
 	execute func(context.Context, []*riverdriver.JobInsertFastParams) ([]*rivertype.JobInsertResult, error),
 ) ([]*rivertype.JobInsertResult, error) {
 	doInner := func(ctx context.Context) ([]*rivertype.JobInsertResult, error) {
+		for _, params := range insertParams {
+			// TODO(brandur): This range clause and the one below it are
+			// identical, and it'd be nice to merge them together, but in such a
+			// way that doesn't require array allocation. I think we can do this
+			// using iterators after we drop support for Go 1.22.
```
bgentry, line 1528: Maybe this isn't worth being the one thing that pushes us to fully drop 1.22 support, but we're well within our "2 most recent Go versions" policy to support only 1.23+.

job.go
```diff
@@ -24,6 +24,19 @@ type JobArgs interface {
 	Kind() string
 }
 
+type JobArgsWithHooks interface {
+	// Hooks returns specific hooks to run for this job type. These will run
+	// after the global hooks configured on the client.
+	//
+	// Warning: Hooks returned should be based on the job type only and be
+	// invariant of the specific contents of a job. Hooks are extract by
```
bgentry, line 32:
```suggestion
	// invariant of the specific contents of a job. Hooks are extracted by
```

bgentry (approved):
Needs a changelog entry but otherwise lgtm :v: :shipit:

job.go
brandur: Ah good catch.

client.go
brandur: Yeah true. Didn't quite feel like I should mess with the Go version for this, but I'm sure something will end up bumping it intentionally or not pretty soon, so I'll take a look then.

brandur:
ty man! Added changelog and addressed feedback.

PR #793, opened 2025-03-06
Upgrade to sqlc 1.28.0

brandur:
Just a small one to upgrade to sqlc 1.28.0. I'd already upgraded for
other projects, and it ends up bleeding into River because all the file
headers get updated when I run `make generate`.


bgentry (approved):

brandur:
Thx!

PR #795, opened 2025-03-08
Implement job list with sqlc using templates

brandur:
Here, reimplement the job list API's internals using sqlc using the new
sqlc template system introduced in #794. Admittedly this one isn't quite
as big of a win as we expect the templating to be in some cases, but it
gets rid of a case of SQL code in a raw string, and lets us drop the
`JobListFields` driver interface function.

We don't add tests because the APIs are relatively well tested already,
but make some small modifications to the driver `JobList` test cases so
they're a little more thorough.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,14 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed
-
 ⚠️ Version 0.19.0 has minor breaking changes for the `Worker.Middleware`, introduced fairly recently in 0.17.0. We tried not to make this change, but found the existing middleware interface insufficient to provide the necessary range of functionality we wanted, and this is a secondary middleware facility that won't be in use for many users, so it seemed worthwhile.
 
 ### Changed
 
 - The `river.RecordOutput` function now returns an error if the output is too large. The output is limited to 32MB in size. [PR #782](https://github.com/riverqueue/river/pull/782).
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -7,14 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
-### Changed
-
 ⚠️ Version 0.19.0 has minor breaking changes for the `Worker.Middleware`, introduced fairly recently in 0.17.0. We tried not to make this change, but found the existing middleware interface insufficient to provide the necessary range of functionality we wanted, and this is a secondary middleware facility that won't be in use for many users, so it seemed worthwhile.
 
 ### Changed
 
 - The `river.RecordOutput` function now returns an error if the output is too large. The output is limited to 32MB in size. [PR #782](https://github.com/riverqueue/river/pull/782).
 - **Breaking change:** The `Worker` interface's `Middleware` function now takes a `JobRow` parameter instead of a generic `Job[T]`. This was necessary to expand the potential of what middleware can do: by letting the executor extract a middleware stack from a worker before a job is fully unmarshaled, the middleware can also participate in the unmarshaling process. [PR #783](https://github.com/riverqueue/river/pull/783).
+- `JobList` has been reimplemented to use sqlc. [PR #XXX](https://github.com/riverqueue/river/pull/XXX).
```
bgentry, line 16: Not even sure this needs to be in here as there's no user-facing impact, but if you keep it:

```suggestion
- `JobList` has been reimplemented to use sqlc. [PR #795](https://github.com/riverqueue/river/pull/795).
```

bgentry (approved):

CHANGELOG.md
brandur: Oops, thanks.

PR #796, opened 2025-03-08
Fix intermittent test in `sqlctemplate` relating to arbitrary map sorting

brandur:
Here, fix an intermittent test [1] that's come up _already_ since #795
was merged in. This is a problem that gets me a lot, which is that map
sorting in Go is not guaranteed, despite in tests small maps often
coming out in the order that you're looking for so you don't notice a
problem until later.

Here, when processing named args in a template, walk through them in
alphabetical order. This has no meaningful effect on the rendered SQL,
but provides stability for the test suite.

[1] https://github.com/riverqueue/river/actions/runs/13734150171/job/38415749775?pr=795

### DIFF
--- a/rivershared/sqlctemplate/sqlc_template.go
+++ b/rivershared/sqlctemplate/sqlc_template.go
@@ -177,12 +177,18 @@ func (r *Replacer) RunSafely(ctx context.Context, sql string, args []any) (strin
 
 	if len(container.NamedArgs) > 0 {
 		placeholderNum := len(args)
-		for arg, val := range container.NamedArgs {
+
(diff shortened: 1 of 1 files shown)


bgentry (approved):
Fun fact. In an interview years ago I couldn't recall how to randomly shuffle a slice and wasn't supposed to look things up, so I made it work by relying on a map for a weak pseudo-random sort 😆

brandur:
> Fun fact. In an interview years ago I couldn't recall how to randomly shuffle a slice and wasn't supposed to look things up, so I made it work by relying on a map for a weak pseudo-random sort 😆

LOL!!! A+ Go deep cut rating.

PR #798, opened 2025-03-08
Proof out explicit schema injection without `search_path`

brandur:
Here, follow up #794 to proof out the injection of an explicit schema
into SQL queries so that putting River in a custom schema doesn't
require the use of `search_path`. We use the newly introduced sqlc
template package to prefix table names like so:

    -- name: JobCountByState :one
    SELECT count(*)
    WHERE state = @state;

The main advantage of this approach compared to `search_path` is that
connections no longer require that their `search_path` be set with an
explicit value to work. `search_path` can be set or misset, but because
table names are prefixed with the right schema name already, queries
still go through. This is especially useful in the context of PgBouncer,
where a valid `search_path` setting can't be guaranteed.

Notably, this change doesn't bring in enough new testing to prove that
River really works with explicit schema injection, so for the time being
this setting remains completely internal. What I'd like to try and do is
get some basic infrastructure like this in place first, then prove it
out by starting to move the test suite over to schema-specific tests. By
virtue of doing that we'd be putting the entire load of the River test
suite on the new paths, which should give us high confidence that it's
all working as intended.

### DIFF
--- a/client.go
+++ b/client.go
@@ -282,6 +282,13 @@ type Config struct {
 	// Defaults to DefaultRetryPolicy.
 	RetryPolicy ClientRetryPolicy
 
+	// schema is a non-standard schema where River tables are located. All table
+	// references in database queries will use this value as a prefix.
+	//
+	// Defaults to empty, which causes the client to look for tables using the
+	// setting of Postgres `search_path`.
+	schema string
+
 	// SkipUnknownJobCheck is a flag to control whether the client should skip
 	// checking to see if a registered worker exists in the client's worker bundle
 	// for a job arg prior to insertion.
@@ -376,6 +383,7 @@ func (c *Config) WithDefaults() *Config {
 		ReindexerSchedule:           c.ReindexerSchedule,
 		RescueStuckJobsAfter:        valutil.ValOrDefault(c.RescueStuckJobsAfter, rescueAfter),
 		RetryPolicy:                 retryPolicy,
+		schema:                      c.schema,
 		SkipUnknownJobCheck:         c.SkipUnknownJobCheck,
 		Test:                        c.Test,
 		TestOnly:                    c.TestOnly,
@@ -668,7 +676,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 			// uses listen/notify. Instead, each service polls for changes it's
 			// interested in. e.g. Elector polls to see if leader has expired.
 			if !config.PollOnly {
-				client.notifier = notifier.New(archetype, driver.GetListener())
+				client.notifier = notifier.New(archetype, driver.GetListener(config.schema))
 				client.services = append(client.services, client.notifier)
 			}
 		} else {
@@ -705,6 +713,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 				CancelledJobRetentionPeriod: config.CancelledJobRetentionPeriod,
 				CompletedJobRetentionPeriod: config.CompletedJobRetentionPeriod,
 				DiscardedJobRetentionPeriod: config.DiscardedJobRetentionPeriod,
+				Schema:                      config.schema,
 				Timeout:                     config.JobCleanerTimeout,
 			}, driver.GetExecutor())
 			maintenanceServices = append(maintenanceServices, jobCleaner)
@@ -715,6 +724,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 			jobRescuer := maintenance.NewRescuer(archetype, &maintenance.JobRescuerConfig{
 				ClientRetryPolicy: config.RetryPolicy,
 				RescueAfter:       config.RescueStuckJobsAfter,
+				Schema:            config.schema,
 				WorkUnitFactoryFunc: func(kind string) workunit.WorkUnitFactory {
 					if workerInfo, ok := config.Workers.workersMap[kind]; ok {
 						return workerInfo.workUnitFactory
@@ -730,6 +740,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 			jobScheduler := maintenance.NewJobScheduler(archetype, &maintenance.JobSchedulerConfig{
 				Interval:     config.schedulerInterval,
 				NotifyInsert: client.maybeNotifyInsertForQueues,
+				Schema:       config.schema,
 			}, driver.GetExecutor())
 			maintenanceServices = append(maintenanceServices, jobScheduler)
 			client.testSignals.jobScheduler = &jobScheduler.TestSignals
@@ -750,6 +761,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		{
 			queueCleaner := maintenance.NewQueueCleaner(archetype, &maintenance.QueueCleanerConfig{
 				RetentionPeriod: maintenance.QueueRetentionPeriodDefault,
+				Schema:          config.schema,
 			}, driver.GetExecutor())
 			maintenanceServices = append(maintenanceServices, queueCleaner)
 			client.testSignals.queueCleaner = &queueCleaner.TestSignals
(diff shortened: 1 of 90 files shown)


riverdriver/river_driver_interface.go
```diff
@@ -209,9 +209,19 @@ type Notification struct {
...
 
 type JobCancelParams struct {
+	ID                int64
 	CancelAttemptedAt time.Time
 	ControlTopic      string
-	ID                int64
+}
+
+type JobCountByStateParams struct {
+	Schema string
```
brandur, line 218: So I'd originally had `Schema` as a configuration parameter to the driver, but I realized later that I don't think that's appropriate ... not only is the UX a bit questionable, but it should really be only the client that controls its insert/work schema, and it should be possible for two clients to share a single driver and be using two separate schemas.

The slightly unfortunate part is that this approach would necessitate adding a `Schema string` parameter to every driver params structs where the underlying query targets a River table, but compared to the other possibilities, this might be the best/most explicit way to go.

brandur:
@bgentry This one's going to be quite a bit more work to take all the way to fruition, so I just converted two queries to get your thoughts on the general approach before doing anything more.

riverdriver/river_driver_interface.go
bgentry: Yep, I think that's reasonable. Explicitly passing through the schema makes the most sense and also makes it clear the driver needs to handle running its queries on an arbitrary schema.

bgentry (commented):
I definitely don't _love_ this because it makes the queries uglier + harder to read, and it's quite the (clever) hack to work around sqlc limitations. But, it's also probably the best we can do without changing out that core piece of the stack. I think we should move forward with it.

Maybe we also want to make it so we can run a full test suite against a database that only has our migrations run in an alternate schema? Even if it mainly runs in CI it would be an effective way to ensure this all works & keeps working.

brandur:
> I definitely don't _love_ this because it makes the queries uglier + harder to read, and it's quite the (clever) hack to work around sqlc limitations. But, it's also probably the best we can do without changing out that core piece of the stack. I think we should move forward with it.

Yeah, I'm going to see if I can pass this along to Kyle. Low probability of anything happening, but maybe it'd inspire him as to what a potentially native solution for this would look like. Ideally, this lives in the sqlc layer.

> Maybe we also want to make it so we can run a full test suite against a database that only has our migrations run in an alternate schema? Even if it mainly runs in CI it would be an effective way to ensure this all works & keeps working.

Yeah, we'll need something like that I think. Another one I'm considering is if we could move our test databases over to test schemas, we'd get full 100% vetting that things are working because by necessity, the test suite depends on it.

I worked on this another couple hours last night though, and even just getting the basic plumbing in is thousands of LOCs changed, and I haven't even started on testing or rehabilitating the listen/notify system to be schema-aware. I'm kind of wondering if it might be smart to bring in some of the high churn changes early, but keep the new `Schema` config unexported for now (so `schema`), then follow on by doing some of the other major work streams from there, only releasing it once it's fully ready/tested.

bgentry:
> I'm kind of wondering if it might be smart to bring in some of the high churn changes early, but keep the new `Schema` config unexported for now (so `schema`), then follow on by doing some of the other major work streams from there, only releasing it once it's fully ready/tested.

Makes sense. It's best if we can avoid exposing it until we're confident it's ready-to-go. And given the low chance of upstream sqlc changes in a short timeframe, IMO it's also fine for us to move forward with the hacky DIY approach. We can always switch over if an official easier option becomes available.

riverdriver/river_driver_interface.go
```diff
@@ -100,76 +100,75 @@ type Executor interface {
 
 	// ColumnExists checks whether a column for a particular table exists for
 	// the schema in the current search schema.
-	ColumnExists(ctx context.Context, tableName, columnName string) (bool, error)
+	ColumnExists(ctx context.Context, params *ColumnExistsParams) (bool, error)
```
brandur, line 103: Since everything was getting a new `schema` argument, I ended up just changing everything into a parameter struct so (1) we don't have some awkward functions that need to take three strings like this one, and (2) just so it's all predictable. Everything just takes a param struct.

This ended up requiring a lot of refactoring so not sure it was worth it in the end, but oh well.

brandur:
@bgentry Alright, so I ended up going through and converting everything over to take a schema template param.

As mentioned above, it's unlikely this fully works given we'd have to add significant new test coverage to root out all the edges, so all the functionality is kept internal-only for the time being until the feature can be fully finished. I'm thinking next we could try to switch over to schema-based testing, and use that strategy to fully vet what's happening here as schema-based testing would imply the entire weight of the test suite on the schema injection logic here.

Thoughts?

client.go
```diff
@@ -282,6 +282,13 @@ type Config struct {
 	// Defaults to DefaultRetryPolicy.
 	RetryPolicy ClientRetryPolicy
 
+	// schema is a non-standard schema where River tables are located. All table
+	// references in database queries will use this value as a prefix.
+	//
+	// Defaults to empty, which causes the client to look for tables using the
+	// setting of Postgres `search_path`.
+	schema string
```
bgentry, line 290: I wonder, would we be better off defaulting to this explicitly setting `public` vs letting it default to conn string search path? Probably not, just trying to think through the full cycle of existing users migrating to this setup once we have full schema support.

rivershared/riverpilot/pilot.go
```diff
@@ -26,7 +26,7 @@ type Pilot interface {
 	JobInsertMany(
 		ctx context.Context,
 		tx riverdriver.ExecutorTx,
-		params []*riverdriver.JobInsertFastParams,
+		params *riverdriver.JobInsertFastManyParams,
```
bgentry, line 29: I'm sure you've already considered it but this will definitely require corresponding changes in Pro. Then again so will a lot of this PR :)

internal/jobexecutor/job_executor_test.go
```diff
@@ -654,7 +718,10 @@ func TestJobExecutor_Execute(t *testing.T) {
 		executor.Execute(ctx)
 		riversharedtest.WaitOrTimeout(t, bundle.updateCh)
 
-		job, err := bundle.exec.JobGetByID(ctx, bundle.jobRow.ID)
+		job, err := bundle.exec.JobGetByID(ctx, &riverdriver.JobGetByIDParams{
+			ID:     bundle.jobRow.ID,
+			Schema: "",
+		})
```
bgentry, line 724: These all get quite a bit more verbose, unfortunately 😕

internal/leadership/elector_test.go
```diff
@@ -80,9 +80,10 @@ func TestElector_WithNotifier(t *testing.T) {
 				archetype = riversharedtest.BaseServiceArchetype(t)
 				dbPool    = riverinternaltest.TestDB(ctx, t)
 				driver    = riverpgxv5.New(dbPool)
+				schema    = "" // try to make tests schema-based rather than database-based in the future
```
bgentry, line 83: 💯 I have thoughts on this, might be worth talking about possible approaches together before diving in

bgentry (approved):
Herculean effort 💪 I wish this wasn't all necessary but it seems it's either this brute force method or ditching sqlc altogether. Looking forward to exercising this further with schema-based tests instead of DB-based 🚀

client.go
brandur: I thinkk at this point we'll probably have to leave empty as default for backwards compatibility if nothing else.

Starting from scratch, yeah fair question. If we gave it a default though, we'd have to also provide a way to zero it for anyone who'd prefer to use `search_path` by choice.

rivershared/riverpilot/pilot.go
brandur: Yeah, I think I finished most of those up this evening thankfully.  I'll send them over today or tomorrow.

internal/jobexecutor/job_executor_test.go
brandur: Yeah :/ I think we can clean a lot of this up again with a few more helpers for test functions that end up generating a lot of objects.

internal/leadership/elector_test.go
brandur: Soo I ended up writing a semi-functional implementation yesterday, but yeah, happy to cover more. I think this is going to be a good path though — there are a ton of advantages compared to test DBs (e.g. can create as many schemas as you need so nothing blocks waiting on DBs, full parallel testing, still get efficient reuse where possible, no need to run testdbman).

brandur:
Thanks! Okay this is still a bit rough (schemas _definitely_ do not work yet), but I have another branch based on this one where I've started to get large swathes of the test suite working a schema-based design. There's still a bunch of debugging that needs to be done, but I think I can get something up by EOW.

The schema stuff will be a bit of a mess in `master` and not usable, but as far as I know, not harmful.

PR #799, opened 2025-03-08
Safer template usage with more aggressive errors on likely problems

brandur:
Follows up the addition of `sqlctemplate` in #794. I noticed while
adding functionality in with templates that it was quite easy to (1) add
a `/* TEMPLATE` tag to `river_job.sql`, (2) put in context parameters to
a driver, but then (3) forget to run `make generate`. The context
parameters are injected, but `sqlctemplate` no ops with a fast short
circuit because there's no `/* TEMPLATE` tag present in the generated
Go code that the driver is executing. This leads to confusion.

Here, add a few more error conditions:

* If a context container is present without any `/* TEMPLATE` tags,
  error.

* If any `/* TEMPLATE` tags are present without a context container,
  error.

This makes dumb bugs easier to catch because we get an explicit error
instead of them failing silently. Tests are updated to check for them.

### DIFF
--- a/rivershared/sqlctemplate/sqlc_template.go
+++ b/rivershared/sqlctemplate/sqlc_template.go
@@ -110,10 +110,20 @@ func (r *Replacer) Run(ctx context.Context, sql string, args []any) (string, []a
 // RunSafely is the same as Run, but returns an error in case of missing or
 // extra templates.
 func (r *Replacer) RunSafely(ctx context.Context, sql string, args []any) (string, []any, error) {
-	// If nothing present in context, short circuit quickly.
-	container, containerOK := ctx.Value(contextKey{}).(*contextContainer)
-	if !containerOK {
+	var (
+		container, containerOK = ctx.Value(contextKey{}).(*contextContainer)
+		sqlContainsTemplate    = strings.Contains(sql, "/* TEMPLATE")
+	)
+	switch {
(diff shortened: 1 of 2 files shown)


rivershared/sqlctemplate/sqlc_template.go
```diff
@@ -110,9 +110,19 @@ func (r *Replacer) Run(ctx context.Context, sql string, args []any) (string, []a
...
+	switch {
+	case containerOK && !sqlContainsTemplate:
+		return "", nil, errors.New("sqlctemplate found context container but SQL contains no templates; bug?")
+
+	case !containerOK && sqlContainsTemplate:
+		return "", nil, errors.New("sqlctemplate found template(s) in SQL, but no context container; bug?")
+
+	case !containerOK && !sqlContainsTemplate:
+		// Neither context container or template in SQL; short circuit fast because there's no work to do.
 		return sql, args, nil
```
bgentry, line 126: Switch statements are executed top-to-bottom, so maybe you want this fast path short circuit to go first?

bgentry (approved):

rivershared/sqlctemplate/sqlc_template.go
brandur: Done. Thx.

PR #800, opened 2025-03-09
More consistent naming for `sqlctemplate` + docs

brandur:
A few more small follow ups for #794. I started renaming things late in
the process while refactoring it, and a few of the renames didn't end up
getting bubbled all the way through.

Try to maintain roughly this vernacular:

* **Template:** A template itself like `/* TEMPLATE: ...` along with its
  assigned template name.

* **Replacement:** A replacement value for a template, possibly
  including some metadata like `Stable`.

Get rid of the use of `tmpl` in favor of `replacement` for locals.


bgentry (approved):

brandur:
Thx.

PR #805, opened 2025-03-12
Hooks + middleware: internal sentinel functions

brandur:
Here, I wanted to make sure we had a record of the general idea to make
the `rivertype.hook` and `rivertype.Middleware` "sentinel" functions
package internal so that it wouldn't even be possible to implement them
otherwise.

This does require that `HookDefaults` and `MiddlewareDefaults` get moved
from `river` to `rivertype`, along with function implementations like
`HookWorkBeginFunc`.

Roughly, advantages:

* Keeps more implementation internal/out of Godoc.
* Forces use of one of the defaults structs, making them impossible to
  accidentally regret.
* Means that hook/middleware implementations generally only need to
  import one package: `rivetype`. Previously they'd also need `river`
  because that's where defaults lived. This looks a bit nicer.

Disadvantages:

* Having defaults and funcs in `rivertype` inconsistent compared to
  workers.
* Makes certain types of `Hook` and `Middleware` interface
  implementations impossible for users, even for those who know what
  they're doing. For example, users could not write their own version of
  the `HookWorkBeginFunc` type because it's not a struct, and cannot
  embed `HookDefaults`.


internal/jobexecutor/job_executor_test.go
```diff
@@ -736,11 +736,10 @@ func TestJobExecutor_Execute(t *testing.T) {
 }
 
 type testMiddleware struct {
+	rivertype.MiddlewareDefaults
 	work func(ctx context.Context, job *rivertype.JobRow, next func(context.Context) error) error
 }
 
-func (m *testMiddleware) IsMiddleware() bool { return true }
```
brandur, line 742: This could also be considered a minor benefit. Previously, adding a middleware to tests for an internal package like `jobexecutor` here forced you to provide your implementation because `MiddlewareDefaults` was in `river`, and you can't import that lest you introduce a cycle. Putting `MiddlewareDefaults` in `rivertype` means it's all good.

brandur:
@bgentry Mostly just putting this up here as reference. All else being equal I might lean slightly towards doing it this way, but TBH I can make strong arguments in either direction, so not sure about it.

bgentry (approved):
I'm honestly right on the fence at like 50/50 with no strong preference, so I will you do the favor of saying this is entirely your call 😆 Implementation here looks good if you want to move forward with it.

brandur:
K cool, thank you! I thought about this overnight and I think I'm slightly leaning toward what we have now (i.e. *not* this variant). I'll leave this open until our next release, but won't merge it.

brandur:
Decided not to go this way. Closing.

PR #807, opened 2025-03-16
Prepare release 0.19.0

brandur:
It's been a while since the last release, and it'd be nice to get a new
release out including some middleware changes and the addition of hooks
so that we can do a fast follow with additions like the upcoming
OpenTelemetry middleware.

[skip ci]

### DIFF
--- a/cmd/river/go.mod
+++ b/cmd/river/go.mod
@@ -1,16 +1,17 @@
 module github.com/riverqueue/river/cmd/river
 
 go 1.22.0
+
 toolchain go1.23.5
 
 require (
 	github.com/jackc/pgx/v5 v5.7.2
 	github.com/lmittmann/tint v1.0.7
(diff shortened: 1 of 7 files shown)


cmd/river/go.mod
```diff
@@ -1,16 +1,17 @@
 module github.com/riverqueue/river/cmd/river
 
 go 1.22.0
```
bgentry, line 3: should we think about bumping this next release, or not until really needed?

bgentry (approved):

cmd/river/go.mod
brandur: Ah yeah, good question. I try not to be too aggressive about it and just bump when we want it or someone asks for it. That said, Go 1.24's been out a while now, so I think it could be plausible be done at any time.

brandur:
Thanks!

PR #808, opened 2025-03-17
Let `update-mod-version` take a `PACKAGE_PREFIX` env var

brandur:
Here, have `update-mod-version` take a `PACKAGE_PREFIX` env var that
dictates which prefix it'll operate on, defaulting the previous behavior
of `github.com/riverqueue/river`.

The idea here is that we should be able to use reuse this command from
other River repositories like `rivercontrib` to update internal cross
dependencies.

### DIFF
--- a/rivershared/cmd/update-mod-version/main.go
+++ b/rivershared/cmd/update-mod-version/main.go
@@ -26,6 +26,13 @@ func main() {
 }
 
 func run() error {
+	// Allows secondary River repositories to make use of this command by
+	// specifying their own prefix.
+	packagePrefix := os.Getenv("PACKAGE_PREFIX")
+	if packagePrefix == "" {
+		return errors.New("expected to find PACKAGE_PREFIX in env")
+	}
+
 	version := os.Getenv("VERSION")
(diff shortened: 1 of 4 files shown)


rivershared/cmd/update-mod-version/main.go
```diff
@@ -26,6 +27,10 @@ func main() {
 }
 
 func run() error {
+	// Allows secondary River repositories to make use of this command by
+	// specifying their own prefix.
+	packagePrefix := cmp.Or(os.Getenv("PACKAGE_PREFIX"), "github.com/riverqueue/river")
```
bgentry, line 32: Could just make this a required input and skip having a default value? Your call.

bgentry (approved):

rivershared/cmd/update-mod-version/main.go
brandur: Yeah, good point. Can't hurt to be explicit about this. Updated the command and added `PATH_PREFIX` in this project's `Makefile`.

brandur:
Thanks!

PR #809, opened 2025-03-18
(Failed) iterator experiment instead of multiple for loops

brandur:
Here, experiment with upgrading to Go 1.23 so we can replace a couple
hooks-related TODOs where we have two separate for loops with a single
invocation that uses an iterator to iterate two slices without an
additional slice invocation.

I'm putting this up for the sake of interest, but it unfortunately turns
out to be quite a failed experiment. The iterators work, but using them
is slower than both the previous two `for` loops, and also slower than
just allocating a new slice with the contents of the previous two and
iterating that.

At larger slice sizes, the iterator becomes marginally faster than
allocating a new slice, but only _marginally_. Using two for loops
continues to be bit a bit faster (perhaps unsurprisingly).

Anyway, opened this for general interest, but am going to close it, and
probably just replace the two for loops with a slice allocation, since
it looks nicer and it's the fastest option for small slices sizes (and
we generally expect hooks to be relatively small slices).

    $ go test ./rivershared/util/sliceutil -bench=.
    goos: darwin
    goarch: arm64
    pkg: github.com/riverqueue/river/rivershared/util/sliceutil
    cpu: Apple M4
    BenchmarkIterateCombined/IterateCombined-10              1481770               822.0 ns/op
    BenchmarkIterateCombined/SliceAllocation-10              1617640               754.1 ns/op
    BenchmarkIterateCombined/TwoForLoops-10                  1560508               775.7 ns/op
    BenchmarkIterateCombined/IterateCombinedLong-10          1393824               863.5 ns/op
    BenchmarkIterateCombined/SliceAllocationLong-10          1399813               866.7 ns/op
    BenchmarkIterateCombined/TwoForLoopsLong-10              1445377               835.2 ns/op
    PASS
    ok      github.com/riverqueue/river/rivershared/util/sliceutil  12.272s

### DIFF
--- a/client.go
+++ b/client.go
@@ -1597,17 +1597,10 @@ func (c *Client[TTx]) insertManyShared(
 ) ([]*rivertype.JobInsertResult, error) {
 	doInner := func(ctx context.Context) ([]*rivertype.JobInsertResult, error) {
 		for _, params := range insertParams {
-			// TODO(brandur): This range clause and the one below it are
-			// identical, and it'd be nice to merge them together, but in such a
-			// way that doesn't require array allocation. I think we can do this
-			// using iterators after we drop support for Go 1.22.
-			for _, hook := range c.hookLookupGlobal.ByHookKind(hooklookup.HookKindInsertBegin) {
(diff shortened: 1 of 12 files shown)


brandur:
Closing in favor of #810.

PR #810, opened 2025-03-18
Iterate over combined hook slices instead of iterating using two loops

brandur:
Clean up some duplicated code in which we were previously iterating over
two sets of hooks and running the same code on both collections, and I'd
left a TODO to investigate the use of sequences instead to avoid a slice
allocation.

I ran the sequences experiment over in #809, and although I'd still
defend it as a worthwhile prototype, it turns out that using sequences
only becomes more efficient than allocating an extra slice once we get
to fairly large slice sizes, and even there the benefit is quite
marginal.

The benchmarks also show that at smaller slice sizes, it's actually a
tiny bit faster allocating a slice to iterate over once instead of using
two for loops. This is convenient because that sure looks a lot nicer
code-wise anyway, so here we switch over to that approach and remove the
TODOs.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1597,17 +1597,10 @@ func (c *Client[TTx]) insertManyShared(
 ) ([]*rivertype.JobInsertResult, error) {
 	doInner := func(ctx context.Context) ([]*rivertype.JobInsertResult, error) {
 		for _, params := range insertParams {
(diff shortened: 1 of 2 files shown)


bgentry (approved):
Makes sense, thanks for testing both approaches.

brandur:
NP. Thanks!

PR #815, opened 2025-03-23
Deprecate per-operation middleware operation defaults

brandur:
As per #804, middleware has been restructed to look more like the design
of hooks, and I was finding that as I restructuring documentation as
part of [1] it was hard to recommend the use of per-operation middleware
defaults.

Here, deprecate the per-operation defaults in favor of just the single
`river.MiddlewareDefaults` struct. This should also be somewhat more
beneficial in case more middleware operations need to be added in the
future because we'll just have fewer total new types emerge out of it.

[1] https://github.com/riverqueue/homepage/pull/198

### DIFF
--- a/middleware_defaults.go
+++ b/middleware_defaults.go
@@ -6,6 +6,9 @@ import (
 	"github.com/riverqueue/river/rivertype"
 )
 
+// MiddlewareDefaults should be embedded on any middleware implementation. It
+// helps identify a struct as middleware, and guarantees forward compatibility in
+// case additions are necessary to the rivertype.Middleware interface.
 type MiddlewareDefaults struct{}
 
 func (d *MiddlewareDefaults) IsMiddleware() bool { return true }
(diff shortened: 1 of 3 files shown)


middleware_defaults.go
```diff
@@ -6,6 +6,9 @@ import (
 	"github.com/riverqueue/river/rivertype"
 )
 
+// MiddlewareDefaults should be embedded on any middleware implementation. It
+// helps identify a struct as middleware, and guarantee forward compatibility in
```
bgentry, line 10: Might be slightly more readable as "guarantees":

```suggestion
// helps identify a struct as middleware, and guarantees forward compatibility in
```

bgentry (approved):

middleware_defaults.go
brandur: Good catch. Thanks.

PR #818, opened 2025-03-25
Upgrade golangci-lint to v2

brandur:
For better or rwose, golangci-lint has released a v2 [1] that's
incompatible with v1. It'll only be a matter of time until we're forced
to upgrade to it, so I figure we may as well do it early.

There's a bunch of new obnoxious linters that I've disabled. Most of
these complain about "cyclomatic complexity" or "cognitive complexity"
or functions being too long (they all seem to want everything broken
down into functions of ten lines).

Aside from that, `testifylint` seems to have gotten even more effective
at finding misuses of testify, so it fixed a bunch more arguable misuses
of various assertions, which seems good for normalizing code.

The structure of the YAML file changes fairly significantly, but it's
all basically the same thing with various aspects reorganized. I made
the new one using the built-in `golangci-lint migrate` and copied
comments over manually.

If we can get this in, the next step will be to distribute it out to
other River projects. Most of them are using copies of this repo's
configuration file, so it should be pretty easy to do.

[1] https://ldez.github.io/blog/2025/03/23/golangci-lint-v2/
[2] https://github.com/golangci/golangci-lint/pull/5606


.golangci.yaml
```diff
@@ -21,82 +12,108 @@ linters:
...
+    - cyclop # screams into the void at "cyclomatic complexity"
+    - funlen # screams when functions are more than 60 lines long; what are we even doing here guys
     - interfacebloat # we do in fact want >10 methods on the Adapter interface or wherever we see fit.
+    - gocognit # yells that "cognitive complexity" is too high; why
+    - gocyclo # ANOTHER "cyclomatic complexity" checker (see also "cyclop" and "gocyclo")
     - godox # bans TODO statements; total non-starter at the moment
     - err113 # wants all errors to be defined as variables at the package level; quite obnoxious
+    - maintidx # ANOTHER ANOTHER "cyclomatic complexity" lint (see also "cyclop" and "gocyclo")
     - mnd # detects "magic numbers", which it defines as any number; annoying
+    - nestif # yells when if blocks are nested; what planet do these people come from?
```
bgentry, line 24: 🤣

.golangci.yaml
```diff
@@ -21,82 +12,108 @@ linters:
     - testpackage # requires tests in test packages like `river_test`
 
     # disabled because they're annoying/bad
+    - cyclop # screams into the void at "cyclomatic complexity"
+    - funlen # screams when functions are more than 60 lines long; what are we even doing here guys
```
bgentry, line 16: These are popular in Ruby and Elixir. It makes a bit more sense in Ruby with DSLs and the low cost of extracting tons of little methods, but less so in Go IMO.

.golangci.yaml
```diff
@@ -21,82 +12,108 @@ linters:
...
-            pkg: "github.com/riverqueue/river/internal/riverinternaltest"
+  settings:
+    depguard:
+      rules:
+        all:
+          files:
+            - $all
+          deny:
+            - pkg: github.com/xtgo/uuid
+              desc: Use `github.com/google/uuid` package for UUIDs instead.
```
bgentry, line 38: intentional non-alphabetical resorts of attr ordering in here? There's a whole bunch of them.

internal/dblist/db_list.go
```diff
@@ -99,10 +99,12 @@ func JobList(ctx context.Context, exec riverdriver.Executor, params *JobListPara
...
 	for i, orderBy := range params.OrderBy {
 		orderByBuilder.WriteString(orderBy.Expr)
-		if orderBy.Order == SortOrderAsc {
+		switch orderBy.Order {
+		case SortOrderAsc:
 			orderByBuilder.WriteString(" ASC")
-		} else if orderBy.Order == SortOrderDesc {
+		case SortOrderDesc:
 			orderByBuilder.WriteString(" DESC")
+		case SortOrderUnspecified:
```
bgentry, line 107: Maybe this should actually be an error 🤔

bgentry (approved):
I'm a little nervous about shipping this in Pro prior to landing the next big PR there. Will try to get that finished up ASAP so it's a non-issue.

.golangci.yaml
brandur: Yeah exactly. IMO it seems fine to keep function lengths generally short, but requiring it of every single function is just too much. There are many places where a slightly longer function can be *more* clear than a shorter one! Dogma is the enemy quality.

.golangci.yaml
brandur: Ah yeah I didn't even notice.

I guess it doesn't matter as much for these sorts of two-item tuples, but just so we have a system, I put them all back the way they were including the Forbidigo rules below.

internal/dblist/db_list.go
brandur: Ah yeah, I wasn't sure if this was supposed to be possible or not.

I just replaced it with an error instead. Seems to pass the tests so hopefully it's okay ...

PR #822, opened 2025-03-26
Expand pilot APIs for state tracking, queue metadata changes

bgentry:
This adds expanded integration points at the Pilot level for enabling producer job state tracking. It also adds producer-level monitoring of queue metadata (propagated to the pilot upon change) and a `QueueUpdate` API on the client.

I'm debating a bit on how to document `QueueUpdateMetadata` because it's an advanced API and potentially very footgun-prone. Also how to document it in the changelog for the same reason. The UI will call this for some upcoming features. Thoughts?

### DIFF
--- a/client.go
+++ b/client.go
@@ -873,8 +873,8 @@ func (c *Client[TTx]) Start(ctx context.Context) error {
 
 		for _, producer := range c.producersByQueueName {
 			if err := producer.StartWorkContext(fetchCtx, workCtx); err != nil {
-				startstop.StopAllParallel(producersAsServices()...)
 				workCancel(err)
+				startstop.StopAllParallel(producersAsServices()...)
 				stopServicesOnError()
 				return err
 			}
@@ -1846,23 +1846,24 @@ func (c *Client[TTx]) validateJobArgs(args JobArgs) error {
 }
 
 func (c *Client[TTx]) addProducer(queueName string, queueConfig QueueConfig) *producer {
-	producer := newProducer(&c.baseService.Archetype, c.driver.GetExecutor(), &producerConfig{
-		ClientID:               c.config.ID,
-		Completer:              c.completer,
-		ErrorHandler:           c.config.ErrorHandler,
-		FetchCooldown:          c.config.FetchCooldown,
-		FetchPollInterval:      c.config.FetchPollInterval,
-		HookLookupByJob:        c.hookLookupByJob,
-		HookLookupGlobal:       c.hookLookupGlobal,
-		JobTimeout:             c.config.JobTimeout,
-		MaxWorkers:             queueConfig.MaxWorkers,
-		MiddlewareLookupGlobal: c.middlewareLookupGlobal,
-		Notifier:               c.notifier,
-		Queue:                  queueName,
-		QueueEventCallback:     c.subscriptionManager.distributeQueueEvent,
-		RetryPolicy:            c.config.RetryPolicy,
-		SchedulerInterval:      c.config.schedulerInterval,
-		Workers:                c.config.Workers,
+	producer := newProducer(&c.baseService.Archetype, c.driver.GetExecutor(), c.pilot, &producerConfig{
+		ClientID:                     c.config.ID,
+		Completer:                    c.completer,
+		ErrorHandler:                 c.config.ErrorHandler,
+		FetchCooldown:                c.config.FetchCooldown,
+		FetchPollInterval:            c.config.FetchPollInterval,
+		HookLookupByJob:              c.hookLookupByJob,
+		HookLookupGlobal:             c.hookLookupGlobal,
+		JobTimeout:                   c.config.JobTimeout,
+		MaxWorkers:                   queueConfig.MaxWorkers,
+		MiddlewareLookupGlobal:       c.middlewareLookupGlobal,
+		Notifier:                     c.notifier,
+		Queue:                        queueName,
+		QueueEventCallback:           c.subscriptionManager.distributeQueueEvent,
+		RetryPolicy:                  c.config.RetryPolicy,
+		SchedulerInterval:            c.config.schedulerInterval,
+		StaleProducerRetentionPeriod: 5 * time.Minute,
+		Workers:                      c.config.Workers,
 	})
 	c.producersByQueueName[queueName] = producer
 	return producer
@@ -2175,6 +2176,9 @@ type QueueBundle struct {
 // Add adds a new queue to the client. If the client is already started, a
 // producer for the queue is started. Context is inherited from the one given to
 // Client.Start.
+//
+// TODO: there is no way for this to work at runtime using a separate pro queue
+// config, unless we put pro configs like concurrency within QueueConfig.
 func (b *QueueBundle) Add(queueName string, queueConfig QueueConfig) error {
 	if !b.clientWillExecuteJobs {
 		return errors.New("client is not configured to execute jobs, cannot add queue")
--- a/client.go
+++ b/client.go
@@ -863,7 +863,7 @@ func (c *Client[TTx]) Start(ctx context.Context) error {
 		// We use separate contexts for fetching and working to allow for a graceful
 		// stop. Both inherit from the provided context, so if it's cancelled, a
 		// more aggressive stop will be initiated.
-		workCtx, workCancel := context.WithCancelCause(withClient[TTx](ctx, c))
+		workCtx, workCancel := context.WithCancelCause(withClient(ctx, c))
 
 		if err := startstop.StartAll(fetchCtx, c.services...); err != nil {
 			workCancel(err)
--- a/client.go
+++ b/client.go
@@ -1809,7 +1809,7 @@ func (c *Client[TTx]) notifyQueuePauseOrResume(ctx context.Context, tx riverdriv
 		slog.String("opts", fmt.Sprintf("%+v", opts)),
 	)
 
-	payload, err := json.Marshal(jobControlPayload{Action: action, Queue: queue})
+	payload, err := json.Marshal(controlEventPayload{Action: action, Queue: queue})
 	if err != nil {
 		return err
 	}
--- a/client.go
+++ b/client.go
@@ -2155,6 +2155,58 @@ func (c *Client[TTx]) QueueResumeTx(ctx context.Context, tx TTx, name string, op
 	return nil
 }
 
+// QueueUpdateParams are the parameters for a QueueUpdate operation.
+type QueueUpdateParams struct {
+	// Metadata is the new metadata for the queue. If nil or empty, the metadata
+	// will not be changed.
+	Metadata []byte
+}
+
+// QueueUpdate updates a queue's settings in the database. These settings
+// override the settings in the client (if applied).
+func (c *Client[TTx]) QueueUpdate(ctx context.Context, name string, params *QueueUpdateParams) (*rivertype.Queue, error) {
+	tx, err := c.driver.GetExecutor().Begin(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer tx.Rollback(ctx)
+
+	updateMetadata := len(params.Metadata) > 0
+
+	queue, err := tx.QueueUpdate(ctx, &riverdriver.QueueUpdateParams{
+		Metadata:         params.Metadata,
+		MetadataDoUpdate: updateMetadata,
+		Name:             name,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if updateMetadata {
+		payload, err := json.Marshal(controlEventPayload{
+			Action:   controlActionMetadataChanged,
+			Metadata: params.Metadata,
+			Queue:    queue.Name,
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		if err := tx.NotifyMany(ctx, &riverdriver.NotifyManyParams{
+			Topic:   string(notifier.NotificationTopicControl),
+			Payload: []string{string(payload)},
+		}); err != nil {
+			return nil, err
+		}
+	}
+
+	if err := tx.Commit(ctx); err != nil {
+		return nil, err
+	}
+
+	return queue, nil
+}
+
 // QueueBundle is a bundle for adding additional queues. It's made accessible
 // through Client.Queues.
 type QueueBundle struct {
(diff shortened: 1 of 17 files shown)


client.go
```diff
@@ -873,6 +873,7 @@ func (c *Client[TTx]) Start(ctx context.Context) error {
 
 		for _, producer := range c.producersByQueueName {
 			if err := producer.StartWorkContext(fetchCtx, workCtx); err != nil {
+				workCancel(err)
```
bgentry, line 876: Hmm, spurious diff perhaps, although upon reviewing it I'm not sure whether the "right" behavior is to cancel the work context before or after stopping all producers. I'm thinking maybe before because something has obviously gone wrong on startup and we should bail out sooner rather than letting long-running jobs keep going indefinitely?

producer.go
```diff
@@ -343,15 +399,17 @@ func (p *producer) StartWorkContext(fetchCtx, workCtx context.Context) error {
...
+	controlActionMetadataChanged controlAction = "metadata_changed"
+	controlActionPause           controlAction = "pause"
+	controlActionResume          controlAction = "resume"
 )
 
-type jobControlPayload struct {
-	Action controlAction `json:"action"`
-	JobID  int64         `json:"job_id"`
-	Queue  string        `json:"queue"`
+type controlEventPayload struct {
```
bgentry, line 408: Renamed to generalize since it's not purely for job control events.

internal/util/jsonutil/json_util.go
bgentry, line 1: Got any better ideas for this? The intention is to be able to tell when the metadata/config has changed so that we can give the pilot an updated config. In notify mode that's easy because the message is published as part of the update. However for poll-only mode, the best I could think of is to diff the JSON.

This is obviously not particularly efficient, but it shouldn't matter much—we're talking about decoding two small JSON payloads once every few seconds.

producer.go
```diff
@@ -258,15 +288,29 @@ func (p *producer) StartWorkContext(fetchCtx, workCtx context.Context) error {
...
 
-	initiallyPaused := queue != nil && (queue.PausedAt != nil)
-	p.paused = initiallyPaused
+	var initialQueue rivertype.Queue
+	if fetchedQueue != nil {
+		initialQueue = *fetchedQueue
+	}
+	p.paused = fetchedQueue.PausedAt != nil
+
+	// TODO: how does this handle being reinitialized after stop/start?
```
bgentry, line 304: Shoot, need to resolve this one.

producer.go
```diff
@@ -258,15 +288,29 @@ func (p *producer) StartWorkContext(fetchCtx, workCtx context.Context) error {
...
 		return err
 	}
 
-	initiallyPaused := queue != nil && (queue.PausedAt != nil)
-	p.paused = initiallyPaused
+	var initialQueue rivertype.Queue
+	if fetchedQueue != nil {
+		initialQueue = *fetchedQueue
+	}
+	p.paused = fetchedQueue.PausedAt != nil
```
bgentry, line 302: I think this was a little clearer previously, going to refactor it a little

client.go
```diff
@@ -2173,6 +2214,9 @@ type QueueBundle struct {
 // Add adds a new queue to the client. If the client is already started, a
 // producer for the queue is started. Context is inherited from the one given to
 // Client.Start.
+//
+// TODO: there is no way for this to work at runtime using a separate pro queue
+// config, unless we put pro configs like concurrency within QueueConfig.
```
bgentry, line 2219: @brandur this is re: the comment on the other PR

producer.go
```diff
@@ -527,7 +596,39 @@ func (p *producer) executorShutdownLoop() {
...
-		p.removeActiveJob(result.ID)
+		p.removeActiveJob(result)
+	}
+}
+
+func (p *producer) finalizeShutdown(baseCtx context.Context) {
+	p.Logger.Debug(p.Name + ": Finalizing shutdown")
+
+	for attempt := 1; attempt <= 5; attempt++ {
+		timeout := min(time.Second*time.Duration(attempt*attempt), 10*time.Second)
```
brandur, line 607: Can you add a comment to try and rationalize this this timeout logic? There's no more standard/less bespoke way this could be accomplished?

brandur (commented):
Cool man. Left a few comments here, but looking pretty close.

producer.go
```diff
@@ -208,9 +232,11 @@ func newProducer(archetype *baseservice.Archetype, exec riverdriver.Executor, co
 		config:         config.mustValidate(),
 		exec:           exec,
 		errorHandler:   errorHandler,
+		id:             uuid.New(),
```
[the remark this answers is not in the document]
bgentry: Not quite following what you’re suggesting re: basing producer IDs off client IDs. The producer schema specifically uses a UUID primary key—what would you change that to? Client IDs can be essentially whatever string a user chooses, right?

I went back and forth on including this dependency and just doing our own `byte[16]`, but when I looked at the code I realized it was a small self contained dep and stopped caring as much.

On the main river side yeah it’s not doing much besides generating and passing the values down, though pro does do a bit more with the type. It’s super important that these are kept unique so any alternate setup needs to still reliably ensure that.

client.go
```diff
@@ -2154,6 +2156,45 @@ func (c *Client[TTx]) QueueResumeTx(ctx context.Context, tx TTx, name string, op
 	return nil
 }
 
+func (c *Client[TTx]) QueueUpdateMetadata(ctx context.Context, name string, metadata []byte) (*rivertype.Queue, error) {
```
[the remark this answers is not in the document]
bgentry: Went back and forth on the right design for this because bools like `MetadataDoUpdate` feel unconventional in a public API. In this case, I realized we can't actually set the metadata to a nil or empty value anyway (that's not valid) so we can use its presence to determine whether or not to update the value. Other fields in the future can use pointers.

I kept `name` as a standalone function arg because it is required and is the way to identify the row being updated.

client.go
bgentry: Yeah, I think so. `QueueBundle` is a struct (not an interface) and it has a `func (b *QueueBundle) Add(queueName string, queueConfig QueueConfig) error` on it where `QueueConfig` is also a struct.

If we overrode it, it would be a minor breaking change for current Pro users to need to provide a `riverpro.QueueConfig` instead of a `river.QueueConfig`. The larger issue is that this becomes a friction point for anybody who is using this `.Add()` method and who later adopts River Pro, because all of them will need to to that conversion to onboard/offboard from Pro.

Another option is to expose a `ProQueues() riverpro.QueueBundle` to the `riverpro.Client` alongside this existing method on the `river.Client`. This way, existing code adding non-pro queues continues to work after switching, and it's only when adding queues leveraging pro-specific features that you'd need to change your code to use that route. When offboarding to a non-Pro client, that updated code would generate compile errors to help figure out which Pro features are still in use.

The latter option is probably more seamless. Thoughts?

internal/util/jsonutil/json_util.go
```diff
@@ -0,0 +1,20 @@
...
+func Equal(a, b []byte) bool {
+	var unmarshaledA, unmarshaledB map[string]any
+	if err := json.Unmarshal(a, &unmarshaledA); err != nil {
+		return false
+	}
+	if err := json.Unmarshal(b, &unmarshaledB); err != nil {
+		return false
+	}
+	return reflect.DeepEqual(unmarshaledA, unmarshaledB)
+}
```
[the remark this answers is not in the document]
bgentry: I think you're right that `encoding/json` will marshal in a consistent, repeatable, sorted fashion. Granted it's still using `reflect` internally so it's a superficial improvement, but I'll make that change.

producer.go
bgentry: @brandur hmm yeah it's been like 4mo since I originally wrote this so my memory is hazy on the rationale. Essentially we want to try to make an update at shutdown to mark the producer as no longer running (paused). We don't want to potentially hang forever and block shutdown, and we don't want to keep trying if we hit an unrecoverable error that indicates we can never succeed (such as if the pool is closed or if the context is canceled instead of timed out). As currently written, it sleeps 1 second, then 4 seconds, then 9, then 10 between the attempts before giving up.

I think this was an attempt to adhere somewhat to common guidance for reducing tail latency by starting with a short timeout in a scenario where it's better to give up quickly and retry vs being stuck in a potential long tail.

I'm definitely down to simplify it if possible, though I don't think there's much that can be done other than tweaking the max attempts and the timeout logic. What do you think would be appropriate?

client.go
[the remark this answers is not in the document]
brandur: > I kept name as a standalone function arg because it is required and is the way to identify the row being updated.

SG.

producer.go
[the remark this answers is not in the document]
brandur: So I think my main objection here is that we're adding a new primary key type, even though we already have all the other primary key types. If we're already rolling with int64s and text fields, I think we should keep rolling with those instead of adding yet another new convention.

I was just reading through all this code, and it actually looks like this could be tweaked a little to have the ID lazily marshal itself, so that in case of the first startup, it acquires a new one from `ProducerInit`, but after doing so (subsequent startups), it'll send its own ID again to attempt to retain it.

If we did that, we could just make `producer.id` a standard bigserial — more conventional, easy for humans to work with, and half the size — thoughts?

producer.go
brandur: > I'm definitely down to simplify it if possible, though I don't think there's much that can be done other than tweaking the max attempts and the timeout logic. What do you think would be appropriate?

I'm not sure, but every step of the process should be defensible in why it's like that. For example, it takes a `min(...)` between the algo you chose + 10 seconds. Why 10 seconds? Why 5 attempts? Should any of these numbers be more standard?

I was just looking at the numbers produced by this algo, and the `min` on 10 seconds truncates those later runs a lot. That's probably a good thing, but if the exponential part of the algorithm is really not even used then, should we even bother with it?

```
irb(main):002:0> 5.times.map { |i| i ** i }
=> [1, 1, 4, 27, 256]
irb(main):003:0> 5.times.map { |i| [i ** i, 10].min }
=> [1, 1, 4, 10, 10]
```

Even if we don't standardize things right now, we should still try to rationalize it in a comment or something.

As I was re-reading this, I realized also that none of this is actually having any effect anyway. If you look closely at `baseCtx` versus `ctx`, the `ctx` is where the timeout is applied, but it's `baseCtx` gets used everywhere. The only place `ctx` (and the timeout) is used is for the error log on the line just below the shutdown.

Side note on that: use of things like `baseCtx` is contrary to normal convention, and I think this demonstrates that this sort of thing can be quite error prone and confusing because you've got multiple independent contexts doing multiple things and the smallest copy/paste error can start using the wrong one in the wrong place (I thought maybe it was just me, but this mistake seems to indicate that it's not). IMO we should get everything back to `ctx`, and use scoping blocks (`{ ... }`) to control context lifetime.

producer.go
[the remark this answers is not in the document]
bgentry: I think that could work, although I'm not quite sure what to do with the standard pilot and non-pro producers where there is no DB record and thus no int64 to hand out at `ProducerInit`. Maybe use an atomic sequence on the `StandardProducer` and don't worry about the fact that the same IDs will be used on different clients? Hopefully that assumption holds true.

producer.go
bgentry: Reworked this to utilize a closure and use a simpler formula, adjusting only timeouts and with no sleeping between attempts. New values:

```go
		// Attempt 1: 100ms, Attempt 2: 500ms, Attempt 3: 2.5s, Attempt 4: 12.5s
		timeout := baseTimeout * time.Duration(math.Pow(5, float64(attempt-1)))
```

producer_test.go
```diff
@@ -160,6 +161,7 @@ func TestProducer_PollOnly(t *testing.T) {
 			archetype = riversharedtest.BaseServiceArchetype(t)
 			driver    = riverpgxv5.New(nil)
 			pilot     = &riverpilot.StandardPilot{}
+			queueName = fmt.Sprintf("test-producer-poll-only-%05d", randutil.IntBetween(1, 100_000))
```
bgentry, line 164: Somehow I keep ending up with a `default` queue left in my test DB. I'm not sure where it's coming from, but its presence can cause local flakiness. I'm hoping to track down the source, but meanwhile we can also use random queue names with test-specific prefixes to avoid any collisions and to help track down orphaned rows that somehow aren't deleted by truncation logic.

producer.go
brandur: K thanks, that looks good.

producer_test.go
brandur: Yeah I've been getting that too. I think it comes from running the River Pro test suite which is using the same same test DB, but I haven't spent the time to definitively track it down yet either.

brandur: Added https://github.com/riverqueue/riverpro/pull/112, which in my localized experiments, seems to fix the problem.

producer.go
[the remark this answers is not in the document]
bgentry: This is what I opted for in the latest version :v;

client.go
bgentry: Ok @brandur I think this queue bundle / queue add concern is the last open comment to resolve here, lmk your thoughts :pray:

producer.go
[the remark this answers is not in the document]
bgentry: Dang, unfortunately this causes a data race because `StartWorkContext` attempts to write the producer ID while `Stop()` might be attempting to read it. I can mutex or make it into an `atomic.Int64` I guess, but that feels pretty gross for an internal ID value 🤔 Will come back at it with a fresh brain in the morning to see if I can spot a good fix.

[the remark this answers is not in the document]
bgentry: I went ahead and converted it to an `atomic.Int64`. A bit ugly, but it seems unavoidable given the access pattern.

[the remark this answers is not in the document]
brandur: Nice.

Can you push the schema changes over in the pro section half of this? We should either have a globally unique int ID or a compound key of (client_id, producer_id), and just want to double check that it looks good over there.

client.go
brandur: Ugh, yeah none of the available options seem that great to me, so no strong preference I guess.

An alternative similar to the last option would be to override `riverpro.Client.Queues()` to return a `ProQueueBundle` struct that embeds `river.QueueBundle`. It'd be minimally functional for the time being, but we could add an `AddPro` to it in addition to the existing `Add`.

brandur (approved):
Thx for the fixes.

PR #826, opened 2025-04-01
cleanly error adding queue to client w/out workers

bgentry:
Prior to this change, calling `.Add()` on the `QueueBundle` for a `Client` which wasn't configured to run any jobs (i.e. no `Workers` configured) resulted in a panic due to a missing completer internally:

```
panic: producerConfig.Completer is required [recovered]
	panic: producerConfig.Completer is required
```

This changes the behavior to cleanly error with a descriptive message instead. Open to any cleaner way to improve this if you have ideas! :pray:


brandur (approved):
Ah, doh. Yeah improving the error seems like a good option for now.

PR #830, opened 2025-04-03
Expose `ErrShutdown` sentinel error for client shutdown indication

mastercactapus:
This PR exposes `ErrShutdown` from `internal/rivercommon` so callers can use `errors.Is` to determine if the error was caused by shutdown.

### DIFF
--- a/error.go
+++ b/error.go
@@ -3,11 +3,15 @@ package river
 import (
 	"time"
 
+	"github.com/riverqueue/river/internal/rivercommon"
 	"github.com/riverqueue/river/rivertype"
 )
 
 // ErrJobCancelledRemotely is a sentinel error indicating that the job was cancelled remotely.
 var ErrJobCancelledRemotely = rivertype.ErrJobCancelledRemotely
+
+// ErrShutdown is a sentinel error indicating that the client is shutting down.
+var ErrShutdown = rivercommon.ErrShutdown
 
 // JobCancelError is the error type returned by JobCancel. It should not be
 // initialized directly, but is returned from the [JobCancel] function and can
--- a/internal/rivercommon/river_common.go
+++ b/internal/rivercommon/river_common.go
@@ -20,6 +20,5 @@ type ContextKeyClient struct{}
 
 // ErrShutdown is a special error injected by the client into its fetch and work
 // CancelCauseFuncs when it's stopping. It may be used by components for such
-// cases like avoiding logging an error during a normal shutdown procedure. This
-// is internal for the time being, but we could also consider exposing it.
+// cases like avoiding logging an error during a normal shutdown procedure.
 var ErrShutdown = errors.New("shutdown initiated")


brandur:
@mastercactapus Can you say a little more about why you ended up needing this?

mastercactapus:
@brandur This error is returned by the public `Start` method, but there's no way to validate that other than string comparison (that I'm aware of) -- because it's internal; we're working around it currently [here](https://github.com/target/goalert/blob/a319217fd5bb0b1f0cc1763dd4709af92ae8e622/engine/engine.go#L570).

Our case is essentially direct from the comment on the sentinel error itself:
> It may be used by components for such cases like avoiding logging an error during a normal shutdown procedure.

It came up by causing [test failures in CI](https://ci.dev.target.goalert.me/builds/44099794#L67ec7831:698) where our smoke tests monitor for error logs while stimulating various scenarios (including [restarting the application](https://github.com/target/goalert/blob/a319217fd5bb0b1f0cc1763dd4709af92ae8e622/test/smoke/dataenckeyrotation_test.go#L22)).

bgentry:
@brandur so we're explicitly returning this internal error here:

https://github.com/riverqueue/river/blob/6865991eb1b35c4d02c60819af744167887379fa/client.go#L888-L894

I think there's a reasonable argument that given we're intentionally trying to return a specific error if it occurs due to shutdown/context cancellation, we may want to expose that in some way to users. That, or it's meant to be kept internal-only and is only to be used as an indicator for our own code to decide i.e. not to log error messages.

In either case, if it's going to be exposed, we should probably move it out of the internal package altogether and not merely alias it via the main `river` package.

brandur:
@bgentry 

> I think there's a reasonable argument that given we're intentionally trying to return a specific error if it occurs due to shutdown/context cancellation, we may want to expose that in some way to users. That, or it's meant to be kept internal-only and is only to be used as an indicator for our own code to decide i.e. not to log error messages.

Yeah makes sense.

Just as a gut check though — did we possibly make a mistake before in having this bubble up to users? I'm just thinking that in most cases in Go, functions should return an error of `nil` unless a real error would occur. Operations like a shut down should generally not result in an error because callers do things like `if err := client.Start(ctx); err != nil { panic(err) }`, and wouldn't know to expect an error as part of normal operations unless they did a very thorough read of docs or via empirical experimentation.

An alternate fix for this might be to make sure to always intercept the shutdown error and return `nil` instead. e.g. This change might need a bit more work, but it passes tests:

``` diff
index 6367df3..145b5bb 100644
--- a/client.go
+++ b/client.go
@@ -888,7 +888,7 @@ func (c *Client[TTx]) Start(ctx context.Context) error {
 	}(); err != nil {
 		defer stopped()
 		if errors.Is(context.Cause(fetchCtx), startstop.ErrStop) {
-			return rivercommon.ErrShutdown
+			return nil
 		}
 		return err
 	}
diff --git a/client_test.go b/client_test.go
index 167d224..32b8309 100644
--- a/client_test.go
+++ b/client_test.go
@@ -331,10 +331,7 @@ func Test_Client(t *testing.T) {
 					err := client.Queues().Add(fmt.Sprintf("new_queue_%d_%d_before", workerNum, j), QueueConfig{MaxWorkers: 1})
 					require.NoError(t, err)
 
-					err = client.Start(ctx)
-					if !errors.Is(err, rivercommon.ErrShutdown) {
-						require.NoError(t, err)
-					}
+					require.NoError(t, client.Start(ctx))
 
 					err = client.Queues().Add(fmt.Sprintf("new_queue_%d_%d_after", workerNum, j), QueueConfig{MaxWorkers: 1})
 					require.NoError(t, err)
@@ -1091,7 +1088,7 @@ func Test_Client(t *testing.T) {
 
 		clientWithStop := &clientWithSimpleStop[pgx.Tx]{Client: client}
 
-		startstoptest.StressErr(ctx, t, clientWithStop, rivercommon.ErrShutdown)
+		startstoptest.Stress(ctx, t, clientWithStop)
 	})
 }
 
@@ -1392,7 +1389,6 @@ func Test_Client_StopAndCancel(t *testing.T) {
 			t.Logf("Job waiting for context cancellation")
 			defer t.Logf("Job finished")
 			<-ctx.Done()
-			require.ErrorIs(t, context.Cause(ctx), rivercommon.ErrShutdown)
 			t.Logf("Job context done, closing chan and returning")
 			close(jobDoneChan)
 			return nil
```

> In either case, if it's going to be exposed, we should probably move it out of the internal package altogether and not merely alias it via the main `river` package.

Yeah. It doesn't look like anything in `internal/` really uses it, so it shouldn't be a problem to move.

bgentry:
Yep exactly what I was getting it. If it’s a normal shutdown condition and not actually an error, there’s no reason to make callers deal with it. That’s by far my preference if it’s straightforward to avoid.

brandur:
> Yep exactly what I was getting it. If it’s a normal shutdown condition and not actually an error, there’s no reason to make callers deal with it. That’s by far my preference if it’s straightforward to avoid.

Ah gotcha. Okay great. Let me take a shot at putting something together here.

@mastercactapus Just as an added check — if this error went away completely, would that be an adequate alternative for you?

mastercactapus:
@brandur Either approach works for our use case. Removing the error would be fine.

Upon further reflection, we misunderstood how `Start()` works. We assumed it blocked until shutdown (like `http.Serve()`), but since it returns immediately, we should fix our code to not call it in a goroutine. This would eliminate our race condition during early shutdown.

That said, there's a small benefit to keeping the error: it provides a clear guarantee that after `Start()` returns, River is either started or not. Without the error, a nil return means "maybe started" while an error means "definitely not started." This is mostly a correctness consideration rather than a practical issue for us.

mastercactapus:
@brandur That sounds reasonable to me. Since our issue is resolved and #841 has been merged, I'll close this one.

PR #833, opened 2025-04-04
Fix documentation regarding uniqueness conflicts in `InsertMany`

haines:
As far as I understand it, this caveat hasn't applied since https://github.com/riverqueue/river/pull/614?

### DIFF
--- a/client.go
+++ b/client.go
@@ -1498,14 +1498,6 @@ type InsertManyParams struct {
 //	if err != nil {
 //		// handle error
 //	}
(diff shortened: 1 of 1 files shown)


bgentry (approved):
Wonderful, thank you. Not sure how I missed those so many months after fixing the limitation 🤦‍♂️ 

@haines could you please add yourself to [the CLA](https://github.com/riverqueue/rivercla) so we can get this merged?

PR #838, opened 2025-04-07
add Client.QueueUpdateTx variant

bgentry:
I should have added this in #834, I already need it for the UI backend.


brandur (commented):
Doh, it doesn't have to be comprehensive, but there should probably be at least one test case for this just to make sure it gets some exercise in the test suite.

bgentry:
@brandur yeah my bad, I added sufficient tests for both tx and non-tx, thanks!

brandur (approved):

PR #841, opened 2025-04-09
Don't error on cancelled startup to make it not necessary to export `ErrShutdown`

brandur:
This change's driven by the proposal in #830. When client start up is
cancelled by a stop, we'd return `ErrShutdown` as a special signal. The
problem is that `ErrShutdown` was never exported, so this wasn't really
usable by outside callers.

In #830 it was proposed to export `ErrShutdown`. However, after some
discussion, we suspect that it might be better to leave it unexported,
and then have `Start` return no error on cancelled startup. This
minimizes the number of startup conditions callers have to handle.

Because `ErrShutdown` wasn't previously exported, we should be able to
make this change safely without breaking anyone.

@mastercactapus did bring up a fair point that the alternative of
exporting `ErrShutdown` would allow a caller to definitively tell the
difference between the client successfully starting and backgrounding
itself versus one whose start was cancelled. I think we're okay not to
require that differentiation though because if `ErrShutdown` would've be
returned, then the caller would have called `Stop`, so presumably they'd
know to expect an successful start.

I figure that in case a strong case for wanting to differentiate the
cases comes through, we could always resurrect the error.

Also, rename the internal `ErrShutdown` to `ErrStop` to be more
consistent with the naming of `Client.Stop` (which had previously been
called `Shutdown`). This won't make any difference, but will be a little
better in case we ever do export it and then forget to rename it at that
time.


brandur:
Thx.

PR #842, opened 2025-04-09
Link live demo from main River README

brandur:
I was looking for the live demo this morning. You can get to it by
clicking through to "River UI" and then clicking live demo from there,
but it probably doesn't hurt to have a direct link to it from the main
README so it's one fewer hop to get there (and more people might see
it).


bgentry (approved):
Excellent, thanks!

PR #844, opened 2025-04-11
Logging package to add context logger for workers and stores output to metadata

brandur:
After adding support for job output, it got me thinking that it might
not be _that_ bad if we put in a way for users to be able to store
limited job-specific logging to job records. Like with output, this
would go in TOAST, and although it could be bad if too much data
accumulated, it shouldn't generally effect performance.

Here, prototype what that would look like. It's implemented as a
middleware in a separate `riverlog` package, which must be installed to
a client or worker to become active. Once it is, logging looks like:

    func (w *LoggingWorker) Work(ctx context.Context, job *river.Job[LoggingArgs]) error {
        riverlog.Logger(ctx).InfoContext(ctx, "Logged from worker")
        return nil
    }

The middleware takes a function that returns a slog handler, which has
the purpose of letting the user select between JSON, text, or something
else, and lets them select options like logging level.

I've left in enough flexibility that we should be able to extend the
interface only slightly to implement a system that sends logs to S3
instead of metadata, but metadata is a good default.

I figure that if we go down this route we can add special handling for
it to River UI to make logs easily accessible for installations even
without a full logging platform.

I also ended up adding support to `Client` and `rivertest.Worker` for
`rivershared.BaseService` in hooks and middleware so that if a hook or
middleware has a base service embedded, it gets its archetype
initialized just like other services in the client. This gives us an
easy way to inherit a client's log instead of a separate one having to
be injected which is _very_ convenient. This shouldn't be used outside
of River core packages, but it should be okay if we make use it.

Also added convenience helpers to make middleware like we already had
for hooks:

* `JobInsertMiddlewareFunc`
* `WorkerMiddlewareFunc `

I put these in because they're great for creating a hook or middleware
that's scoped to a single test case and which is a struct that can embed
something like `baseservice.BaseService`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -597,6 +597,12 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		}
 	}
 
+	for _, hook := range config.Hooks {
+		if withBaseService, ok := hook.(baseservice.WithBaseService); ok {
+			baseservice.Init(archetype, withBaseService)
+		}
+	}
+
 	client := &Client[TTx]{
 		config:               config,
 		driver:               driver,
@@ -636,6 +642,13 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 
 			middleware = append(middleware, workerMiddleware)
 		}
+
+		for _, middleware := range middleware {
+			if withBaseService, ok := middleware.(baseservice.WithBaseService); ok {
+				baseservice.Init(archetype, withBaseService)
+			}
+		}
+
 		client.middlewareLookupGlobal = middlewarelookup.NewMiddlewareLookup(middleware)
 	}
 
--- a/example_graceful_shutdown_test.go
+++ b/example_graceful_shutdown_test.go
@@ -169,5 +169,5 @@ func Example_gracefulShutdown() {
 	// Received SIGINT/SIGTERM; initiating soft stop (try to wait for jobs to finish)
 	// Received SIGINT/SIGTERM again; initiating hard stop (cancel everything)
 	// Job cancelled
-	// JobExecutor: Job errored; retrying
+	// jobexecutor.JobExecutor: Job errored; retrying
 }
--- a/middleware_defaults.go
+++ b/middleware_defaults.go
@@ -25,6 +25,16 @@ func (d *JobInsertMiddlewareDefaults) InsertMany(ctx context.Context, manyParams
 	return doInner(ctx)
 }
 
+// JobInsertMiddlewareFunc is a convenience helper for implementing
+// rivertype.JobInsertMiddleware using a simple function instead of a struct.
+type JobInsertMiddlewareFunc func(ctx context.Context, manyParams []*rivertype.JobInsertParams, doInner func(ctx context.Context) ([]*rivertype.JobInsertResult, error)) ([]*rivertype.JobInsertResult, error)
+
+func (f JobInsertMiddlewareFunc) InsertMany(ctx context.Context, manyParams []*rivertype.JobInsertParams, doInner func(ctx context.Context) ([]*rivertype.JobInsertResult, error)) ([]*rivertype.JobInsertResult, error) {
+	return f(ctx, manyParams, doInner)
+}
+
+func (f JobInsertMiddlewareFunc) IsMiddleware() bool { return true }
+
 // WorkerInsertMiddlewareDefaults is an embeddable struct that provides default
 // implementations for the rivertype.WorkerMiddleware. Use of this struct is
 // recommended in case rivertype.WorkerMiddleware is expanded in the future so
@@ -36,3 +46,13 @@ type WorkerMiddlewareDefaults struct{ MiddlewareDefaults }
 func (d *WorkerMiddlewareDefaults) Work(ctx context.Context, job *rivertype.JobRow, doInner func(ctx context.Context) error) error {
 	return doInner(ctx)
 }
+
+// WorkerMiddlewareFunc is a convenience helper for implementing
+// rivertype.WorkerMiddleware using a simple function instead of a struct.
+type WorkerMiddlewareFunc func(ctx context.Context, job *rivertype.JobRow, doInner func(ctx context.Context) error) error
+
+func (f WorkerMiddlewareFunc) Work(ctx context.Context, job *rivertype.JobRow, doInner func(ctx context.Context) error) error {
+	return f(ctx, job, doInner)
+}
+
+func (f WorkerMiddlewareFunc) IsMiddleware() bool { return true }
--- a/recorded_output.go
+++ b/recorded_output.go
@@ -10,7 +10,10 @@ import (
 	"github.com/riverqueue/river/rivertype"
 )
 
-const maxOutputSize = 32 * 1024 * 1024 // 32MB
+const (
+	maxOutputSizeMB    = 32
+	maxOutputSizeBytes = maxOutputSizeMB * 1024 * 1024
+)
 
 // RecordOutput records output JSON from a job. The "output" can be any
 // JSON-encodable value and will be stored in the database on the job row after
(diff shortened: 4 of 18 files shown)


brandur:
@bgentry Wanted to get a rough gauge on this one before writing tests or anything, but I put in a quick example test to show what the usage looks like. General thoughts?

recorded_output.go
```diff
@@ -55,8 +58,8 @@ func RecordOutput(ctx context.Context, output any) error {
 
 	// Postgres JSONB is limited to 255MB, but it would be a bad idea to get
 	// anywhere close to that limit here.
-	if len(metadataUpdatesBytes) > maxOutputSize {
-		return fmt.Errorf("output is too large: %d bytes (max 32 MB)", len(metadataUpdatesBytes))
+	if len(metadataUpdatesBytes) > maxOutputSizeBytes {
+		return fmt.Errorf("output is too large: %d bytes (max %d MB)", len(metadataUpdatesBytes), maxOutputSizeMB)
```
bgentry, line 62: definitely better!

riverlog/river_log.go
```diff
@@ -0,0 +1,135 @@
...
+	// Be careful with this number because the maximum total log size is equal
+	// to maximum number of attempts multiplied by this number (each attempt's
+	// logs are kept separately). For example, 25 * 2 MB = 50 MB maximum
+	// theoretical log size. Log data goes into metadata which is a JSONB field,
+	// and JSONB fields have a maximum size of 255 MB, so any number larger than
+	// 255 divided by maximum number of attempts may cause serious operational
+	// problems.
+	//
+	// Defaults to 2 MB (which is per job attempt).
+	MaxSizeBytes int
```
bgentry, line 64: This made me wonder if somebody abusing this is likely to run into issues with the max number of bytes being sent in a single query (for the completer). I know there's a max param count but unsure about the bytes limit, we might want to look that up.

bgentry (commented):
Love it! I know the example needs to be message only in order to be repeatable as a test, but I wonder how we can best illustrate the "right" way to use it? For most people they'd probably want timestamps as well as key/value params to be printed as well.

rivertest/worker_test.go
```diff
@@ -326,6 +327,88 @@ func TestWorker_Work(t *testing.T) {
 		require.ErrorContains(t, err, "failed to insert job: tx is closed")
 		require.Nil(t, res)
 	})
+
+	t.Run("WorkUsesHooks", func(t *testing.T) {
```
brandur, line 331: These test cases verify that (1) hooks/middlewares are used in the test worker, and (2) that a hook/middleware with base service is initialized. We weren't previously testing for (1), so I wrapped them up together.

client.go
```diff
@@ -597,6 +597,12 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		}
 	}
 
+	for _, hook := range config.Hooks {
+		if withBaseService, ok := hook.(baseservice.WithBaseService); ok {
+			baseservice.Init(archetype, withBaseService)
+		}
+	}
```
brandur, line 604: I added some code to the client and test worker so that if hooks/middleware are base services, we'll initialize them. This isn't intended for external use, but makes it really convenient for sharing things like a logger so that those don't need to be configured into the hook/middleware separately.

example_graceful_shutdown_test.go
```diff
@@ -169,5 +169,5 @@ func Example_gracefulShutdown() {
 	// Received SIGINT/SIGTERM; initiating soft stop (try to wait for jobs to finish)
 	// Received SIGINT/SIGTERM again; initiating hard stop (cancel everything)
 	// Job cancelled
-	// JobExecutor: Job errored; retrying
+	// jobexecutor.JobExecutor: Job errored; retrying
```
brandur, line 172: I modified the service naming logic a little below so that we produce names like `riverlog.Middleware`. Top-level River constructs don't include the package so it just looks like `Client` (with the idea that it's in River being implied).

middleware_defaults.go
```diff
@@ -25,6 +25,16 @@ func (d *JobInsertMiddlewareDefaults) InsertMany(ctx context.Context, manyParams
 	return doInner(ctx)
 }
 
+// JobInsertMiddlewareFunc is a convenience helper for implementing
+// rivertype.JobInsertMiddleware using a simple function instead of a struct.
+type JobInsertMiddlewareFunc func(ctx context.Context, manyParams []*rivertype.JobInsertParams, doInner func(ctx context.Context) ([]*rivertype.JobInsertResult, error)) ([]*rivertype.JobInsertResult, error)
```
brandur, line 30: Added func helpers for middlewares because it makes it really easy to isolate middlewares to specific test cases, even if the middleware needs to embed something like `baseservice.BaseService`.

riverlog/river_log.go
brandur: I looked into this a bit and I think the limit is very large in practice. This answer indicates 1 GB:

https://dba.stackexchange.com/a/131425

I think a good improvement for this would be to give the metadata updater a way to append to array so that we could make sure that only one attempt's logs are being sent at a time. That'll definitely take some rework to make happen though.

brandur (commented):
@bgentry Okay added some improvements and a bunch of tests. Mind taking another look?

> Love it! I know the example needs to be message only in order to be repeatable as a test, but I wonder how we can best illustrate the "right" way to use it? For most people they'd probably want timestamps as well as key/value params to be printed as well.

Hm, yeah tricky for sure. I ended up adding a comment line like:

``` go
			riverlog.NewMiddleware(func(w io.Writer) slog.Handler {
				// We have to use a specialized handler without timestamps to
				// make test output reproducible, but in reality this would as
				// simple as something like:
				//
				// 	return slog.NewJSONHandler(w, nil)
				return &slogutil.SlogMessageOnlyHandler{Out: w}
			}, nil),
```

I guess you could use a normal handler and then strip out the message below, but that also adds a lot of pretty gnarly code too.

Did you have any other ideas you were thinking of?

riverlog/common_test.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+//
+// This file used as a holding place for test helpers for examples so that the
+// helpers aren't included in Godoc and keep each example more succinct.
+//
+
+// Wait on the given subscription channel for numJobs. Times out with a panic if
+// jobs take too long to be received.
+func waitForNJobs(subscribeChan <-chan *river.Event, numJobs int) []*rivertype.JobRow {
```
bgentry, line 20: not worth putting this into an internal pkg to share it vs copy/pasting?

bgentry (approved):
Great stuff 👏

riverlog/common_test.go
brandur: Yeah, I'll put it in a `rivershared` thing. We needed it in a pro example test before too.

riverlog/river_log.go
```diff
@@ -0,0 +1,149 @@
...
+
+	"github.com/riverqueue/river/internal/jobexecutor"
+	"github.com/riverqueue/river/rivershared/baseservice"
+	"github.com/riverqueue/river/rivertype"
+)
+
+const (
+	maxSizeMB    = 2
+	maxSizeBytes = maxSizeMB * 1024 * 1024
+	metadataKey  = "river:log"
```
brandur, line 22: @bgentry Oh yeah I forgot to ask about this one. I put a "namespace" in here because I feel like something like "log" or "logs" it too generic and risks collision with things that users might've been doing.

This could also be `river_log` or something like that too. The colon is kind of XML inspired I guess. It occurred to me that maybe we should've namespaced the "job output" key as well for the same reason. Regardless, we should probably have some convention for future additions.

Thoughts?

bgentry: Yeah I saw, I don't have a strong opinion on what it should be tbh. So far I think I was thinking that for metadata we could assume it's safe to claim any top-level plain key and then tell users _they_ should namespace if using metadata; most won't touch it so this would be pretty safe advice. But not a strong preference.

brandur:
Thanks! Alright, going to get this merged in then.

PR #845, opened 2025-04-12
Replace uses of `valutil.FirstNonZero` with `cmp.Or` + deprecate former

brandur:
`cmp.Or` wasn't available originally in River, but was add in 1.22.
We're now minimum version 1.23, so it's fine to switch over it given it
does exactly the same thing as the custom `valutil.FirstNonZero`.

We have some dependencies using `valutil.FirstNonZero` so we shouldn't
remove it until there's at least one more version of those that've been
cut to avoid potential version compatibility problems. For now, mark the
function as deprecated to warn against use.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1,6 +1,7 @@
 package river
 
 import (
+	"cmp"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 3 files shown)


bgentry (approved):
Wow, thought this was way more widely used than just 4 places. Thanks!

PR #846, opened 2025-04-12
Fix bad bug in `riversharedtest.DBPool` (which is used by `TestTx`)

brandur:
Fixes a bad bug in `riversharedtest.DBPool` in which while lazily
initializing a database pool, *we never actually set the database pool*,
so repeated invocations of `DBPool` would check for an existing pool
that would never exist and create a new one each time. This project
seems to be brute forcing its way by, but I was seeing Postgres
connection exhaustions in other projects.

Boiled down, the problem is that we had this:

    dbPool, err := pgxpool.New(ctx, cmp.Or(
        os.Getenv("TEST_DATABASE_URL"),
        "postgres://localhost:5432/river_test",
    ))

When it should've been this:

    var err error
    dbPool, err = pgxpool.New(ctx, cmp.Or(
        os.Getenv("TEST_DATABASE_URL"),
        "postgres://localhost:5432/river_test",
    ))

Here, fix the problem, and go a step further to simplify the code
somewhat to use a `sync.Once` instead of mutex. The final product is
fewer lines of code, which will hopefully make a bug of this magnitude
easier to spot should it reoccur.

### DIFF
--- a/rivershared/riversharedtest/riversharedtest.go
+++ b/rivershared/riversharedtest/riversharedtest.go
@@ -32,12 +32,12 @@ func BaseServiceArchetype(tb testing.TB) *baseservice.Archetype {
 	}
 }
 
-// A pool and mutex to protect it, lazily initialized by TestTx. Once open, this
+// A pool and sync.Once to initialize it, invoked by TestTx. Once open, this
 // pool is never explicitly closed, instead closing implicitly as the package
 // tests finish.
 var (
-	dbPool   *pgxpool.Pool //nolint:gochecknoglobals
-	dbPoolMu sync.RWMutex  //nolint:gochecknoglobals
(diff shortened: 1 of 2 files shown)


bgentry (approved):
This was a a pre-existing issue, but the only remaining problem I'm seeing is that if a transient error occurs during `pgxpool.New` it becomes unrecoverable. We could in theory add some retry logic if that happens much, but hopefully it's not needed when connecting to a local DB (even in bad CI envs).

brandur:
> This was a a pre-existing issue, but the only remaining problem I'm seeing is that if a transient error occurs during `pgxpool.New` it becomes unrecoverable.

I added another `require` after the once so that each test case makes sure that the `dbPool` has a reasonable value to use.

PR #848, opened 2025-04-17
Full schema insertion + schema testing

brandur:
Here, introduce a full version of schema injection as initially started
in #798, and exposing a new `Schema` option to make it configurable even
outside the test suite.

This involved fixing a lot of bugs from #798, and the only way to make
it possible to root them all out was to make full use of schemas across
the entire test suite. The original "test DB manager" system has been
replaced with a new `riverschematest.TestSchema` helper that generates a
schema for use with a test case or prefers an existing idle one that was
already generated for the same test run.

`TestSchema` runs migrations for generated schemas which also means we
don't need to use `testdbman` anymore, with tests capable of
bootstrapping themselves at run time. We reuse schemas to avoid this
extra work when possible, but migrating a new schema is also
surprisingly fast, taking up to 50ms, but getting more down to a stable
~10ms once things are warmed up. Here's a series of sample timings:

    $ go test ./rivermigrate -run TestMigrator/MigrateUpDefault -test.v -count 50 | grep 'migrations ran in'
        river_migrate_test.go:492: migrations ran in 45.571209ms
        river_migrate_test.go:492: migrations ran in 24.642458ms
        river_migrate_test.go:492: migrations ran in 16.749708ms
        river_migrate_test.go:492: migrations ran in 22.970375ms
        river_migrate_test.go:492: migrations ran in 16.201375ms
        river_migrate_test.go:492: migrations ran in 15.727625ms
        river_migrate_test.go:492: migrations ran in 13.291333ms
        river_migrate_test.go:492: migrations ran in 13.680708ms
        river_migrate_test.go:492: migrations ran in 14.867416ms
        river_migrate_test.go:492: migrations ran in 15.631916ms
        river_migrate_test.go:492: migrations ran in 13.873791ms
        river_migrate_test.go:492: migrations ran in 14.8645ms
        river_migrate_test.go:492: migrations ran in 14.92575ms
        river_migrate_test.go:492: migrations ran in 12.541834ms
        river_migrate_test.go:492: migrations ran in 14.753875ms
        river_migrate_test.go:492: migrations ran in 12.694334ms
        river_migrate_test.go:492: migrations ran in 13.955917ms
        river_migrate_test.go:492: migrations ran in 12.126458ms
        river_migrate_test.go:492: migrations ran in 14.095958ms
        river_migrate_test.go:492: migrations ran in 13.273375ms
        river_migrate_test.go:492: migrations ran in 13.988917ms
        river_migrate_test.go:492: migrations ran in 13.141459ms
        river_migrate_test.go:492: migrations ran in 12.394417ms
        river_migrate_test.go:492: migrations ran in 11.539208ms
        river_migrate_test.go:492: migrations ran in 11.577834ms
        river_migrate_test.go:492: migrations ran in 10.883375ms
        river_migrate_test.go:492: migrations ran in 10.547417ms
        river_migrate_test.go:492: migrations ran in 12.330375ms
        river_migrate_test.go:492: migrations ran in 11.54575ms
        river_migrate_test.go:492: migrations ran in 11.437458ms
        river_migrate_test.go:492: migrations ran in 10.957ms
        river_migrate_test.go:492: migrations ran in 10.589083ms
        river_migrate_test.go:492: migrations ran in 9.758583ms

Removal of the "test DB manager" system also means that we can ungate
test from `-p 1` because they're all able to run in parallel now. The
limiting factor I ran in is that we need to keep max pool connections
within each package's tests to a relatively modest number (I found 15
seemed to maximum success) so parallel packages don't exceed the default
Postgres configuration of 100 connections.

Something that can be kind of annoying is that in case a schema isn't
used properly somewhere in a test case (i.e. `TestSchema` is run, but
then not used), inserts/operations will go the default schema, which
will leave debris there, and that will interfere with test cases using
`TestTx` (with test DB manager, all debris would go to a different
database so you wouldn't notice). To remediate this, I've added a
cleanup hook to `TestSchema` that looks for leftovers that may have been
added to the default schema. This isn't perfect because those leftovers
may have come from another test case running in parallel or which ran
previously, but it helps to zero in on the original source of the issue.


brandur:
@bgentry Still needs a little more work, but test runs go from ~2m45s to ~1m, which is pretty nice.

I'm not even completely sure why the improvement is that good, but I think we were waiting a lot longer than we though waiting for test databases. Although only N = num CPUs tests were running in parallel at any given time, I think what happens is that Go boots a whole bunch of test cases, but may pause them while it engages others, some of which may have been holding a test database and leaving the available pool smaller for other test cases that might need one. Ungating the number of schemas allowed at any given time I found that the main package will generate up to 125 of them before tests start finishing fully and it starts reusing them. Not 100% sure about this theory, but whatever it is exactly, the speed increase dramatic.

I'm still blocked on one major thing which is that sqlc's `copyfrom` is totally screwed and won't even work with the `sqlctemplate` system. I got the tests running by editing the generated Go manually to enable a schema to be injected, but still trying to figure out what to do about it.

``` diff
diff --git a/riverdriver/riverpgxv5/internal/dbsqlc/copyfrom.go b/riverdriver/riverpgxv5/internal/dbsqlc/copyfrom.go
index ee29edf..5d18b02 100644
--- a/riverdriver/riverpgxv5/internal/dbsqlc/copyfrom.go
+++ b/riverdriver/riverpgxv5/internal/dbsqlc/copyfrom.go
@@ -48,6 +48,11 @@ func (r iteratorForJobInsertFastManyCopyFrom) Err() error {
 	return nil
 }
 
-func (q *Queries) JobInsertFastManyCopyFrom(ctx context.Context, db DBTX, arg []*JobInsertFastManyCopyFromParams) (int64, error) {
-	return db.CopyFrom(ctx, []string{"river_job"}, []string{"args", "created_at", "kind", "max_attempts", "metadata", "priority", "queue", "scheduled_at", "state", "tags", "unique_key", "unique_states"}, &iteratorForJobInsertFastManyCopyFrom{rows: arg})
+func (q *Queries) JobInsertFastManyCopyFrom(ctx context.Context, db DBTX, schema string, arg []*JobInsertFastManyCopyFromParams) (int64, error) {
+	tableName := []string{"river_job"}
+	if schema != "" {
+		tableName = append([]string{schema}, tableName...)
+	}
+
+	return db.CopyFrom(ctx, tableName, []string{"args", "created_at", "kind", "max_attempts", "metadata", "priority", "queue", "scheduled_at", "state", "tags", "unique_key", "unique_states"}, &iteratorForJobInsertFastManyCopyFrom{rows: arg})
 }
```

brandur:
@bgentry Alright, I managed to resolve the copyfrom problem by leveraging pgx's generated interface once again and putting some code in `templateReplaceWrapper` that'll add a schema if necessary. It's all cleaner than I expected.

I'm still writing tests for `riverschematest` and did find a couple minor problems with the design there, but I think this is pretty much good to start looking at when you have a chance.

I also rebuilt the `TestSchema` API to use a driver instead of pgxpool, which will make sure that River Pro's migration lines are available to it so we can get a quick implementation over there.

brandur:
Alright, new tests in place for `riverschematest`. Everything here should be 100% functional.

brandur:
I put in one small improvement that zeroes the `search_path` when using a schema which _forces_ callers to use a schema lest they error because there's no opportunity to fall back on `public`. This helps suss out mistakes immediately as they happen rather than having to wait until the end of the test.

example_job_cancel_from_client_test.go
```diff
@@ -38,19 +40,14 @@ func (w *SleepingWorker) Work(ctx context.Context, job *river.Job[CancellingArgs
 func Example_jobCancelFromClient() {
 	ctx := context.Background()
 
-	dbPool, err := pgxpool.NewWithConfig(ctx, riverinternaltest.DatabaseConfig("river_test_example"))
+	jobChan := make(chan int64)
```
bgentry, line 43: unintentional relocation? IMO it makes more sense down near where it's used

internal/maintenance/job_scheduler_test.go
```diff
@@ -30,13 +31,18 @@ func TestJobScheduler(t *testing.T) {
 		notificationsByQueue map[string]int
 	}
 
-	setup := func(t *testing.T, exec riverdriver.Executor) (*JobScheduler, *testBundle) {
+	type testOpts struct {
+		exec   riverdriver.Executor
+		schema string // for use when using a non-Testx
```
bgentry, line 36:
```suggestion
		schema string // for use when using a non-TestTx
```

riverdriver/river_driver_interface.go
```diff
@@ -453,7 +479,7 @@ type JobSetStateIfRunningManyParams struct {
 	MetadataDoMerge []bool
 	MetadataUpdates [][]byte
 	ScheduledAt     []*time.Time
-	Schema          string
+	Schema          string // TODO(brandur): Try to push this up a layer; it's awkward being the only non-slice on here.
```
bgentry, line 482: Hmm, while it's maybe a little awkward, it would also seem weird to do anything else here given there can only be one value no matter how may rows are updated (which is unique from the other fields).

riverdriver/riverdatabasesql/river_database_sql_driver.go
```diff
@@ -873,6 +901,12 @@ func (t *ExecutorSubTx) Begin(ctx context.Context) (riverdriver.ExecutorTx, erro
 	if err != nil {
 		return nil, err
 	}
+
+	_, err = t.Exec(ctx, "SET search_path TO 'public'")
+	if err != nil {
+		return nil, err
+	}
```
bgentry, line 908: Hmm, an added round trip for every use here? That doesn't seem good, any way we can avoid it?

riverschematest/riverschematest.go
```diff
@@ -0,0 +1,377 @@
...
+		if packageName == "riverschematest" && !opts.skipPackageNameCheck {
+			panic("package name should not resolve to riverschematest")
+		}
+
+		// Notification topics are prefixed with schemas. The max Postgres
+		// length of topics is 63, and "river_leadership" is the longest topic
+		// name. If the package suffix is longer than the max that could fit
+		// into 63 when combined with a schema name and leadership, trim it down
+		// a bit. The only package where this is needed as I write this is
+		// `riverencrypt_test`. If this happens in too many places we may want
```
bgentry, line 122: Do we need to validate the length of the schema param in the river config to avoid issues with this limit?

rivertest/example_require_inserted_test.go
```diff
@@ -81,12 +87,17 @@ func Example_requireInserted() {
...
+	// way of injecting a schema at the pool level. The rivertest API will need
+	// to be expanded to allow it.
+	/*
+		// Insert and verify one on a pool instead of transaction.
+		_, err = riverClient.Insert(ctx, &RequiredArgs{Message: "Hello from pool."}, nil)
+		if err != nil {
+			panic(err)
+		}
+		_ = rivertest.RequireInserted(ctx, t, riverpgxv5.New(dbPool), &RequiredArgs{}, nil)
+	*/
```
bgentry, line 100: Ugh, not great 😕 But maybe the schema can just go on [`RequireInsertedOpts`](https://pkg.go.dev/github.com/riverqueue/river/rivertest#RequireInsertedOpts) for those using non-default schemas?

riverschematest/riverschematest.go
brandur: Yeah good call. I just put that in.

riverdriver/riverdatabasesql/river_database_sql_driver.go
brandur: Oops, this was a mistake. I took it back out again.

client_test.go
```diff
@@ -200,25 +208,36 @@ func Test_Client(t *testing.T) {
...
 		}
 	}
 
 	setup := func(t *testing.T) (*Client[pgx.Tx], *testBundle) {
 		t.Helper()
 
 		config, bundle := setupConfig(t)
+
+		// TODO(brandur): It'd be better if we could reuse the driver object
+		// rfrom setupConfig here, but it'll take a lot of unwinding of all
```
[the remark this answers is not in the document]
brandur: Thx.

client_test.go
```diff
@@ -326,6 +345,7 @@ func Test_Client(t *testing.T) {
 
 	t.Run("Queues_Add_Stress", func(t *testing.T) {
 		t.Parallel()
+		t.Skip() // TODO
```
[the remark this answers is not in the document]
brandur: Whoops, removed.

client_test.go
```diff
@@ -1130,6 +1159,7 @@ func Test_Client(t *testing.T) {
 
 	t.Run("StartStopStress", func(t *testing.T) {
 		t.Parallel()
+		t.Skip() // TODO
```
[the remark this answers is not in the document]
brandur: Removed.

internal/maintenance/job_scheduler_test.go
brandur: Fixed, thx.

example_job_cancel_from_client_test.go
brandur: Yeah agreed. I just moved it back.

client_test.go
```diff
@@ -5168,36 +5517,44 @@ func Test_Client_JobCompletion(t *testing.T) {
...
 				State:               rivertype.JobStateCompleted,
 			})
-			require.NoError(err)
+			require.NoError(t, err)
 			return nil
 		})
 
 		client, bundle := setup(t, config)
 		exec = client.driver.GetExecutor()
+		schema = client.config.Schema
```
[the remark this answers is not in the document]
brandur: Yeah, great point. I think I'd just been on a run of trying to fix like 100 tests in a row when I wrote this and was too lazy to bother. Switched over to an internal job args and it was fixed up easily.

riverdriver/river_driver_interface.go
brandur: Yeah, true. I took that TODO back out. This seems fine for now.

client_test.go
```diff
@@ -5325,7 +5682,7 @@ func Test_Client_JobCompletion(t *testing.T) {
 		require.NotNil(updatedJob.FinalizedAt)
 
 		// Make sure the FinalizedAt is approximately ~now:
-		require.WithinDuration(now, *updatedJob.FinalizedAt, 2*time.Second)
+		require.WithinDuration(now, *updatedJob.FinalizedAt, 5*time.Second) // TODO: Bad clock-based test
```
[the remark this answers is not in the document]
brandur: Yeah that works. I did have to change `JobCompleteTx` to use the stub instead of `time.Now()` though:

``` diff
diff --git a/job_complete_tx.go b/job_complete_tx.go
index be7d0e1..6f75c60 100644
--- a/job_complete_tx.go
+++ b/job_complete_tx.go
@@ -53,7 +53,7 @@ func JobCompleteTx[TDriver riverdriver.Driver[TTx], TTx any, TArgs JobArgs](ctx
 	}
 
 	execTx := driver.UnwrapExecutor(tx)
-	params := riverdriver.JobSetStateCompleted(client.config.Schema, job.ID, time.Now(), nil)
+	params := riverdriver.JobSetStateCompleted(client.config.Schema, job.ID, client.baseService.Time.NowUTC(), nil)
 	rows, err := pilot.JobSetStateIfRunningMany(ctx, execTx, &riverdriver.JobSetStateIfRunningManyParams{
 		ID:              []int64{params.ID},
 		Attempt:         []*int{params.Attempt},
```

rivershared/testfactory/test_factory.go
```diff
@@ -31,6 +31,7 @@ type JobOpts struct {
 	Priority     *int
 	Queue        *string
 	ScheduledAt  *time.Time
+	Schema       string
```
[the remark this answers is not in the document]
brandur: Yeah true, but hmm, I'm not sure that we can get around this for now as the schema needs to be applied for every operation. I could set a `search_path` for everything, but that does reduce the test coverage we have on explicit injection.

I think there may be some better set of helpers that we could make here where tests are able to exert some context on the DB factory by telling them what schema they've already generated, but I don't think it's perfectly clear without some experimentation whether it'd be better or worse compared to the explicit version.

rivertest/example_require_inserted_test.go
brandur: Yeah I know :/ Dang, yeah I definitely thought of putting it in `RequireInsertedOpts`, but the problem is basically two-fold:

* `RequireInsertedOpts` can be specified multiple times when requiring many jobs at once, but the implementation of the test assertions means that only one schema can be pulled out of at a time because we try to be as efficient as possible by fetching everything at once. That means that if different schemas are specified in different `RequiredInsertedOpts`, we'd need to reimplement everything to enable this.
* It just seems ugly that you'd have to specify this on every single use of the insertion assertions.

I'm not sure I'll be able to find another way because the generics make things awkward, but I think we'd at least have to try as many other options as we can think of to avoid the problems above.

`Schema` being broken here in the interim isn't great, but I figure this is really a place where use of `search_path` has no disadvantages whatsoever. You can make the argument that `search_path` in prod is a problem because of PgBouncer, but you'll never be using that in tests, so I think we can recommend that.

brandur:
@bgentry Alright, I think I addressed all feedback. Mind taking another look?

Also, sorry to do this so late, but I pushed a second commit that moves `TestTx` over to use `TestSchema` as well. This is kind of nice because:

* We no longer need to run migrations in the set up instructions. The test tx schema is migrated once, then used for the rest of the test run.
* No more weird collisions in `river_test` due to alternatively running tests for River and River Pro. Both projects get their own schema with the right migrations in each.
* Any preexisting data that might've been left over in `river_test` is ignored.

internal/jobcompleter/job_completer_test.go
```diff
@@ -167,14 +172,14 @@ func TestAsyncJobCompleter_Complete(t *testing.T) {
 
 	// launch 4 completions, only 2 can be inline due to the concurrency limit:
 	for i := range int64(2) {
-		if err := completer.JobSetStateIfRunning(ctx, &jobstats.JobStatistics{}, riverdriver.JobSetStateCompleted(i, time.Now(), nil)); err != nil {
+		if err := completer.JobSetStateIfRunning(ctx, &jobstats.JobStatistics{}, riverdriver.JobSetStateCompleted(schema, i, time.Now(), nil)); err != nil {
```
bgentry, line 175: Related to a previous comment, do we actually need/want the schema to be part of the individual completed params i.e. `JobSetStateCompleted`? Every job being completed in a single query must be part of the same schema—maybe it's instead a config-level property of the completer set at init and doesn't need to be passed around all over the place like this? Might clean things up a lot.

internal/jobcompleter/job_completer_test.go
```diff
@@ -485,6 +505,14 @@ func testCompleter[TCompleter JobCompleter](
 	// interface like jobCompleterWithTestFacilities to expose the additional
 	// functionality, although that's not particularly beautiful either.
 	disableSleep func(completer TCompleter),
+
+	// Number of jobs to insert for the CompleteManyJobs subtest. The
+	// BatchCompleter should be tested against a very large number, but this
+	// really just isn't necessary for simpler comleters that might also take a
```
bgentry, line 511:
```suggestion
	// really just isn't necessary for simpler completers that might also take a
```

riverdriver/river_driver_interface.go
```diff
@@ -363,81 +383,89 @@ type JobSetStateIfRunningParams struct {
 	MetadataDoMerge bool
 	MetadataUpdates []byte
 	ScheduledAt     *time.Time
+	Schema          string
```
bgentry, line 386: see my other comment on this, I'm wondering if we can just get rid of this field and make it a completer level config? It can't update more than one schema at once anyway. Each of the helpers below would lose the extra arg too.

rivertest/example_require_inserted_test.go
bgentry: > * `RequireInsertedOpts` can be specified multiple times when requiring many jobs at once, but the implementation of the test assertions means that only one schema can be pulled out of at a time because we try to be as efficient as possible by fetching everything at once. That means that if different schemas are specified in different `RequiredInsertedOpts`, we'd need to reimplement everything to enable this.

Maybe we need to do some batching by schema for those queries?

bgentry (approved):
Looks great other than one main suggestion! I think we need to do a fast follow to sort out the test helpers with multiple schemas, but we can do that in a subsequent PR.

internal/jobcompleter/job_completer_test.go
brandur: Yeah, good idea. I just changed to this, and it does clean things up quite a bit. The job executor doesn't even need to know about schema anymore because it's entirely internalized in the completer.

The one piece that I don't love about this approach is that `JobSetStateIfRunningManyParams` still needs to have `Schema` in it, because it's also what the completer uses to perform an operation with the driver. So schema is removed for each helper init function like `JobSetStateCancelled`, creates a partial `JobSetStateIfRunningManyParams` without a schema, then the completer adds schema into it later, which feels a tad janky.

It might be better to move the helpers over to be in `jobcompleter` instead, then have them create a separate schema-less struct which is then converted to the driver version of `JobSetStateIfRunningManyParams` that includes a schema. I almost did this, but it'd require 100s of lines of test changes to get there, so going to leave it as a separate improvement.

internal/jobcompleter/job_completer_test.go
brandur: Fixed, thx.

riverdriver/river_driver_interface.go
brandur: See above, but most of this has changed, although `Schema` is still needed here because it's still used by the completer when invoking the driver.

rivertest/example_require_inserted_test.go
brandur: Yeah, it's definitely possible, but I think it kinda has a couple downsides:

* It sucks that users would have to pass schema to every single `RequiredInsertedOpts`. Just not very optimal for callers.
* In practice no one will ever be using different schemas (at least per test, but probably never in general), so it's not great that callers would have to duplicate schema every time.

It'd work, but hoping we can come up with something a bit better.

brandur:
@bgentry Thx! I think I got everything then. Want to take one more look at those completer changes?

bgentry (approved):

brandur:
Thx.

PR #852, opened 2025-04-26
Fix intermittent test failures checking job `retryable`

brandur:
We got a lot of the remaining intermittent test failures in #848, but
one that I was punting on because I couldn't immediately see the problem
is this one [1]:

    --- FAIL: Test_Client_JobCompletion (0.00s)
        --- FAIL: Test_Client_JobCompletion/JobThatReturnsErrIsRetryable (1.65s)
            client_test.go:5561: Reusing idle schema "river_2025_04_26t03_40_43_schema_31" after cleaning in 194.808714ms [32 generated] [28 reused]
            logger.go:256: time=2025-04-26T03:40:53.907Z level=INFO msg="River client started" client_id=fv-az1970-529_2025_04_26T03_40_53_541948
            logger.go:256: time=2025-04-26T03:40:53.966Z level=WARN msg="jobexecutor.JobExecutor: Job errored; retrying" error=oops job_id=2 job_kind=callback
            logger.go:256: time=2025-04-26T03:40:54.947Z level=INFO msg="maintenance.JobScheduler: Ran successfully" num_jobs_scheduled=1
            client_test.go:5573:
                    Error Trace:	/home/runner/work/river/river/client_test.go:5573
                    Error:      	Not equal:
                                    expected: "retryable"
                                    actual  : "available"

                                    Diff:
                                    --- Expected
                                    +++ Actual
                                    @@ -1,2 +1,2 @@
                                    -(rivertype.JobState) (len=9) "retryable"
                                    +(rivertype.JobState) (len=9) "available"

                    Test:       	Test_Client_JobCompletion/JobThatReturnsErrIsRetryable
            logger.go:256: time=2025-04-26T03:40:54.997Z level=INFO msg="River client stopped" client_id=fv-az1970-529_2025_04_26T03_40_53_541948
    FAIL
    FAIL	github.com/riverqueue/river	23.010s

This test fails fairly often in CI, but won't fail locally no matter how
many times you run it (or with counts up to 1,000 anyway).

I think I finally see the issue, which is that when checking `retryable`,
by the time we go to look up the job by ID, in slow environments like
GitHub Actions it's possible for the test case goroutine to have been
paused for more than one second, which means that the job may have
already been scheduled to be available again because the first retry
backoff is 1 second. I've even seen failures where the job is actually
running again already because it went to available, then was picked up
by the client to be run again.

Here, fix the problem by defining a special `RetryPolicySlow` retry
policy that retries very slowly and makes this problem impossible. We
don't use this everywhere for things to stay realistic, but configure
the client with it in the failing test case.

[1] https://github.com/riverqueue/river/actions/runs/14677353625/job/41195546754

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -5652,6 +5652,9 @@ func Test_Client_JobCompletion(t *testing.T) {
 		})
 		config.RetryPolicy = &retrypolicytest.RetryPolicyNoJitter{}
 
+		retryPolicy := &retrypolicytest.RetryPolicySlow{} // make sure job isn't set back to available/running by the time we check it below
+		config.RetryPolicy = retryPolicy
+
 		client, bundle := setup(t, config)
 
 		now := client.baseService.Time.StubNowUTC(time.Now().UTC())
@@ -5667,7 +5670,7 @@ func Test_Client_JobCompletion(t *testing.T) {
 		require.NoError(err)
 
 		require.Equal(rivertype.JobStateRetryable, reloadedJob.State)
-		require.WithinDuration(now.Add(1*time.Second), reloadedJob.ScheduledAt, time.Microsecond)
+		require.WithinDuration(now.Add(retryPolicy.Interval()), reloadedJob.ScheduledAt, time.Microsecond)
 		require.Nil(reloadedJob.FinalizedAt)
 	})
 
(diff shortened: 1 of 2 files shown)


bgentry (approved):
So many different ways for slow environments to expose these test glitches 😆 good fix!

brandur:
Yeah I know right? It's crazy.

Thanks!

PR #853, opened 2025-04-26
Make batch completer alter behavior for expedience in `testOnly` mode

brandur:
Here, try another experiment in an attempt to speed up the main
package's test suite in CI. Make the batch completer react to a
`testOnly` flag so that it doesn't wait as long to start completing jobs
(normally ~250ms) for client test cases.


brandur:
Unfortunately only saves about three second in CI tests. Does a little better locally, although still somewhat modest.

brandur:
Closing as "not worth it", although it'd still be nice to find the hot spots for what takes so long in client start/stop to get the main package tests sped up more.

PR #855, opened 2025-04-26
Consolidate `pollForSettingChanges` with other producer subroutines

brandur:
I made some changes to the producer in #848 to fix an intermittent test
introduced by queue pause/unpause by moving a goroutine into the
producer's main start goroutine and making sure to wait on it finishing
before leaving the producer.

I didn't notice at the time that there's already a place where some
other producer subroutines are started, and we can clean things up
somewhat by moving `pollForSettingChanges` down to where the other
subroutines are booted and consolidating the wait group.


bgentry (approved):

brandur:
Thx.

PR #856, opened 2025-04-26
Drop callback args in favor of single, uniform test convention

brandur:
This one's largely a cosmetic change. We remove callback args/worker in
favor of using test case local args using `JobArgsReflectKind` and a
standard invocation to `AddWorker`:

    type JobArgs struct {
        JobArgsReflectKind[JobArgs]
    }

    AddWorker(config.Workers, WorkFunc(func(ctx context.Context, job *Job[JobArgs]) error {
        return JobCancel(errors.New("oops"))
    }))

It's a little more verbose, but has the advantage of using only normal
River primitives, meaning that one has to be familiar with fewer test
helpers to understand what's going on.

Besides that and convention, another small benefit is that it introduces
a little more type safety and makes job args impossible to accidentally
cross reference between test cases. Previously, a test could insert a
callback args while intending to use a callback for them registered in a
different test case. With the new approach, it's very obvious when
`JobArgs` is either forgotten (a compilation error), or when there's a
missing `AddWorker` (because that's supposed to live close to the
definition of `JobArgs`).

### DIFF
--- a/periodic_job_test.go
+++ b/periodic_job_test.go
@@ -42,7 +42,7 @@ func TestPeriodicJobBundle(t *testing.T) {
 			nil,
 		)
 
-		return newPeriodicJobBundle(newTestConfig(t, nil), periodicJobEnqueuer), &testBundle{}
+		return newPeriodicJobBundle(newTestConfig(t, ""), periodicJobEnqueuer), &testBundle{}
 	}
 
 	t.Run("ConstructorFuncGeneratesNewArgsOnEachCall", func(t *testing.T) {
--- a/plugin_test.go
+++ b/plugin_test.go
@@ -29,13 +29,13 @@ func TestClientDriverPlugin(t *testing.T) {
 		t.Helper()
 
 		var (
-			config       = newTestConfig(t, nil)
-			dbPool       = riversharedtest.DBPool(ctx, t)
-			driver       = riverpgxv5.New(dbPool)
-			schema       = riverdbtest.TestSchema(ctx, t, driver, nil)
-			pluginDriver = newDriverWithPlugin(t, dbPool)
+			dbPool = riversharedtest.DBPool(ctx, t)
+			driver = riverpgxv5.New(dbPool)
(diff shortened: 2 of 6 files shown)


periodic_job_test.go
```diff
@@ -42,7 +42,7 @@ func TestPeriodicJobBundle(t *testing.T) {
 			nil,
 		)
 
-		return newPeriodicJobBundle(newTestConfig(t, nil), periodicJobEnqueuer), &testBundle{}
```
brandur, line 45: Made `schema` the second argument to `newTestConfig` instead, but honestly could see it with or without. It does save one LOC for test case where schemas are used (which is a lot of them), but it tends to be empty for test cases using the `runNewTestClient` helper, which injects its own schema.

plugin_test.go
```diff
@@ -98,14 +98,16 @@ func TestClientPilotPlugin(t *testing.T) {
...
-			dbPool       = riversharedtest.DBPool(ctx, t)
-			driver       = riverpgxv5.New(dbPool)
-			schema       = riverdbtest.TestSchema(ctx, t, driver, nil)
+			dbPool = riversharedtest.DBPool(ctx, t)
+			driver = riverpgxv5.New(dbPool)
+			schema = riverdbtest.TestSchema(ctx, t, driver, nil)
+			config = newTestConfig(t, schema)
+		)
+
+		var (
```
bgentry, line 107: I'm seeing a few cases where you have two consecutive var blocks—what's the strategy there? It's more verbose so I'm wanting to understand when this would be beneficial for readability or grouping.

bgentry (approved):

plugin_test.go
brandur: Ah yeah, I think I was just thinking that since dbPool / driver / schema / config are commonly found in a single group, it might make sense for the other two vars to get their own.

I don't feel strongly about it though. I just changed it back to a single `var` block.

brandur:
Thanks!

PR #858, opened 2025-04-26
sqlctemplate: Error on template-like artifacts that are probably mistakes

brandur:
While implementing #848 I made a couple typos in SQL queries at a few
points where I made some minor problem in a template tag like putting in
`/* TEMPLATE schema */` instead of `/* TEMPLATE: schema */`.

sqlctemplate isn't a real compiler, so instead of these problems getting
flagged, they're silently ignored, and this is made even worse because
sqlc strips comments from queries so we often end up with absolutely
nothing to debug with. This results in a long, frustrating debugging
experience.

Here, perform a post-replace match to check for obvious problems and
tell the caller about them immediately. It's not perfect, but will save
time in diagnosing common problems.

### DIFF
--- a/rivershared/sqlctemplate/sqlc_template_test.go
+++ b/rivershared/sqlctemplate/sqlc_template_test.go
@@ -115,6 +115,53 @@ func TestReplacer(t *testing.T) {
 		`, updatedSQL)
 	})
 
+	t.Run("SyntaxMistakes", func(t *testing.T) {
+		t.Parallel()
+
+		replacer, _ := setup(t)
+
+		// Missing colon after "TEMPLATE".
+		{
+			ctx := WithReplacements(ctx, map[string]Replacement{}, nil)
+
+			_, _, err := replacer.RunSafely(ctx, `
+			-- name: JobCountByState :one
+			SELECT count(*)
+			FROM /* TEMPLATE schema */river_job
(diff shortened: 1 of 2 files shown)


rivershared/sqlctemplate/sqlc_template_test.go
```diff
@@ -115,6 +115,53 @@ func TestReplacer(t *testing.T) {
...
+			_, _, err := replacer.RunSafely(ctx, `
+			-- name: JobCountByState :one
+			SELECT count(*)
+			FROM /* TEMPLATE_BEGIN: schema */river_job
+			WHERE state = @state;
+		`, nil)
+			require.EqualError(t, err, "sqlctemplate found template-like tag after replacements; probably syntax error or missing end tag: /* TEMPLATE_BEGIN")
+		}
+
+		// Missing "TEMPLATE_END".
```
bgentry, line 149: wrong comment? This one has `TEMPLATE_END`

bgentry (approved):
LGTM aside from comment issue :v:

rivershared/sqlctemplate/sqlc_template_test.go
brandur: Whoops, that's right. Fixed.

PR #859, opened 2025-04-26
Unskip test to check that listener can take prefix from `search_path`

brandur:
The change in #848 added a capability to the listener so that in case a
schema is _not_ explicitly provided, it can be detected from
`search_path` so that multiple Rivers could share a database as long as
their search paths are configured correctly.

This was tricky to test though because listeners acquire their own
connections, and we don't want to set search path on the database pool
because it'd taint a connection on it and we wouldn't be guaranteed to
get that connection for the listener anyway. I marked the test case as
skipped until I could go and look at it again.

Here, add a function to listeners that causes them to invoke an SQL
command after a connection is acquired. This should only be used in
tests, but lets us check this case. It's a little gnarly to be sure, but
the river driver interface is explicitly marked as unstable, and I can't
think of any alternatives that aren't worse.


bgentry (approved):
Seems perfectly fine for now, thanks!

PR #862, opened 2025-04-26
Fix intermittently failing leader-related tests

brandur:
There's a bunch of clock-based testing happening in the leader tests for
drivers, resulting in intermittent failures like this one [1]:

    --- FAIL: TestDriverRiverPgxV5 (23.18s)
        driver_test.go:73: Reusing idle schema "river_2025_04_26t17_52_50_schema_07" after cleaning in 15.347655ms [7 generated] [6 reused]
        --- FAIL: TestDriverRiverPgxV5/LeaderAttemptElect (0.00s)
            --- FAIL: TestDriverRiverPgxV5/LeaderAttemptElect/ElectsLeader (0.22s)
                riverdrivertest.go:2244: TestTx using schema: river_2025_04_26t17_52_50_schema_06
                riverdrivertest.go:2258:
                            Error Trace:	/home/runner/work/river/river/internal/riverinternaltest/riverdrivertest/riverdrivertest.go:2258
                            Error:      	Max difference between 2025-04-26 17:53:14.813635844 +0000 UTC m=+24.487676847 and 2025-04-26 17:53:14.59636 +0000 UTC allowed is 100ms, but difference was 217.275844ms
                            Test:       	TestDriverRiverPgxV5/LeaderAttemptElect/ElectsLeader
    FAIL
    FAIL	github.com/riverqueue/river	31.602s

As with all these other clock-based failures, the problem is that in
slower environments like GitHub Actions, goroutines may be paused for
extended periods, so by the time they wake up `time.Now()` might be
significantly different, and easily off by a few hundred milliseconds.

Here, make the clock for these functions injectable so we can inject our
own time and get to better test reliability. I'm not necessarily against
other approaches either, but I'm still getting reasonable frequent
failures on these, so I'm definitely in favor of getting some sort of
fix into place.

A side benefit is that it makes clock injection from the client-level a
little more thorough as well.

[1] https://github.com/riverqueue/river/actions/runs/14683687209/job/41209541920

### DIFF
--- a/internal/riverinternaltest/riverdrivertest/riverdrivertest.go
+++ b/internal/riverinternaltest/riverdrivertest/riverdrivertest.go
@@ -2204,37 +2204,6 @@ func Exercise[TTx any](ctx context.Context, t *testing.T,
 
 	const leaderTTL = 10 * time.Second
 
-	t.Run("LeaderDeleteExpired", func(t *testing.T) {
-		t.Parallel()
-
-		exec, _ := setup(ctx, t)
-
-		now := time.Now().UTC()
-
-		{
-			numDeleted, err := exec.LeaderDeleteExpired(ctx, &riverdriver.LeaderDeleteExpiredParams{
-				Schema: "",
-			})
-			require.NoError(t, err)
-			require.Zero(t, numDeleted)
-		}
-
-		_ = testfactory.Leader(ctx, t, exec, &testfactory.LeaderOpts{
-			ElectedAt: ptrutil.Ptr(now.Add(-2 * time.Hour)),
-			ExpiresAt: ptrutil.Ptr(now.Add(-1 * time.Hour)),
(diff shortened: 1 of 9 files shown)


internal/riverinternaltest/riverdrivertest/riverdrivertest.go
```diff
@@ -2204,37 +2204,6 @@ func Exercise[TTx any](ctx context.Context, t *testing.T,
 
 	const leaderTTL = 10 * time.Second
 
-	t.Run("LeaderDeleteExpired", func(t *testing.T) {
```
brandur, line 2207: I just moved this set of tests down so it's in alphabetical order and easier to find.

bgentry (approved):
Good find and fix, thanks 🚀

brandur:
ty.

PR #863, opened 2025-04-26
Add new `HookWorkEnd` interface that runs after workers finish

brandur:
Here, add a new complimentary pair for `HookWorkBegin`: `HookWorkEnd`,
which runs after workers finish, taking in an error result. `HookWorkEnd`
hooks may or may not modify the error result, choosing to suppress an
error on pass it along the stack unchanged.

This is driven by trying to add a new `nilerror` contrib package [1]
that helps detect nil error-compliant structs that return non-nil error
interfaces, which is a common footgun in Go [2].

[1] https://github.com/riverqueue/rivercontrib/pull/25
[2] https://go.dev/doc/faq#nil_error


brandur:
@bgentry I didn't add tests for this one yet. Wanted to get your gut check first in case you disagree with how the error handling works. What do you think?

brandur:
Okay, actually I ended up putting some tests in.

bgentry (approved):

brandur:
ty ty.

PR #878, opened 2025-05-03
Have test signals take a `testutil.TestingTB` for attributable failures

brandur:
A problem with the current test signals system is that when they fail
they panic, which means that in tests you a generic error which can't be
traced back to any particular test [1]:

    panic: test only signal channel is full

    goroutine 55152 [running]:
    github.com/riverqueue/river/rivershared/testsignal.(*TestSignal[...]).Signal(...)
            /home/runner/work/river/river/rivershared/testsignal/test_signal.go:51
    github.com/riverqueue/river/internal/maintenance.(*Reindexer).Start.func1()
            /home/runner/work/river/river/internal/maintenance/reindexer.go:130 +0xd38
    created by github.com/riverqueue/river/internal/maintenance.(*Reindexer).Start in goroutine 55141
            /home/runner/work/river/river/internal/maintenance/reindexer.go:105 +0x1a7
    FAIL	github.com/riverqueue/river	22.678s

This makes the problem ~impossible to track down, especially in the
event the problem is intermittent and can't reproduce with `-parallel 1`
or any other trick.

Here, I propose that when initializing a test signal, its `Init` takes a
`testutil.TestingTB`. On failure, we feed the error into that instead of
panicking with it, thus attributing it to the test case that failed.

I think I was hesitant to do this originally because it feels pretty bad
to reference `*testing.T` or `*testing.B` in non-test code, but given
that `testutil.TestingTB` is a fairly minimal interface that doesn't
import the `testing` package, it seems a little more acceptable to use
this trick.

[1] https://github.com/riverqueue/river/actions/runs/14814948318/job/41594255006

### DIFF
--- a/client.go
+++ b/client.go
@@ -31,6 +31,7 @@ import (
 	"github.com/riverqueue/river/rivershared/testsignal"
 	"github.com/riverqueue/river/rivershared/util/maputil"
 	"github.com/riverqueue/river/rivershared/util/sliceutil"
+	"github.com/riverqueue/river/rivershared/util/testutil"
 	"github.com/riverqueue/river/rivershared/util/valutil"
 	"github.com/riverqueue/river/rivertype"
 )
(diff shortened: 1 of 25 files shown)


bgentry (approved):
Nice improvement ✨

brandur:
ty.

PR #879, opened 2025-05-04
Require standard format for job kinds, mainly to avoid commas and spaces

brandur:
Here, add a client configuration check that requires job kinds adhere to
a standard format, mainly in an attempt to avoid commas and spaces to
make inputs more predictable for an upcoming search UI.

For compatibility purposes, for the time being the check can be disabled
with the new client option `SkipJobKindValidation`, although this option
will be removed in the future.

The character set I've allowed is a little arbitrary (although a lot of
the characters are there to let `JobArgsReflectKind` keep working
properly), so we could extend it to more characters before release if we
get feedback on it.

### DIFF
--- a/client.go
+++ b/client.go
@@ -289,6 +289,15 @@ type Config struct {
 	// setting of Postgres `search_path`.
 	Schema string
 
+	// SkipJobKindValidation causes the job kind format validation check to be
+	// skipped. This is available as an interim stopgap for users that have
+	// invalid job kind names, but would rather disable the check rather than
+	// fix them immediately.
+	//
+	// Deprecated: This option will be removed in a future versions so that job
+	// kinds will always have to have a valid format.
+	SkipJobKindValidation bool
+
 	// SkipUnknownJobCheck is a flag to control whether the client should skip
 	// checking to see if a registered worker exists in the client's worker bundle
 	// for a job arg prior to insertion.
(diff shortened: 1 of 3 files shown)


brandur:
@bgentry One caveat here: I was thinking about it a bit more after we discussed it, and I would bet money that by far the number one character that people are using which is not in the valid character set here is a space, which incidentally is the character we're trying to get rid of. I'm not sure what to do about it, but if we go this route unfortunately it's going to be pretty breaking I think.

We might want to write a compatibility tool to make it renaming job kinds easier. e.g. Support something like `JobArgs.KindAlternate() string` that lets you install a worker with a second kind name.

client.go
```diff
@@ -289,6 +289,15 @@ type Config struct {
...
 	Schema string
 
+	// SkipJobKindValidation causes the job kind format validation check to be
+	// skipped. This is available as an interim stopgap for users that have
+	// invalid job kind names, but would rather disable the check rather than
+	// fix them immediately.
+	//
+	// Deprecated: This option will be removed in a future versions so that job
+	// kinds will always have to have a valid format.
+	SkipJobKindValidation bool
```
bgentry, line 299: So this is more of an opt-out than an opt-in. Maybe it will be fine though? Really hard to predict how people are using this IRL.

bgentry (approved):
Thanks for putting this together!

client.go
brandur: Yeah, I'd originally started with an opt-in, but two things occurred to me while doing that:

* Adding/removing one of these configs is a breaking change, and with an opt-in, we'd probably _also_ want an opt-out later, so: (1) add opt-in, (2) make validation required + add opt-out, (3) deprecate opt-in option, (4) deprecate opt-out option. So that's two breaking changes instead of one.
* Very likely, few people would notice the new opt-in option and actually use it.

So I'm pretty sure the net effect would be more churn and not much pain saved from the extra opt-in, so just decided to jump to an opt-out instead.

brandur:
Thanks!

PR #880, opened 2025-05-04
Add `JobArgsWithKindAliases`, providing an easy/safe way of renaming kinds

brandur:
Here, add `JobArgsWithKindAliases`, a job args interface that lets job
args implement a second `KindAliases` that returns alternate kinds
that its worker will respond to. This provides an easy and safe way of
renaming jobs even when existing jobs under the original name are still
in the database.

Renaming a job is a three part process. To begin, a job args with its
original name:

    type jobArgsBeingRenamed struct{}

    func (a jobArgsBeingRenamed) Kind() string { return "old_name" }

Rename by putting the new name in `Kind` and moving the old name to
`KindAliases`:

    type jobArgsBeingRenamed struct{}

    func (a jobArgsBeingRenamed) Kind() string          { return "new_name" }
    func (a jobArgsBeingRenamed) KindAliases() []string { return "old_name_still_recognized" }

After all jobs inserted under the original name have finished working
(including all their possible retries, which notably might take up to
three weeks on the default retry policy), remove `KindAliases`:

    type jobArgsBeingRenamed struct{}

    func (a jobArgsBeingRenamed) Kind() string { return "new_name" }

This addition is largely driven by #879, which may make some users kinds
invalid if they contain banned characters like spaces or commas. Without
this change, users wouldn't have a particularly easy way of renaming
kinds that aren't compliant, and would have to resort to something like
fully duplicating a job args and worker and doing a find/replace.

### DIFF
--- a/job.go
+++ b/job.go
@@ -24,6 +24,39 @@ type JobArgs interface {
 	Kind() string
 }
 
+// JobArgsWithKindAliases  is an interface that jobs args can implement to
+// provide an alternate kind which a worker will be registered under in addition
+// to the primary kind. This is useful for renaming a job kind in a safe manner
+// so that any jobs already in the database aren't orphaned.
+//
+// Renaming a job is a three part process. To begin, a job args with its
+// original name:
+//
+//	type jobArgsBeingRenamed struct{}
+//
+//	func (a jobArgsBeingRenamed) Kind() string { return "old_name" }
+//
+// Rename by putting the new name in Kind and moving the old name to
+// KindAliases:
+//
+//	type jobArgsBeingRenamed struct{}
+//
+//	func (a jobArgsBeingRenamed) Kind() string          { return "new_name" }
+//	func (a jobArgsBeingRenamed) KindAliases() []string { return []string{"old_name_still_recognized"} }
+//
+// After all jobs inserted under the original name have finished working
+// (including all their possible retries, which notably might take up to three
+// weeks on the default retry policy), remove KindAliases:
+//
+//	type jobArgsBeingRenamed struct{}
+//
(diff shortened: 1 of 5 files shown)


brandur:
@bgentry Any immediate reactions on this one? I wouldn't mind shipping a new release so we could try putting the full lib/pq support in front of people, but since #879 would go out at the same time, it wouldn't be the worst idea to make this available as a migration aid.

job.go
```diff
@@ -24,6 +24,37 @@ type JobArgs interface {
...
+//
+// After all jobs inserted under the original name have finished working
+// (including all their possible retries, which notably might take up to three
+// weeks on the default retry policy), remove KindAlternate:
+//
+//	type jobArgsBeingRenamed struct{}
+//
+//	func (a jobArgsBeingRenamed) Kind() string { return "new_name" }
+type JobArgsWithKindAlternate interface {
+	KindAlternate() string
```
bgentry, line 55: Thoughts on not limiting this to a single alternate name/alias? I'm imagining a use case where multiple old job types are consolidated into one new one—would be great to handle more flexible situations like that. `KindAliases() []string` for example.

brandur: Yeah I thought about it, but TBH, I couldn't think of any particularly compelling cases where something like that would be needed.

That said, I don't feel particularly strongly about it. Updated the PR to add `KindAliases`.

bgentry (approved):
Looks great! We can probably add a docs section to one of our pages for this, though I'm not quite sure which one makes the most sense 🤔

brandur:
Great, thx.

RE docs: Yeah, I'm not sure yet either, but I was kind of thinking there might be a dedicated "renaming jobs" how-to or something of that nature. It's a big enough/non-obvious enough subject that there's a decent amount of content.

Issue #545, opened 2024-08-21
Elector interval field value out of range


gottafixthat:
I was evaluating riverqueue for use in one of my projects and after attempting to run it for the first time, I'm getting the following error:

>time=2024-08-21T19:45:12.744Z level=ERROR msg="Elector: Error attempting to elect" client_id=5e589f722171_2024_08_21T19_45_11_805369 err="pq: interval field value out of range: \"15000000000\"" num_errors=2 sleep_duration=1.917471608s

Postgres logs:

> 2024-08-21 19:07:10.854 UTC [70] ERROR:  interval field value out of range: "15000000000"
>  2024-08-21 19:07:10.854 UTC [70] CONTEXT:  unnamed portal parameter $2 = '...'
>  2024-08-21 19:07:10.854 UTC [70] STATEMENT:  -- name: LeaderAttemptElect :execrows
>        INSERT INTO river_leader(leader_id, elected_at, expires_at)
>            VALUES ($1, now(), now() + $2::interval)
>        ON CONFLICT (name)
>            DO NOTHING

I have validated that my migrations worked correctly.

Postgres v14.6 (I'm locked to this version at the moment)
Go v1.23.0
River v0.11.4

Any suggestions?

bgentry:
Seems likely you’re using the dbsql driver? This appears to be a bug with that driver not handling duration types as db interval types correctly.

gottafixthat:
I tried both, actually. I thought it might have been the dbsql driver at first so I refactored to use riverpgxv5 and it was the same result.

bgentry:
Hmm, you must have at least gotten a different error when using pgx? The message above is specific to pq which isn’t used at all by the pgx driver.

gottafixthat:
I'll verify this tomorrow morning. It's entirely possible that I'm mixing things in the wrong way.

brandur:
It occurred to me while looking at this that 15000000000 is the elector's default 15 second interval in nanoseconds, so 15 * 1_000_000_000.

You can see here that when pgx encodes a duration, it encodes it as microseconds, which is the maximum precision that Postgres will accept:

https://github.com/jackc/pgx/blob/4f7e19d67df4d411ac1c19373390021a2f23aa07/pgtype/builtin_wrappers.go#L498-L514

For some reason your pgx is trying to write _nanoseconds_ instead of microseconds. Do you have a custom encoding override somewhere?

gottafixthat:
[...]

PR #882, opened 2025-05-05
[fix] Support time.Duration as interval in database/sql driver compatibility

nickgala:
We started implementing **River** in our services, which use `database/sql`, and came across this issue: https://github.com/riverqueue/river/issues/545.

Since migrating to `pgx` will take some time, I looked into the issue. Instead of updating the SQL queries generated by sqlc (e.g., using `MAKE_INTERVAL` or modifying the `go_type`), which would also affect `riverpgxv5`, I applied a focused quick-fix for the database/sql driver only.

The issue arises because `database/sql` in Go doesn't know how to serialize `time.Duration` into a format that PostgreSQL accepts as an `INTERVAL`.

The fix I’m applying casts `time.Duration` to a string (e.g., `"60 seconds"`), which PostgreSQL can safely parse as an interval.

### DIFF
--- a/riverdriver/riverdatabasesql/river_database_sql_driver.go
+++ b/riverdriver/riverdatabasesql/river_database_sql_driver.go
@@ -981,6 +981,7 @@ type templateReplaceWrapper struct {
 
 func (w templateReplaceWrapper) ExecContext(ctx context.Context, sql string, args ...interface{}) (sql.Result, error) {
 	sql, args = w.replacer.Run(ctx, sql, args)
+	convertDurationToInterval(sql, args)
 	return w.dbtx.ExecContext(ctx, sql, args...)
 }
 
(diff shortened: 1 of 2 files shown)


brandur:
Thanks @nickgala.

So I was just testing this locally, and although the durations are one aspect that doesn't work with raw lib/pq, there are a bunch of other pieces too that also don't work, but don't manifest quite as early.

I think what we need to do here is get full test coverage on top of pure lib/pq, which will also prevent future regressions. I have a branch that mostly accomplishes this now, and just need to fix a couple more things and we should be good to go.

brandur:
Opened: https://github.com/riverqueue/river/pull/883

brandur:
Check out v0.22.0 which should fix this problem.

https://github.com/riverqueue/river/releases/tag/v0.22.0

Issue #566, opened 2024-08-26
Leader election fails with pq driver


wydengyre:
A repo demonstrating the issue is available here: https://github.com/wydengyre/dbsqlintervaldemo

Using `riverdatabasesql` with the `pq` driver will result in failure of the leader election system.

The leader election SQL will result in `expires_at` being set to a time in the year 2183: https://github.com/FiveBellsSettlement/river/blob/62ce779744c2d2434cc41eb02a6f19db419dc79e/riverdriver/riverdatabasesql/internal/dbsqlc/river_leader.sql.go#L13

brandur:
Yeah, we've suspect that River isn't compatible with lib/pq. `riverdatabasesql` is mainly meant for compatibility through a `database/sql` interface that's using an underlying Pgx driver. I might try to fix it depending on time, but given lib/pq's ancient and been in maintenance mode only for years, I'm not sure how soon I'll get to it.

OOC, what's the main thing blocking you from migrating to a more modern driver like Pgx?

brandur:
Check out v0.22.0 which should fix this problem.

https://github.com/riverqueue/river/releases/tag/v0.22.0

PR #883, opened 2025-05-06
Driver tests for full `lib/pq` connection + necessary fixes

### LINKED ISSUE #545: Elector interval field value out of range

brandur:
So it turns out that while bearing some relation to `lib/pq` through
various pieces that sqlc pulls in, the `riverdatabasesql` driver didn't
actually work when used with `lib/pq`. When testing it, we'd open a
`sql.DB` pool _via_ Pgx, so Pgx was doing all the heavy lifting.

Here, add an alternate driver test strategy to test `riverdatabasesql`
with a `lib/pq` pool, thereby ensuring that the full range of operations
works even with no Pgx in the mix at all.

Things were broken, so we get those fixed up. There we two main
problems:

* As described in #545, `lib/pq` doesn't handle intervals correctly, and
  reads them in as ints, which is a big problem because they're in
  nanoseconds. We work around this by injecting intervals (like leader
  TTLs) an second floats instead, which really doesn't feel that much
  worse.

* The custom bits class for `unique_states` didn't work. To fix this I
  move to an alternate strategy of inserts `unique_states` as an
  integer, a technique that I've been using in SQLite because there's no
  `bits` type. I could make a strong argument that the use of an integer
  over `bits` would've been a better design from the get go anyway,
  because it makes all the bit-wise arithmetic much simpler as you use
  C-style conventions like `states & (1 << 3)` that are very widespread
  and well understood. Either way though, things work reasonably well if
  we insert as an integer, then transition to storage in bits.

Fixes #545 and #566. Supersedes #882.

### DIFF
--- a/riverdriver/riverdatabasesql/river_database_sql_driver.go
+++ b/riverdriver/riverdatabasesql/river_database_sql_driver.go
@@ -19,13 +19,12 @@ import (
 	"strings"
 	"time"
 
-	"github.com/jackc/pgx/v5/pgtype"
-
 	"github.com/riverqueue/river/internal/dbunique"
 	"github.com/riverqueue/river/riverdriver"
 	"github.com/riverqueue/river/riverdriver/riverdatabasesql/internal/dbsqlc"
 	"github.com/riverqueue/river/riverdriver/riverdatabasesql/internal/pgtypealias"
 	"github.com/riverqueue/river/rivershared/sqlctemplate"
+	"github.com/riverqueue/river/rivershared/util/ptrutil"
 	"github.com/riverqueue/river/rivershared/util/sliceutil"
 	"github.com/riverqueue/river/rivertype"
 )
@@ -236,7 +235,7 @@ func (e *Executor) JobInsertFastMany(ctx context.Context, params *riverdriver.Jo
 		State:        make([]string, len(params.Jobs)),
 		Tags:         make([]string, len(params.Jobs)),
 		UniqueKey:    make([]pgtypealias.NullBytea, len(params.Jobs)),
-		UniqueStates: make([]pgtypealias.Bits, len(params.Jobs)),
+		UniqueStates: make([]int32, len(params.Jobs)),
 	}
 	now := time.Now().UTC()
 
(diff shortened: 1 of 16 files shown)


brandur:
@bgentry Mind taking a look at this one?

riverdriver/riverpgxv5/internal/dbsqlc/river_leader.sql
```diff
@@ -9,16 +9,17 @@ CREATE UNLOGGED TABLE river_leader(
 
 -- name: LeaderAttemptElect :execrows
 INSERT INTO /* TEMPLATE: schema */river_leader (leader_id, elected_at, expires_at)
-    VALUES (@leader_id, coalesce(sqlc.narg('now')::timestamptz, now()), coalesce(sqlc.narg('now')::timestamptz, now()) + @ttl::interval)
+    -- @ttl is inserted as
```
bgentry, line 12: not sure what this comment is meant to convey

riverdriver/riverdatabasesql/river_database_sql_driver.go
```diff
@@ -999,6 +994,18 @@ func (w templateReplaceWrapper) QueryRowContext(ctx context.Context, sql string,
 	return w.dbtx.QueryRowContext(ctx, sql, args...)
 }
 
+// lib/pq reads in `bits` (like `bits(8)`) as a bit string that's a normal
+// decimal integer like 1111_1111 (_not_ 0b1111_1111). This function converts it
+// back to binary representation. See the test suite for examples.
```
bgentry, line 999: pq 🤦‍♂️

bgentry (approved):
It's always a bit unfortunate to have to move more in the direction of the lowest common denominator between drivers, but it's tolerable here. In the event we ever need to add another job state it might we'll need more than 8 bits and will need to adjust the approach.

philippgille:
Does this also fix https://github.com/riverqueue/river/issues/566?

bgentry:
@philippgille it should fix it! 🤞

riverdriver/riverpgxv5/internal/dbsqlc/river_leader.sql
brandur: Ah doh, I started a comment, but didn't finish it. I just put this more expounded version in instead:

``` sql
-- @ttl is inserted as as seconds rather than a duration because `lib/pq` doesn't support the latter
```

riverdriver/riverdatabasesql/river_database_sql_driver.go
brandur: Yeah haha. There might be some other way around this, but at least this isn't _that_ bad.

brandur:
> It's always a bit unfortunate to have to move more in the direction of the lowest common denominator between drivers, but it's tolerable here.

Yeah for sure.  I was reasonably happy that that none of the hacks to get this working were that gross. Code's different, but not obvious that much worse.

> In the event we ever need to add another job state it might we'll need more than 8 bits and will need to adjust the approach.

Yeah, although most of the integers we're working with are at least four bytes, so they should still mostly work in case we need more than those 8 bits.

brandur:
Thanks!

PR #884, opened 2025-05-08
Finish accidentally not pushed `@ttl` comment

brandur:
This is just a dumb one where I was addressing code review feedback from
the `lib/pq` change #883, specifically a half finished comment on
`@ttl`, then forgot to push the change up to my branch before merging
it. Here, address the problem by finishing the comment.


bgentry (approved):

brandur:
Thx. Had an intermittent test failure. I'll fix that separately.

PR #892, opened 2025-05-10
Additional docs for `JobArgs.Kind`

brandur:
Follow up #879 to add a little more useful context to `JobArgs.Kind`:

* Explain why `Kind` is a string rather than based on type names.
* Mention that special characters like spaces and commas are not
  allowed. Suggest examples on what _is_ allowed.
* Mention how changing a kind generally isn't safe unless the database
  is completely empty and that `JobArgsWithKindAliases` can be used to
  rename jobs safely.


bgentry (approved):
Great additions.

brandur:
ty ty.

PR #896, opened 2025-05-14
Suppress producer error log on shutdown

brandur:
I noticed when running some benchmarking tonight that I'd occasionally
get an error line at the end even though things seem to finish cleanly:

    May 13 18:13:36.622 ERR producer: Error fetching queue settings err="context canceled"

Looking at the implementation, it looks like this is something that
could happen randomly in the poll-only path. If the producer happens to
be polling for queue changes as shutdown happens, it'll log an error as
the context is cancelled. This is more likely to happen for SQLite
because we constrain the connection pool to only one active connection.

Here, only produce an error log conditionally in cases where the parent
context has not been cancelled.

### DIFF
--- a/producer.go
+++ b/producer.go
@@ -797,9 +797,21 @@ func (p *producer) pollForSettingChanges(ctx context.Context, wg *sync.WaitGroup
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			updatedQueue, err := p.fetchQueueSettings(ctx)
+			updatedQueue, err := func() (*rivertype.Queue, error) {
+				ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+				defer cancel()
+
+				return p.exec.QueueGet(ctx, &riverdriver.QueueGetParams{
+					Name:   p.config.Queue,
+					Schema: p.config.Schema,
+				})
+			}()
 			if err != nil {
-				p.Logger.ErrorContext(ctx, p.Name+": Error fetching queue settings", slog.String("err", err.Error()))
+				// Don't log if this is part of a standard shutdown.
+				if !errors.Is(context.Cause(ctx), startstop.ErrStop) {
+					p.Logger.ErrorContext(ctx, p.Name+": Error fetching queue settings", slog.String("err", err.Error()))
+
+				}
 				continue
 			}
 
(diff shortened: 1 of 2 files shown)


producer.go
```diff
@@ -797,9 +797,20 @@ func (p *producer) pollForSettingChanges(ctx context.Context, wg *sync.WaitGroup
...
-			updatedQueue, err := p.fetchQueueSettings(ctx)
+			updatedQueue, err := func() (*rivertype.Queue, error) {
+				ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+				defer cancel()
+
+				return p.exec.QueueGet(ctx, &riverdriver.QueueGetParams{
+					Name:   p.config.Queue,
+					Schema: p.config.Schema,
+				})
+			}()
```
brandur, line 808: @bgentry Are you okay on inlining some of these trivial functions like this? IMO makes it easier to see what it's going to do without having to go search for the helper's implementation elsewhere in the file.

producer.go
```diff
@@ -797,9 +797,20 @@ func (p *producer) pollForSettingChanges(ctx context.Context, wg *sync.WaitGroup
...
 			if err != nil {
-				p.Logger.ErrorContext(ctx, p.Name+": Error fetching queue settings", slog.String("err", err.Error()))
+				// Don't log if this is part of a standard shutdown.
+				if ctx.Err() == nil {
```
bgentry, line 811: hmm, what if the error is because of a context timeout or cancellation from the above `context.WithTimeout`? We would probably want to log about that. Could leverage `context.WithTImeoutCause` up there to get a specific identifiable error in that case?

bgentry (approved):
Mostly LGTM but I think it needs one tweak on the `if ctx.Err() == nil` bit to make sure it doesn't ignore timeouts that our own function adds.

producer.go
brandur: So in this case the `ctx` with timeout added still gets scoped to the anonymous function — so even if the timeout triggers, the `ctx` that timed out is gone by the time the functions returns. On this line `ctx` is back to the one injected into `pollForSettingChanges`.

That said, you reminded me that we actually already have `ErrStop` for this type of circumstance, and we already use it for checks elsewhere in `producer.go` I changed the line over to this:

``` go
			if err != nil {
				// Don't log if this is part of a standard shutdown.
				if !errors.Is(context.Cause(ctx), startstop.ErrStop) {
```

Seems more obvious.

brandur:
Thanks!

PR #900, opened 2025-05-15
Add unique states bitmask helpers to `rivershared/uniquestates`

brandur:
This one's aimed at a bug discovered in Pro wherein because the project
didn't have access to the unique states bitmask helpers, the
`UniqueStates` property was left off job rows. Here, put the headers in
a general location in `rivershared` that can be accessed from both
projects.


bgentry (approved):

brandur:
thx.

PR #903, opened 2025-05-17
Add `--schema` option to `river migrate-get` + hide schema comments otherwise

### LINKED ISSUE #902: SQL migrations don't do schema template substitution
Thanks for your work, this is a really cool project.

When following the steps to add sql migrations [here](https://riverqueue.com/docs/migrations#exporting-sql-for-use-in-other-frameworks), I noticed the output is invalid syntax, the `/* TEMPLATE: schema */` substitution is not done. It' easy enough to find and replace, but maybe a cli flag to specify the schema would be nice, or even just having it default to public for the cli output. Seems like it was introduced in https://github.com/riverqueue/river/releases/tag/v0.21.0 with https://github.com/riverqueue/river/pull/848

brandur:
Here, add a `--schema` option to the `river migrate-get` command that
lets the caller inject a custom schema into the SQL generated by the
command. If the `--schema` command is not provided, we hide the
`/* TEMPLATE: schema */` comments for cleanliness (previously, they were
returned, which wasn't harmful but added visual noise).

Since we're making changes anyway, add a test suite for `migrate-get`.
It should've been there before anyway and it will help protect against
regressions.

Only tangentially related, fix a small bug in the migration "flags" that
are output like `-- River main migration 001 [up]` in that when using
the default "main" line, the line was being accidentally omitted.

Also unrelated, but add missing docs for `river migrate-list` which have
apparently accidentally been a TODO for quite some time.

Fixes #902.

### DIFF
--- a/producer.go
+++ b/producer.go
@@ -810,7 +810,6 @@ func (p *producer) pollForSettingChanges(ctx context.Context, wg *sync.WaitGroup
 				// Don't log if this is part of a standard shutdown.
 				if !errors.Is(context.Cause(ctx), startstop.ErrStop) {
 					p.Logger.ErrorContext(ctx, p.Name+": Error fetching queue settings", slog.String("err", err.Error()))
-
 				}
 				continue
 			}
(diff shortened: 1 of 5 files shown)


producer.go
```diff
@@ -810,7 +810,6 @@ func (p *producer) pollForSettingChanges(ctx context.Context, wg *sync.WaitGroup
 				// Don't log if this is part of a standard shutdown.
 				if !errors.Is(context.Cause(ctx), startstop.ErrStop) {
 					p.Logger.ErrorContext(ctx, p.Name+": Error fetching queue settings", slog.String("err", err.Error()))
-
```
brandur, line 813: Not sure what happened here. My linter is suddenly picking these up, but it seems like it should've picked them up before. I left the changes in though because they look right and I'm sure the linter will want to put them in another PR if not this one.

brandur:
@bgentry Seem okay?

bgentry (approved):

brandur:
thx.

PR #912, opened 2025-05-21
Remove `TimeStub` from `rivertest`

brandur:
I was looking through the public API surface area of `rivertest` [1] and
noticed that we actually have a `TimeStub` in there. I think this was
accidental export from a time before we had a `TimeStub` available in
`rivershared`.

I think we should take this out. Technically it's a breaking change
given it was part of the public API, but there wasn't a whole bunch that
anyone would need this for, so I doubt usage is significant, and
hopefully no one's using it at all.

[1] https://pkg.go.dev/github.com/riverqueue/river/rivertest


bgentry (approved):
Good find and agree, nobody should be using this anyway so hopefully it won't impact anybody.

PR #926, opened 2025-05-24
Allow explicit schema injection to `rivertest.Require*` test functions

### LINKED ISSUE #907: rivertest.RequireX for alternative schema
We have our River client use an alternative schema: "workers". And we go through great lengths to make our tests mirror the actual database setup so we can't really change this. But now that i've started working with the rivertest package this doesn't seem to be supported. I simply get:

```
    River assertion failure:
            Internal failure: error querying jobs: ERROR: relation "river_job" does not exist (SQLSTATE 42P01)
```

This seems like it is a known limitations because of the comment i've found here: https://github.com/riverqueue/river/blob/487814e920948f7b8cce26ba37278e5f7865b2fc/rivertest/rivertest.go#L23 

Is there any workaround?

brandur:
Here, resolve #907 by letting an explicit schema be injected into
`rivertest.Require*` assertions in a similar way that one can be used in
a client.

This approach adds a schema in `RequireInsertedOpts`. This comment does
a good job of highlight all the potential approaches for adding a schema
[1], and unfortunately none of them are all that great. I implemented
one other version of this (a variant of option 2 in that list), which as
some advantages, but in the end it just ended up ballooning the API out
to an uncomfortable degree.

The worst part about adding schema to `RequireInsertedOpts` is its
interact with the `RequireMany*` functions, where each expectation can
set its own schema, and it's not clear what would happen if different
expectations set different schemas. I resolved this ambiguity by making
it an error to mix and match schemas. Assertions are allowed to send a
schema in only the first position like:

    jobs := requireManyInserted(ctx, bundle.mockT, bundle.driver, []ExpectedJob{
        {Args: &Job1Args{String: "foo"}, Opts: bundle.schemaOpts},
        {Args: &Job1Args{String: "bar"}},
    })

Or send the same schema in all positions:

    jobs := requireManyInserted(ctx, bundle.mockT, bundle.driver, []ExpectedJob{
        {Args: &Job1Args{String: "foo"}, Opts: bundle.schemaOpts},
        {Args: &Job1Args{String: "bar"}, Opts: bundle.schemaOpts},
    })

But they aren't allowed to set a schema only in position other than the
first, or mix and match schemas between expectations.

Fixes #907.

[1] https://github.com/riverqueue/river/issues/907#issuecomment-2896783495

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Preliminary River driver for SQLite (`riverdriver/riversqlite`). This driver seems to produce good results as judged by the test suite, but so far has minimal real world vetting. Try it and let us know how it works out. [PR #870](https://github.com/riverqueue/river/pull/870).
 - CLI `river migrate-get` now takes a `--schema` option to inject a custom schema into dumped migrations and schema comments are hidden if `--schema` option isn't provided. [PR #903](https://github.com/riverqueue/river/pull/903).
 - Added `riverlog.NewMiddlewareCustomContext` that makes the use of `riverlog` job-persisted logging possible with non-slog loggers. [PR #919](https://github.com/riverqueue/river/pull/919).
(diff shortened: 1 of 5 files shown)


brandur:
+1, wish there was a better option, but this should do it for now. Thanks!

PR #928, opened 2025-05-25
Out of transaction notification on job insert and queue changes

brandur:
Here, we try to implement a system that'll make drivers that don't
support listen/notify like SQLite more responsive by default (i.e.
without having to crank their fetch intervals down to miniscule numbers
and have them fetch with hyper-frequency) so they respond to a new job
or queue being paused quickly.

Currently, drivers with listeners send out notifications which are then
received by the producer so that it can enact work. Drivers without
listeners are left out in the cold though because they don't have any
good listen/notify alternative.

This proposal takes non-tx functions like `Client.Insert` or
`Client.QueuePause` and makes a special call to the producer to notify
it of what happened so it can respond quickly without waiting for a
poll.

The approach of course has a couple limitations:

* It only works for non-tx functions. If we were to notify the producer
  within a transaction, the producer would wake up and find nothing to
  do because the relevant data hasn't been persisted yet. This is taken
  care of on the listener drivers because listen/notify is transaction
  friendly.

* Only works within the same process. A client obviously can't notify a
  producer in a different Go process, so those will still have to poll.
  Still, if even one producer can get a notification about an inserted
  job immediately, that job still gets worked faster.

    This improvement is also aimed pretty specifically aimed at our test
    suite, where there's only ever one process running. Being able to
    notify a producer within the same process means that we don't need
    to put an aggressive poll loop into ever example or client test that
    uses SQLite.

Longer term, I'd like to replace this with something else, like a
general purpose unlogged table that could process notifications in a
transaction in a similar way to how listen/notify does it (although I'd
have to do a little more work on this, I'm not sure it'd work), but I
figure that this might be good enough for the time being. This doesn't
change the public API at all, so it should be fairly easy to replace all
changes made here with something more suitable if it comes up.

### DIFF
--- a/client.go
+++ b/client.go
@@ -23,6 +23,7 @@ import (
 	"github.com/riverqueue/river/internal/notifier"
 	"github.com/riverqueue/river/internal/notifylimiter"
 	"github.com/riverqueue/river/internal/rivercommon"
+	"github.com/riverqueue/river/internal/util/dbutil"
 	"github.com/riverqueue/river/internal/workunit"
 	"github.com/riverqueue/river/riverdriver"
 	"github.com/riverqueue/river/rivershared/baseservice"
@@ -823,7 +824,10 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		{
 			periodicJobEnqueuer := maintenance.NewPeriodicJobEnqueuer(archetype, &maintenance.PeriodicJobEnqueuerConfig{
 				AdvisoryLockPrefix: config.AdvisoryLockPrefix,
-				Insert:             client.insertMany,
+				Insert: func(ctx context.Context, execTx riverdriver.ExecutorTx, insertParams []*rivertype.JobInsertParams) error {
+					_, err := client.insertMany(ctx, execTx, insertParams)
+					return err
+				},
 			}, driver.GetExecutor())
 			maintenanceServices = append(maintenanceServices, periodicJobEnqueuer)
 			client.testSignals.periodicJobEnqueuer = &periodicJobEnqueuer.TestSignals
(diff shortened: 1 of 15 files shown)


example_sqlite_test.go
```diff
@@ -36,9 +35,7 @@ func Example_sqlite() {
 	river.AddWorker(workers, &SortWorker{})
 
 	riverClient, err := river.NewClient(driver, &river.Config{
-		FetchCooldown:     20 * time.Millisecond,
-		FetchPollInterval: 50 * time.Millisecond, // this driver is poll only, so speed up poll interval so the test runs fater
```
brandur, line 40: The most immediate advantage of this change for the test suite is that when testing SQLite, we don't make the tests super slow if we accidentally omit configuration for increased polling frequency. It can now be safely removed and we can be sure the producer will pick changes up right away.

client.go
```diff
@@ -1756,6 +1778,25 @@ func (c *Client[TTx]) insertManyParams(params []InsertManyParams) ([]*rivertype.
 	return insertParams, nil
 }
 
+// Notifies an internal producer of new jobs being queued for work.  Only
+// invoked if the client's driver doesn't support a listener. If a listener is
+// supported, job notifications go out via listen/notify instead.
+//
+// Should only ever be invoke *outside* a transaction. If invoked within a
```
```suggestion
// Should only ever be invoked *outside* a transaction. If invoked within a
```

bgentry (approved):
Sorry it took me so long to get to this one! I was kinda expecting it'd be a full solution to enabling events on non listen/notify drivers, but this is an incremental improvement anyway. We should probably still chase down a solution to that other problem for dbsql users to get them up to parity.

PR #932, opened 2025-05-26
Don't log context canceled error on shutdown during benchmark

brandur:
I just noticed while running `river bench` with a lot of jobs that when
leaving after finishing up, we'll get errors like these ones:

    May 26 12:25:33.612 ERR Error counting jobs err="context canceled"
    May 26 12:25:33.612 ERR Error counting jobs err="context canceled"

Job counting takes longer due to job volume, so when shutting down, the
count goroutine is more likely to be waiting on a count than sleeping in
the 1 second tick interval, thereby producing the error.

Here, ignore `context.Canceled` so that it's no logged and return as
finished instead.


bgentry (approved):

brandur:
ty.

PR #934, opened 2025-05-28
Add pilot hooks for periodic jobs

brandur:
Add pilot hooks for periodic jobs, thereby giving River plugins the
capability to extend certain parts of the periodic job enqueuer.


brandur:
thx.

PR #935, opened 2025-05-28
Reindexer: Skip work if artifacts from a failed reindex are present + increase timeout

### LINKED ISSUE #909: Reindex timeouts are causing leftover indexes
After investigating job insertion slowness, we discovered a lot of failed index builds.

Image

Assumably those a leftovers of index rebuild timeouts, e.g. we see this in our logs from time to time

Image

According to the the postgres docs https://www.postgresql.org/docs/current/sql-reindex.html#SQL-REINDEX-CONCURRENTLY, those indexes affect inserting performance.

> If a problem arises while rebuilding the indexes, such as a uniqueness violation in a unique index, the REINDEX command will fail but leave behind an “invalid” new index in addition to the pre-existing one. This index will be ignored for querying purposes because it might be incomplete; however it will still consume update overhead.

Would it be possible for river to automatically clean failed reindex attempts?

brandur:
Here, we attempt to resolve #909, in which a user had found a great
number of invalid indexes in their database, caused by the reindex
running, having its context cancelled on timeout, and leaving an invalid
index behind. 24 hours later it would try again, with the same result.

We add a check to the reindexer to look for artifacts of previously
failed reindex operations (or concurrently running reindex operations,
though that shouldn't happen due to the leadership system), and if
found, skip reindexing in favor of logging a warning about what
happened.

As documented by Postgres [1], a concurrent reindex creates a new index
with a prefix of `*_ccnew`, and swaps it into place if reindexing
succeeds. The existing index is renamed to `*_ccold` and dropped
concurrently. We look for both these suffixes, and skip reindexing if we
find either of them.

I opted not to clean reindexing artifacts for the same reason that we're
skipping reindexing if finding them in the first place: even after
cleaning them out, reindexing is likely to fail again. If it failed once
at a given timeout threshold, then it's probably fated to fail again.

When running and failing multiple times, reindexing will add numbers to
each new attempt like `*_ccnew1`, `*_ccnew2`, ... `*_ccnew10`,
`*_ccnew11`. I opted not to look for these because if any of them
exists, we can generally assume that the first `*_ccnew` does too. If
not (say a user manually dropped it), the reindexing will occur, fail
again, and `*_ccnew` will be there. I don't think we have to worry about
this edge too much.

Lastly, I also increase the reindex context timeout from 15 seconds to 1
minute. As seen in #909, a large table will have reindex builds that
take at least 45 seconds, so 1 minute is that plus some margin. Use of
`CONCURRENTLY` should ensure safe continued table access even while a
longer reindex is running. Even 1 minute might not be long enough for
every case, but it's a lot, and should reduce the number of users who
see failures dramatically (from few users to very, very few users). The
choice of 15 seconds originally was quite arbitrary and there's no
reason to be married to it.

Fixes #909.

[1] https://www.postgresql.org/docs/current/sql-reindex.html#SQL-REINDEX-CONCURRENTLY


internal/maintenance/reindexer.go
```diff
@@ -158,16 +166,47 @@ func (s *Reindexer) Start(ctx context.Context) error {
...
 
-func (s *Reindexer) reindexOne(ctx context.Context, indexName string) error {
-	ctx, cancel := context.WithTimeout(ctx, s.Config.Timeout)
-	defer cancel()
+func (s *Reindexer) reindexOne(ctx context.Context, indexName string) (bool, error) {
+	var cancel func()
+	if s.Config.Timeout > -1 {
+		ctx, cancel = context.WithTimeout(ctx, s.Config.Timeout)
+		defer cancel()
+	}
```
bgentry, line 174: Thinking about this more, it seems likely that even if a long enough timeout is used, an ill-timed shutdown will cancel the context and leave an incomplete index which will block future attempts. AFAICT we have two options:

1. Detach this context from the parent's cancellation with [`WithoutCancel`](https://pkg.go.dev/context#WithoutCancel) so that this operation can continue until it completes. This runs into another issue though which is that our maintenance process model isn't designed to consider long-running operations that shouldn't be interrupted. Similar to the `Client` we'd need to consider separating the "stop initiating new work but take your time and finish ongiong work before shutting down" from the "cancel all work contexts and shut down as fast as possible" case—IF we still want the reindexer to be responsive to shutdown requests.
2. Go down the more dangerous route of attempting to clean up failed (not active) reindex attempts so that being interrupted doesn't cause future attempts to fail.

I think with this PR we're still in a much better place than before so either of these should be ok to come as a follow up, but I thought it was important to acknowledge this isn't quite fully robust as-is.

bgentry (approved):

internal/maintenance/reindexer.go
brandur: Ah actually, that's a good point. I just put in a little bit of new infrastructure that causes the reindexer to delete any partial work if an error is specifically an `ErrStop` when failing from the reindex step specifically. It still doesn't try to remove artifacts in other contexts though.

bgentry (approved):
Looks great! :shipit:

brandur:
ty.

PR #936, opened 2025-05-28
Drop driver's `Executor.Exec` first return value (which is empty)

brandur:
Here, drop the first return value of the driver's `Executor.Exec`
function, which previously returned a `struct{}` in first position. When
I was first putting this all in, I figured it'd be better for consistency
if all driver functions returned exactly two values, even if one wasn't
really needed. Since then, we've picked up lots of driver functions that
return only an error, so that original premise doesn't hold anymore.

Returning only an error is advantageous in some cases because it lets
the return value be passed directly into functions. e.g.

    require.NoError(t, exec.Exec(ctx, ...))

I've been meaning to make this change for a while, and it seems like a
good time now given we're going to be making lots of changes to the
driver interface for the next release anyway.

### DIFF
--- a/client.go
+++ b/client.go
@@ -909,8 +909,7 @@ func (c *Client[TTx]) Start(ctx context.Context) error {
 		// available, the client appears to have started even though it's completely
 		// non-functional. Here we try to make an initial assessment of health and
 		// return quickly in case of an apparent problem.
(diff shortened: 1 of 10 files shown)


brandur:
ty.

PR #938, opened 2025-05-29
Better error in panic when accessing periodic jobs bundle in client that won't work

### LINKED ISSUE #937: Unable to insert a periodic job from external service
During the implementation of enqueuing a periodic job from external service, a panic occurred. The external service does not contain any code for registering workers, it stores data and passes it to the worker as parameters when enqueuing periodic jobs.

## How to reproduce

Initialize in an external service a client without worker registration:

```go
client, err := river.NewClient(riverpgxv5.New(workerPool), &river.Config{})
```

Because of [this condition](https://github.com/riverqueue/river/blob/master/client.go#L727), internal `client.periodicJobs` will not be initialized, which leads to panic when trying to insert a periodic job:

```go
client.PeriodicJobs().Add(j)
```

Is this behavior intentional, or is it a bug?

brandur:
Related to #937. If an insert-only client is created and the
`PeriodicJobs()` bundle is accessed, it'll panic on a nil pointer error.
The error is appropriate because if the client will never work jobs, it
can never be started, which means it can never win leader election,
which means that adding periodic jobs to it will never have any effect,
but the bad error message is definitely not ideal.

Here, detect this condition and panic with an error that better explains
the situation to the caller.

Fixes #937.


client.go
```diff
@@ -2086,7 +2086,13 @@ func (c *Client[TTx]) JobListTx(ctx context.Context, tx TTx, params *JobListPara
 
 // PeriodicJobs returns the currently configured set of periodic jobs for the
 // client, and can be used to add new ones or remove existing ones.
-func (c *Client[TTx]) PeriodicJobs() *PeriodicJobBundle { return c.periodicJobs }
+func (c *Client[TTx]) PeriodicJobs() *PeriodicJobBundle {
```
bgentry, line 2089: Maybe we want to add a note to the doc string about this behavior too?

brandur: Yep good point. Just added a new paragraph:

```
// This function should only be invoked on clients capable of running perioidc
// jobs. Running periodic jobs requires that the client be electable as leader
// to run maintenance services, and being electable as leader requires that a
// client be started. To be startable, a client must have Queues and Workers
// configured. Invoking this function will panic if these conditions aren't met.
```

PR #941, opened 2025-06-04
Break poll loop out into subroutine guaranteed to quit before producer stop

brandur:
This one's aimed at fixing the race detector failure as observed in [1].
Honestly, I couldn't repro this locally even at 1,000 iteration, but
reading the code, I'm pretty sure I know what the problem is.

The `fetchAndRunLoop` would start a separate goroutine that'd use a
timer to trigger a poll at `FetchPollInterval` so that fetches are
occasionally performed even if listen/notify missed something. This
inner goroutine would stop on context cancellation, but it wasn't
guaranteed to return before a stop finished. Therefore, as the stress
test called `Start` again, `fetchLimiter` could be modified while the
goroutine was still accessing it from the last run of the producer.

Fix this by making the fetch loop a "subroutine", which is a convention
used already for other goroutines in `Start`. `Start` waits on all these
goroutines to finish before completing a stop, therefore guaranteeing
that the next iteration of `Start` can do so safely.

[1] https://github.com/riverqueue/river/actions/runs/15419505042/job/43390545084

### DIFF
--- a/producer.go
+++ b/producer.go
@@ -392,10 +392,18 @@ func (p *producer) StartWorkContext(fetchCtx, workCtx context.Context) error {
 		}
 
 		var subroutineWG sync.WaitGroup
-		subroutineWG.Add(3)
 		subroutineCtx, cancelSubroutines := context.WithCancelCause(context.WithoutCancel(fetchCtx))
+
+		subroutineWG.Add(1)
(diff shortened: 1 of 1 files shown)


bgentry (approved):
:100:, this definitely looks like the cause. Nice fix! 🎉

brandur:
Thanks!

And oops — a few typos in titles/description that are fixed now.

PR #943, opened 2025-06-04
Prepare version v0.23.0

brandur:
Prepare version v0.23.0. This one is a little trickier than usual
because of the addition of a new package for SQLite. I ended up having
to manually change some version tags around in our `go.mod`s.

[skip ci]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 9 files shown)


brandur:
@bgentry Alright, I thinkkkk I got everything on this one. Removed all the `replace` TODOs I could find and swapped out SQLite versions for the version we're about to release.

bgentry (approved):
I think this looks good. I've been running the demo app locally against my latest branches for concurrency limiting fixes so we should be good to go, those were the last major changes.

brandur:
Great. Alright, going to cut this.

PR #945, opened 2025-06-04
rivercli: restore ability to set custom DriverProcurer

bgentry:
#870 broke a Pro CLI integration point for making Pro migrations available via a custom `DriverProcurer`. This restores that ability to set a custom procurer.

### DIFF
--- a/cmd/river/rivercli/command.go
+++ b/cmd/river/rivercli/command.go
@@ -46,10 +46,11 @@ type CommandOpts interface {
 
 // RunCommandBundle is a bundle of utilities for RunCommand.
 type RunCommandBundle struct {
-	DatabaseURL *string
-	Logger      *slog.Logger
-	OutStd      io.Writer
-	Schema      string
+	DatabaseURL    *string
+	DriverProcurer DriverProcurer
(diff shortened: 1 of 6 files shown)


brandur (approved):

cmd/river/rivercli/command.go
```diff
@@ -76,8 +77,8 @@ func RunCommand[TOpts CommandOpts](ctx context.Context, bundle *RunCommandBundle
 			}
 		}
 
-		var driverProcurer DriverProcurer
-		if databaseURL != nil {
+		driverProcurer := bundle.DriverProcurer
+		if driverProcurer == nil && databaseURL != nil {
```
bgentry: Ugh this doesn't quite cut it unfortunately. We'd have to initialize a DB pool separately in Pro prior to initializing its procurer (different than how this worked previously where it just passed in an existing pgx pool).

bgentry: Added an `InitPgxV5()` optional interface to allow injecting the pgxpool into a driver that can handle it. Tested the integration and this seems to work end-to-end.

PR #949, opened 2025-06-08
Have `PanicTB` respond to `RIVER_DEBUG` en var

brandur:
Just a small one, but I was trying to debug an example test and found it
to be quite annoying because test output is hidden by necessity so as
not to mess with stdout for purposes of the example out matching.

I'd previously added a special `WithLog` helper to `PanicTB` to add a
logger for this sort of debugging, but it's easy to forget it exists.

Here, modify things a bit to try an alternative so that when
`RIVER_DEBUG is enabled, `PanicTB` sends logging to stderr (so it
doesn't mess with stdout for the example). It might be easy to forget
this exists too, but at least it's a convention that exists elsewhere
for the test suite already and can be triggered without any code
changes.

### DIFF
--- a/rivershared/util/testutil/test_util.go
+++ b/rivershared/util/testutil/test_util.go
@@ -4,16 +4,18 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"os"
 )
 
 // See docs on PanicTB.
-type panicTB struct {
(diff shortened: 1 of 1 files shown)


brandur:
thx.

PR #950, opened 2025-06-08
Remove unnecessary transactions

brandur:
Here, remove unnecessary transactions around job completion. We
specifically start a transaction for completion only to run exactly one
operation against it, then commit it again. This is undesirable because:

* Each individual operation is already transaction. So if the completion
  operation fails, it's rolled back even without being in a transaction.

* For drivers that need to start a subtransaction for completion, this
  ends ups up nesting two transactions for no reason. e.g. The SQLite
  driver needs to run more than one DB operation during completion to
  ensure everything's atomic. It doesn't know whether it's in a
  transaction already, so it ends up opening a transaction despite
  already being in a transaction. This isn't that big of a deal, but
  it's unnecessary.


bgentry (approved):
I think the PR description and changelog are maybe slightly out of date based on how this was changed, but in general this looks good. We'll just need a matching tweak on the Pro side, as well as a warning about needing to upgrade the two in lockstep yet again.

These were the original APIs added to the `Pilot` interface and this change brings everything in line with how the others work now: always providing a full executor so the pilot can decided if/when transactions are appropriate.

PR #952, opened 2025-06-11
cleanly error on invalid Config.Schema

bgentry:
Had a thought to check if we were doing this for safety's sake, figured we should.


brandur (approved):
Yep, seems like a good idea.

PR #955, opened 2025-06-18
Bring all driver tests into separate package so they don't leak dependencies

brandur:
Here, modify River to try to do a better job of cloistering dependencies
to the packages that actually use them. We started this process by
package-ifying most things, but we kept most testing at the top level
that used these sub-packages, so although dependencies should end up
getting pulled into final binaries unless projects using River are
actually using them (e.g. say a project pulls in River but doesn't use
SQLite), they're still in `go.mod` and all projects resolving River will
have to resolve them.

We create a new package `riverdriver/riverdrivertest` and move all
driver tests and client driver tests into it. All the previous
dependencies that were required at the top level are pushed down into
this next package. In addition, wherever we needed to be testing edges
like when a driver doesn't support a listener, we add some new synthetic
drivers to simulate that instead of using `riverdatabasesql` directly.

An alternative approach to this would be to push each driver's tests
into its own driver-specific package (so all driver tests for SQLite
would be in `riverdriver/riversqlite` and that'd reference tests in
`riverdriver/riverdrivertest`). I was going to do that originally, but
decided not to go that way because we may want to run into the same
problem described above within any particular driver. For example, if we
want to test `riversqlite` on both SQLite and LibSQL, then both those
packages would have to come into `riversqlite`. However, if tests are
kept in `riverdrivertest` instead, then that package picks up all the
dependencies and driver packages stay clean. Another small benefit is
that it's kind of nice to have all the tests for each driver visible in
a single file `riverdrivertest/driver_test.go` because it makes broad
edits across all of them easier.

An argument could be made too that we should eventually try to rid the
top level of its dependency on `riverpgxv5` as well. This is currently
impossible due to its extremely extensive use in tests everywhere, but
is theoretically possible by moving those tests somewhere else instead.


riverdriver/riverdrivertest/driver_client_test.go
```diff
@@ -104,6 +108,72 @@ func TestClientWithDriverRiverSQLite(t *testing.T) {
...
+
+func (noOpArgs) Kind() string { return "noOp" }
+
+type noOpWorker struct {
+	river.WorkerDefaults[noOpArgs]
+}
+
+func (w *noOpWorker) Work(ctx context.Context, job *river.Job[noOpArgs]) error { return nil }
+
+// Try to keep this helper close to thee one found in th top-level package so we
```
bgentry:
```suggestion
// Try to keep this helper close to the one found in the top-level package so we
```

riverdriver/riverdrivertest/driver_client_test.go
```diff
@@ -104,6 +108,72 @@ func TestClientWithDriverRiverSQLite(t *testing.T) {
...
+		Schema:            schema,
+		Test: river.TestConfig{
+			Time: &riversharedtest.TimeStub{},
+		},
+		TestOnly: true, // disables staggered start in maintenance services
+		Workers:  workers,
+	}
+}
+
+// Try to keep this helper close to thee one found in th top-level package so we
```
bgentry:
```suggestion
// Try to keep this helper close to the one found in the top-level package so we
```

riverdriver/riverdrivertest/driver_client_test.go
```diff
@@ -104,6 +108,72 @@ func TestClientWithDriverRiverSQLite(t *testing.T) {
...
+	require.NoError(t, client.Start(ctx))
+
+	t.Cleanup(func() {
+		ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+		defer cancel()
+		require.NoError(t, client.Stop(ctx))
+	})
+}
+
+// Try to keep this helper close to thee one found in th top-level package so we
```
bgentry:
```suggestion
// Try to keep this helper close to the one found in the top-level package so we
```

riverdriver/riversqlite/example_sqlite_test.go
bgentry, line 1: Definitely makes sense to put this in the driver's pkg/mod, good call!

bgentry (approved):
Awesome, thanks for sorting this out! I think this seems CHANGELOG-worthy, agreed?

riverdriver/riversqlite/example_sqlite_test.go
brandur: Thx. Yeah, it seems to work. We also don't have a choice if we want to get rid of that dependency too :$

PR #957, opened 2025-06-21
Test support for LibSQL using SQLite driver

brandur:
LibSQL is a SQLite fork designed to be a little more to progress and
open-source contribution, and it seems to have picked up some real world
usage. They're trying to be careful in maintaining API compatibility
with SQLite, so we should be able to support it as long as they continue
to do so.

As of #955, we can now add testing for LibSQL without having to bring it
in as a project dependency. That gets pushed down to `riverdrivertest`
instead.

I put the new example test in `riverdrivertest` as well so that
`riversqlite` doesn't need to add a libSQL dependency. For consistency,
I also moved the SQLite example test to `riverdrivertest` as well so
they're both easy to find. A nice byproduct of that move is that
`riversqlite` doesn't need a dependency on `modernc.org/sqlite`, which
is kind of nice. It lets you bring your own SQLite driver.


riverdriver/riverdrivertest/driver_test.go
```diff
@@ -111,7 +112,59 @@ func TestDriverRiverPgxV5(t *testing.T) {
 		})
 }
 
-func TestDriverRiverSQLite(t *testing.T) {
+func TestDriverRiverLiteLibSQL(t *testing.T) { //nolint:dupl
```
bgentry, line 115: Is it worth setting these up to use the same test logic to avoid having to add the `nolint:dupl`? Or are there subtle differences between the two?

bgentry (approved):

riverdriver/riverdrivertest/driver_test.go
brandur: Yeah I struggle with these ones.

It is true that they are 95% the same. However, if you zoom out and look at what's happening, each body is basically one function call and a couple local variable declarations. There is definitely some duplication that falls out of anonymous functions injected into the function call and comments on them, so we could extract the function call into another function that could encapsulate the shared lines, but IMO this does have a non-zero effect to obfuscate (in that you now have to follow another layer of indirection to see what's happening) and reduce clarity.

I guess i'm not too against it, but it's _definitely_ not a clear win like `dupl` would suggest. More like a close to 50/50 tradeoff.

Gonna leave it as is for now I think, and possibly reorient later if it comes up again.

PR #962, opened 2025-06-28
Implement `JobDeleteMany` operations

brandur:
Here, add functions for `Client.JobDeleteMany` which lets jobs be
deleted in batches. The main impetus for this is to give us a way of
implementing a "purge queue" function, which has been previously
requested by users, and which is generally just a good feature.

We copy the `JobList` API very closely and use almost all the same
implementation, with the caveat that I've removed some of the more
complex features that I don't think will be as necessary for deletion,
with the sorting features being the biggest one, but also arbitrary SQL
in `WHERE` queries. These can always be added in later because they're
purely additive to the API, but I'm hoping that we won't need them.

### DIFF
--- a/riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
+++ b/riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
@@ -122,6 +122,29 @@ WHERE id IN (
     LIMIT @max::bigint
 );
 
+-- name: JobDeleteMany :many
+WITH jobs_to_delete AS (
+    SELECT *
+    FROM /* TEMPLATE: schema */river_job
+    WHERE /* TEMPLATE_BEGIN: where_clause */ true /* TEMPLATE_END */
+        AND state != 'running'
+    ORDER BY /* TEMPLATE_BEGIN: order_by_clause */ id /* TEMPLATE_END */
+    LIMIT @max::int
+    FOR UPDATE
+    SKIP LOCKED
+),
+deleted_jobs AS (
+    DELETE FROM /* TEMPLATE: schema */river_job
+    WHERE id IN (SELECT id FROM jobs_to_delete)
+    RETURNING *
+)
+-- this last SELECT step is necessary because there's no other way to define
+-- order records come back from a DELETE statement
+SELECT *
+FROM /* TEMPLATE: schema */river_job
+WHERE id IN (SELECT id FROM deleted_jobs)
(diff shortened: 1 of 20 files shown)


riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
```diff
@@ -122,6 +122,27 @@ WHERE id IN (
...
 );
 
+-- name: JobDeleteMany :many
+WITH jobs_to_delete AS (
+    SELECT *
+    FROM /* TEMPLATE: schema */river_job
+    WHERE /* TEMPLATE_BEGIN: where_clause */ true /* TEMPLATE_END */
+        AND state != 'running'
+    ORDER BY /* TEMPLATE_BEGIN: order_by_clause */ id /* TEMPLATE_END */
+    LIMIT @max::int
```
bgentry, line 132: Would a `FOR UPDATE` in here provide a small bit of extra safety?

```suggestion
    SELECT * FOR UPDATE
    FROM /* TEMPLATE: schema */river_job
    WHERE /* TEMPLATE_BEGIN: where_clause */ true /* TEMPLATE_END */
        AND state != 'running'
    ORDER BY /* TEMPLATE_BEGIN: order_by_clause */ id /* TEMPLATE_END */
    LIMIT @max::int
```

riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
```diff
@@ -122,6 +122,27 @@ WHERE id IN (
...
+),
+deleted_jobs AS (
+    DELETE FROM /* TEMPLATE: schema */river_job
+    WHERE id IN (SELECT id FROM jobs_to_delete)
+    RETURNING *
+)
+-- this last SELECT step is necessary because there's no other way to define
+-- order records back from a DELETE statement
```
bgentry:
```suggestion
-- order records come back from a DELETE statement
```

riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
brandur: WFM. Changed.

riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
brandur: Thx, fixed.

PR #963, opened 2025-06-28
Have the reindexer reindex more `river_job` indexes

brandur:
Here, expand the scope of the reindexer so that beyond the pair of
indexes it reindexes right now, it also reindexes all the other ones in
the `river_job` table. This won't make a huge different for many uses of
River, but it'll make a particular difference in cases where for some
amount of time `river_job` became very large, but has since contracted
to a much more modest size. B-trees expand but never shrink, so in that
case the indexes will be left very large despite holding very little,
and the only way to get them back down to size is to rebuild them.

More reindexes will always put more load on the database, but we're
still doing that as carefully as we can in that indexes are rebuilt one
by one, and using `CONCURRENTLY`. Back in #935 we added protection so
that in case of a situation where index builds are too expensive and
start to fail, the reindexer will bail out instead of trying and failing
over and over again.


client_test.go
```diff
@@ -6428,7 +6428,8 @@ func Test_NewClient_Defaults(t *testing.T) {
 	require.False(t, enqueuer.StaggerStartupIsDisabled())
 
 	reindexer := maintenance.GetService[*maintenance.Reindexer](client.queueMaintainer)
-	require.Equal(t, []string{"river_job_args_index", "river_job_metadata_index"}, reindexer.Config.IndexNames)
+	require.Contains(t, reindexer.Config.IndexNames, "river_job_args_index")
+	require.Contains(t, reindexer.Config.IndexNames, "river_job_metadata_index")
```
bgentry: Want to add assertions for the others too?

brandur: Added, but hmm, FWIW I don't think it's necessary for this to be an exhaustive list. We're really trying to test that the reindexer is being configured with an expected set of default indexes, and as long as that's true, it should be up to the reindexer to define what the default set of indexes are. It's okay from the client's perspective if the list changes somewhat.

bgentry: I guess I was mostly thinking it'd be good to have it be complete for if/when we allow customizing the list, but that's a preemptive concern.

PR #970, opened 2025-06-30
Have `HookWorkEnd.WorkEnd` receive a `JobRow` in addition to an error

brandur:
Here, modify the `HookWorkEnd.WorkEnd` so that it receives a `JobRow` of
the job that just finished working in addition to an error. It's not
great having to change this, but having a `JobRow` to work with will be
crucial for most functionality that people might like to use this hook
to implement.

It's a breaking change, but hopefully not a bad one because I think not
having `JobRow` previously is a sign that not many people had yet tried
to implement anything with `HookWorkEnd` yet, because they would've
needed a `JobRow` for most of what they wanted to do.

### DIFF
--- a/rivertype/river_type.go
+++ b/rivertype/river_type.go
@@ -300,6 +300,7 @@ type JobInsertParams struct {
 // List of hook interfaces that may be implemented:
 // - HookInsertBegin
 // - HookWorkBegin
+// - HookWorkEnd
 //
 // More operation-specific interfaces may be added in future versions.
 type Hook interface {
@@ -314,6 +315,7 @@ type Hook interface {
 type HookInsertBegin interface {
 	Hook
 
+	// InsertBegin is invoked just before a job is inserted to the database.
 	InsertBegin(ctx context.Context, params *JobInsertParams) error
 }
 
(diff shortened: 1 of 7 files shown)


rivertype/river_type.go
```diff
@@ -338,18 +350,22 @@ type HookWorkEnd interface {
...
 	// 	}
 	// 	return err
 	//
 	// If a hook does not want to modify an error result, it should make sure to
 	// return whatever error value it received as its argument whether that
 	// error is nil or not.
 	//
+	// The JobRow received by WorkEnd is the same one passed to HookWorkBegin's
+	// WorkBegin. Its state, errors, next scheduled at time, etc. have not yet
+	// been updated based on the latest work result.
```
nickmarx12345678, line 363: So roughly if I wanted to check "is error and retries exhausted", e.g. "was this the last attempt?" we could look at
```golang
func WorkEnd(ctx,job,err) {
  if err != nil && job.Attempts >= jobMaxRetries - 1 {
    fmt.Sprint("job failed and will not be retried")
  }
}
```
is that fair?

bgentry: Yeah something like that should work.

bgentry (approved):
Yeah I don't love breaking it but this seems worthwhile. good call!

PR #974, opened 2025-07-05
Add maximum bound for `attempted_by` array

brandur:
As described by #972, it may be possible for huge numbers of snoozes to
blow out a job row's `attempted_by` array as a job is locked over and
over again. Multiplied by many jobs, this can produce vast quantities of
data that gets sent over the network.

Here, put in a ratchet on `attempted_by` so that if the array becomes
larger than 100 elements, we knock the oldest one off in favor of the
most recent client and the most fresh 99.

Unfortunately the implementation isn't particularly clean in either
Postgres or SQLite. In Postgres it would've been cleaner if we'd had the
`attempted_by` in reverse order so the new client was on front because
the built-in array functions would be friendlier to that layout, but
because it's not, we have to do something a little hackier involving a
`CASE` statement instead.

SQLite is even worse. SQLite has no array functions at all, which
doesn't help, but moreover every strategy I tried ended up blocked by a
sqlc SQLite bug, so after trying everything I could think of, I ended up
having to extract the piece that does the array truncation into a SQL
template string to get this over the line. This could be removed in the
future if any one of a number of outstanding sqlc bugs are fixed (e.g.
[1]).

[1] https://github.com/sqlc-dev/sqlc/pull/3610

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The reindexer maintenance service now reindexes all `river_job` indexes, including its primary key. This is expected to help in situations where the jobs table has in the past expanded to a very large size (which makes most indexes larger), is now a much more modest size, but has left the indexes in their expanded state. [PR #963](https://github.com/riverqueue/river/pull/963).
 - The River CLI now accepts a `--target-version` of 0 with `river migrate-down` to run all down migrations and remove all River tables (previously, -1 was used for this; -1 still works, but now 0 also works). [PR #966](https://github.com/riverqueue/river/pull/966).
 - **Breaking change:** The `HookWorkEnd` interface's `WorkEnd` function now receives a `JobRow` parameter in addition to the `error` it received before. Having a `JobRow` to work with is fairly crucial to most functionality that a hook would implement, and its previous omission was entirely an error. [PR #970](https://github.com/riverqueue/river/pull/970).
+- Add maximum bound to each job's `attempted_by` array so that in degenerate cases where a job is run many, many times (say it's snoozed hundreds of times), it doesn't grow to unlimited bounds. [PR #974](https://github.com/riverqueue/river/pull/974).
 
 ### Fixed
 
--- a/riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
+++ b/riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
@@ -159,7 +159,7 @@ WITH locked_jobs AS (
         priority ASC,
         scheduled_at ASC,
         id ASC
-    LIMIT @max::integer
+    LIMIT @max_to_lock::integer
     FOR UPDATE
     SKIP LOCKED
 )
@@ -169,7 +169,14 @@ SET
     state = 'running',
     attempt = river_job.attempt + 1,
     attempted_at = coalesce(sqlc.narg('now')::timestamptz, now()),
-    attempted_by = array_append(river_job.attempted_by, @attempted_by::text)
+    attempted_by = array_append(
+        CASE WHEN array_length(river_job.attempted_by, 1) >= @max_attempted_by::int
+        -- +2 instead of +1 because Postgres array indexing starts at 1, not 0.
+        THEN river_job.attempted_by[array_length(river_job.attempted_by, 1) + 2 - @max_attempted_by:]
+        ELSE river_job.attempted_by
+        END,
+        @attempted_by::text
+    )
 FROM
     locked_jobs
 WHERE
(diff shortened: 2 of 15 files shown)


riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
```diff
@@ -169,7 +169,18 @@ SET
...
+                            -- +2 instead of +1 because in one of those history
+                            -- making mistakes that's likely in aggregate cost
+                            -- humanity >$10B in bugs and lost productivity by
+                            -- now, like strings, Postgres array indexing start
+                            -- at 1 instead of 0.
+                            river_job.attempted_by[array_length(river_job.attempted_by, 1) + 2 - @max_attempted_by:],
+                            @attempted_by::text
+                        )
+                        ELSE array_append(river_job.attempted_by, @attempted_by::text)
+                        END
```
brandur: @bgentry I don't love how gross this got. Feel free to try and take a shot at golfing it down a bit.

It's unfortunate that new entries were appended instead of prepended, because if they were prepended, we could do this much more nicely like:

``` sql
    attempted_by = trim_array(
        array_prepend(river_job.attempted_by, @attempted_by::text),
        array_length(river_job.attempted_by, 1) - @max_attempted_by
    )
```

Not worth trying to change it at this point though.

We could also use more of a table-based approach like SQLite, but I left this as array functions as I suspect they're more performant.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - The River CLI now accepts a `--target-version` of 0 with `river migrate-down` to run all down migrations and remove all River tables (previously, -1 was used for this; -1 still works, but now 0 also works). [PR #966](https://github.com/riverqueue/river/pull/966).
 - **Breaking change:** The `HookWorkEnd` interface's `WorkEnd` function now receives a `JobRow` parameter in addition to the `error` it received before. Having a `JobRow` to work with is fairly crucial to most functionality that a hook would implement, and its previous omission was entirely an error. [PR #970](https://github.com/riverqueue/river/pull/970).
+- Add maximum bound to each job's `attempted_by` array so that in degenerate cases where a job is locked many, many times (say it's snoozed hundreds of times), it doesn't grow to unlimited bounds. [PR #974](https://github.com/riverqueue/river/pull/974).
```
bgentry: `s/locked/run`?
```suggestion
- Add maximum bound to each job's `attempted_by` array so that in degenerate cases where a job is run many, many times (say it's snoozed hundreds of times), it doesn't grow to unlimited bounds. [PR #974](https://github.com/riverqueue/river/pull/974).
```

riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
bgentry: This might be just a little cleaner than the `CASE`?

```diff
diff --git a/riverdriver/riverdatabasesql/internal/dbsqlc/river_job.sql.go b/riverdriver/riverdatabasesql/internal/dbsqlc/river_job.sql.go
index df48a3a..9025b95 100644
--- a/riverdriver/riverdatabasesql/internal/dbsqlc/river_job.sql.go
+++ b/riverdriver/riverdatabasesql/internal/dbsqlc/river_job.sql.go
@@ -281,18 +281,14 @@ SET
     state = 'running',
     attempt = river_job.attempt + 1,
     attempted_at = coalesce($1::timestamptz, now()),
-    attempted_by = CASE WHEN array_length(river_job.attempted_by, 1) >= $2::int
-                        THEN array_append(
-                            -- +2 instead of +1 because in one of those history
-                            -- making mistakes that's likely in aggregate cost
-                            -- humanity >$10B in bugs and lost productivity by
-                            -- now, like strings, Postgres array indexing start
-                            -- at 1 instead of 0.
-                            river_job.attempted_by[array_length(river_job.attempted_by, 1) + 2 - $2:],
-                            $3::text
-                        )
-                        ELSE array_append(river_job.attempted_by, $3::text)
-                        END
+    attempted_by = (
+        coalesce(river_job.attempted_by, '{}'::text[])
+        || $2::text
+        -- +2 instead of +1 because in one of those history making mistakes
+        -- that's likely in aggregate cost humanity >$10B in bugs and lost
+        -- productivity by now, like strings, Postgres array indexing start
+        -- at 1 instead of 0.
+    )[ GREATEST(cardinality(river_job.attempted_by) + 2 - $3::int, 1) : ]
 FROM
     locked_jobs
 WHERE
```

brandur: I modified the existing one to flatten it out a bit more, but TBH I have trouble reading this new version, so going to largely leave it as the original.

PR #975, opened 2025-07-07
Add JobRetry to Pilot

### LINKED ISSUE #971: Feature Request: retry job in sequence
Right now, when you have a bunch of discarded jobs that are part of a sequence, you need to retry them one by one manually; otherwise, all of them run at the same time if capacity is available. It would be nice if "Retry" kind of re-adds the job to the sequence, so you can retry many of them at the same time without skipping the sequence behavior

bgentry:
Allow the Pro pilot to further customize this behavior. No changelog entry because it's internal-only.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1412,11 +1412,7 @@ func (c *Client[TTx]) JobGetTx(ctx context.Context, tx TTx, id int64) (*rivertyp
 // MaxAttempts is also incremented by one if the job has already exhausted its
 // max attempts.
 func (c *Client[TTx]) JobRetry(ctx context.Context, id int64) (*rivertype.JobRow, error) {
-	return c.driver.GetExecutor().JobRetry(ctx, &riverdriver.JobRetryParams{
-		ID:     id,
-		Now:    c.baseService.Time.NowUTCOrNil(),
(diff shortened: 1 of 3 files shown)


riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
```diff
@@ -418,7 +418,10 @@ WITH job_to_update AS (
 updated_job AS (
     UPDATE /* TEMPLATE: schema */river_job
     SET
-        state = 'available',
```
brandur: Doh, well, I'm sure you noticed this, but this may be the first case where the Pro functionality is starting to leak out into the OSS project, which isn't ideal. There are no mentions of `seq_key` in riverqueue/river to date because functionality's been isolated to the pro modules.

bgentry: I pushed a new version of this which removes the actual behavioral change and merely adds JobRetry to the Pilot interface. I think we have a clearer plan now on Pro-specific UI extensions so hopefully we don't need to worry about dealing with these Pro behaviors when using the OSS library.

PR #983, opened 2025-07-15
Integrate riverui queries into driver system, add Client.Schema()

bgentry:
This integrates all of riverui's non-Pro queries into this library's driver system, which paves the way for riverui to support multiple driver backends (sqlite) and to add more Pro-specific UI extensions. These queries were generally imported as-is except for the sqlite variants which required some rework.

I also added a `Client.Schema()` accessor because I couldn't see any way for riverui to pull the schema from the `river.Client` in its constructor opts. Rather than add another param for it there, it seemed cleaner to just expose a reader method on the client. This all ends up being required if we want the UI to properly support custom schemas while also having it directly call into pilot and driver methods without having to expose all this functionality directly on the `Client` and bloating its API.


CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - The project now tests against [libSQL](https://github.com/tursodatabase/libsql), a popular SQLite fork. It's used through the same `riversqlite` driver that SQLite uses. [PR #957](https://github.com/riverqueue/river/pull/957)
 - Added `JobDeleteMany` operations that remove many jobs in a single operation according to input criteria. [PR #962](https://github.com/riverqueue/river/pull/962)
+- Added `Client.Schema()` method to return a client's configured schema. [PR #XXX](https://github.com/riverqueue/river/pull/XXX).
+- Integrated riverui queries into the driver system to pave the way for multi-driver UI support. [PR #XXX](https://github.com/riverqueue/river/pull/XXX).
```
brandur: s/XXX/983/ for these two lines.

PR #986, opened 2025-07-18
allow customizing processing order

### LINKED ISSUE #985: Allow users to customize processing order
Jobs are currently processed in a lifo fashion and this is hardcoded in the queries.
I have a use case that would benefit from doing the other way.

TC5027:
Fix #985


TC5027:
I can review my implementation to avoid duplicating the queries if that's a concern, I did so because from my knowledge there is no other way to do so for sqlite
Also if you want me to add tests i can probably do.
Why reasoning on scheduled_at is wrong ?

PR #992, opened 2025-07-31
Set periodic job IDs to metadata where present

brandur:
Set metadata containing the periodic job ID of inserted jobs that were
inserted based on a periodic job with an ID. This lets inserted jobs be
associated with periodic jobs later after the fact if necessary.

### DIFF
--- a/internal/maintenance/periodic_job_enqueuer.go
+++ b/internal/maintenance/periodic_job_enqueuer.go
@@ -336,7 +336,7 @@ func (s *PeriodicJobEnqueuer) Start(ctx context.Context) error {
 					continue
 				}
 
-				if insertParams, ok := s.insertParamsFromConstructor(ctx, periodicJob.ConstructorFunc, now); ok {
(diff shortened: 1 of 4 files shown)


internal/maintenance/periodic_job_enqueuer.go
```diff
@@ -23,6 +23,8 @@ import (
 	"github.com/riverqueue/river/rivertype"
 )
 
+const metadataKeyPeriodicJobID = "river:periodic_job_id"
```
bgentry: Thoughts on putting this const with other metadata keys in https://github.com/riverqueue/river/blob/ae827888c816384130a3a13dcd9f038d504fd167/internal/rivercommon/river_common.go#L23 or https://github.com/riverqueue/river/blob/ae827888c816384130a3a13dcd9f038d504fd167/rivertype/river_type.go#L14 ?

bgentry (approved):
Nice! ⚡

internal/maintenance/periodic_job_enqueuer.go
brandur: Ah yeah works for me.

brandur: Done.

PR #998, opened 2025-08-03
Always upsert initial periodic job records

brandur:
Tweak the behavior for periodic job records inserted through the pilot
so that instead of being upserted the first time they're worked, upsert
them immediately on program start up, even if `RunOnStart` isn't
configured for them. This won't actually cause them to run, but will add
the durable record a little earlier than it would be before.


bgentry (approved):
Awesome! Looks like it was indeed fairly easy to fix, thank you :clap:

PR #1004, opened 2025-08-10
Add job cleaner excluded queues that can be configured through pilot

brandur:
Add a new pilot function that lets it dictate that certain queues should
be excluded from the standard cleaning process that manages retention.
Along with the `JobDeleteBefore` driver function picking up a new
`QueuesExcluded` parameter, we also add its inverse of `QueuesIncluded`
which lets us delete jobs from a specific set of queues.

### DIFF
--- a/riverdriver/riverdrivertest/riverdrivertest.go
+++ b/riverdriver/riverdrivertest/riverdrivertest.go
@@ -926,66 +926,180 @@ func Exercise[TTx any](ctx context.Context, t *testing.T,
 	t.Run("JobDeleteBefore", func(t *testing.T) {
 		t.Parallel()
 
-		exec, _ := setup(ctx, t)
-
 		var (
 			horizon       = time.Now()
 			beforeHorizon = horizon.Add(-1 * time.Minute)
 			afterHorizon  = horizon.Add(1 * time.Minute)
 		)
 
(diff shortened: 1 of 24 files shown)


riverdriver/riverdrivertest/riverdrivertest.go
```diff
@@ -925,66 +925,180 @@ func Exercise[TTx any](ctx context.Context, t *testing.T,
...
+			// instead of positional placeholders (?1) like most parameters. The
+			// trick of putting it at the end works, but only if you have
+			// exactly one `sqlc.slice` needed. If you need multiple and they
+			// need to be interspersed with other parameters (like in the case
+			// of `queues_excluded` and `queues_included`), everything stops
+			// working real fast. I could have worked around this by breaking
+			// the SQLite version of this operation into two sqlc queries, but
+			// since we only expect to need `queues_excluded` on SQLite (and not
+			// `queues_included` for the foreseeable future), I've just set
+			// SQLite to not support `queues_included` for the time being.
```
bgentry, line 1057: 😕

bgentry (approved):

riverdriver/riverdrivertest/riverdrivertest.go
brandur: Yeah, ugh :( We should chat about this, but using Sqlc for SQLite might've been a mistake. I was somewhat optimistic that I'd be able to start landing some fixes over there, but the one fix I got in is still unreleased months later, and a few other high quality patches from others are ignored. I'm honestly not 100% sure what to do because I don't really see any good alternatives. Options are kind of "bad" or "worse".

PR #1010, opened 2025-08-12
Eliminate TODO stored to error trace from job rescuer

brandur:
Here, eliminate a TODO that's stored to an error's `trace` when a job is
rescued by `JobRescuer`. I've changed this to an empty string instead
and documented that it'll be an empty string.


bgentry (approved):
Whoops, good find! Thanks

Issue #995, opened 2025-08-01
Increased IO Wait in postgres


Devidxyz:
### Introduction
We have been using River for about a year, and we’ve had almost no issues that we couldn’t solve by tuning the configs or modifying the workers. Overall, we are very satisfied. However, in the last week, we encountered an issue twice that we haven’t been able to resolve yet.

We have the following versions in production:
```
github.com/riverqueue/river v0.20.2
github.com/riverqueue/river/riverdriver/riverpgxv5 v0.20.2
github.com/riverqueue/river/rivertype v0.20.2
```

### Use-case

We run a go app in k8s, the rest API pods emit events, we insert these events to postgres as river jobs, and another deployment with river workers in it processes the events (make HTTP calls for the webhook subscriptions).

### The problem

We noticed that fewer webhook calls were being made that the usual. 
We checked what's happening in our DB with `select state, count(*) from webhooks.river_job;`:
```
   state   | count  
-----------+--------
 available |   1.404
 completed | 272.082
 discarded |  44.583
 retryable |  98.424
 running   |   2.739
```
Then we checked the error fields of the retryable jobs, and all of them errored with "Job rescued by rescuer".
We initially thought our workers couldn't process the jobs in time, but we only do an HTTP call in each job which has a 2 second timeout, so we were almost certain it's not the case.
Then we found many `BatchCompleter: Completer error (will retry after sleep)` and `BatchCompleter: Too many errors; giving up` log entries, all showing `timeout: context deadline exceeded` error. There were also `JobScheduler: Error scheduling jobs` and `Elector: Error attempting reelection` errors with the same `context deadline exceeded` cause.

For postgres we have a 4vCPU 16GB memory Cloud SQL instance. (Which is shared among some apps, but the current one is the most significant.)
<img width="2177" height="663" alt="Image" src="https://github.com/user-attachments/assets/1607d8b4-6eb4-439c-a712-124fc8b063e9" />
Here we see a significant increase in IO wait. We eventually switch to maintenance mode around 7:30PM and tried to fix the error, that's why the load decreased, it wasn't natural. (The horizontal line in the chart is a different metric, it does not mean we hit the CPU limit or something. The CPU utilization was ~70-80% so we never hit 100%, when we doulbed the resources the utilization was ~60%.)

### What we tried
- We deleted all the completed jobs, but the issue persisted
- We updated all the retryable and running jobs to available status, but the issue persisted
- We increased the DB resources to 8vCPU 32GB ram, the issue persisted
[...]

PR #1013, opened 2025-08-13
Schedule jobs in smaller batches when it appears the database is degraded

brandur:
The job scheduler currently tries to work in batches that are quite
large, trying to schedule 10,000 jobs at once. Normally this is fine
because there's far fewer than 10,000 jobs to schedule, or if there are
many jobs you're generally on a large Postgres instance where scheduling
a lot of jobs at once should be fine.

However, in a situation where Postgres is severely degraded, it might be
possible that get yourself in a situation where the scheduler can never
succeed. It tries to schedule 10k jobs, fails due to timeout, tries
again, fails again, etc. Jobs accumulate and the situation keeps getting
worse. You can see an example of this in #995, where the users in
question ran out of IOPS on their DB and saw their situation steadily
degrade.

Here, put in a partial stab at an approach for ameliorating this. We add
a circuit breaker to the scheduler such that if its scheduling operation
fails three times in a row due to timeout, the circuit breaks, and the
scheduler switches from its default batch of 10k to a much reduced batch
of 1,000 jobs. The timeout must occur three times in a row for this to
occur. If timeouts are intermittent (i.e. in the case of an only
partially degraded database), the breaker is allowed to reset, and
operations continue to use the 10k limit.

An executive decision I made here is to have the circuit breaker stay
open once it trips, so after the scheduler switches to a reduced batch
of 1k, it'll continue using the 1k batch until the program restarts. The
reason I designed it this way is that in my experience after a database
becomes degraded, it tends to stay degraded over the long term. We could
have it so that it switches back to 10k job batches after some time, but
more likely than not those 10k operations will just start failing again,
and make things worse by putting more load on the database while they
do. I could see tweaking or reexamining this behavior in the future, but
I don't think either way we could go here is that much more correct than
the other.

I've only applied this to the job scheduler for this initial design, but
I'd like to eventually bring this to all/most of the other maintenance
services, many of which haven't had a huge amount of thought put into
how their batch size works since they were first introduced.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Jobs inserted from periodic jobs with IDs now have metadata `river:periodic_job_id` set so they can be traced back to the periodic job that inserted them. [PR #992](https://github.com/riverqueue/river/pull/992).
 - The unused function `WorkerDefaults.Hooks` has been removed. This is technically a breaking change, but this function was a vestigal refactoring artifact that was never used by anything, so in practice it shouldn't be breaking. [PR #997](https://github.com/riverqueue/river/pull/997).
 - Periodic job records are upserted immediately through a pilot when a client is started rather than the first time their associated job would run. This doesn't mean they're run immediately (they'll only run if `RunOnStart` is enabled), but rather just tracked immediately. [PR #998](https://github.com/riverqueue/river/pull/998).
+- The job scheduler still schedules jobs in batches of up to 10,000, but when it encounters a series of consecutive timeouts it assumes that the database is in a degraded state and switches to doing work in a smaller batch size of 1,000 jobs. [PR #1013](https://github.com/riverqueue/river/pull/1013).
 
 ### Fixed
 
--- a/internal/maintenance/job_scheduler.go
+++ b/internal/maintenance/job_scheduler.go
@@ -8,6 +8,7 @@ import (
 	"log/slog"
 	"time"
 
+	"github.com/riverqueue/river/internal/circuitbreaker"
 	"github.com/riverqueue/river/riverdriver"
 	"github.com/riverqueue/river/rivershared/baseservice"
 	"github.com/riverqueue/river/rivershared/riversharedmaintenance"
@@ -20,8 +21,9 @@ import (
 )
 
 const (
-	JobSchedulerIntervalDefault = 5 * time.Second
-	JobSchedulerLimitDefault    = 10_000
+	JobSchedulerBatchSizeReduced = 1_000
+	JobSchedulerBatchSizeDefault = 10_000
+	JobSchedulerIntervalDefault  = 5 * time.Second
 )
 
 // Test-only properties.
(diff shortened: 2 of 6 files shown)


internal/maintenance/job_scheduler.go
```diff
@@ -83,15 +91,28 @@ type JobScheduler struct {
...
+		reducedBatchSizeBreaker: circuitbreaker.NewCircuitBreaker(&circuitbreaker.CircuitBreakerOptions{
+			Limit:  3,
+			Window: 10 * time.Minute,
+		}),
 		config: (&JobSchedulerConfig{
-			Interval:     cmp.Or(config.Interval, JobSchedulerIntervalDefault),
-			Limit:        cmp.Or(config.Limit, JobSchedulerLimitDefault),
-			NotifyInsert: config.NotifyInsert,
-			Schema:       config.Schema,
+			BatchSizeDefault: cmp.Or(config.BatchSizeDefault, JobSchedulerBatchSizeDefault),
```
bgentry, line 108: It could also be that 10k is too much as a default for typical systems, but we can still change that later

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Jobs inserted from periodic jobs with IDs now have metadata `river:periodic_job_id` set so they can be traced back to the periodic job that inserted them. [PR #992](https://github.com/riverqueue/river/pull/992).
 - The unused function `WorkerDefaults.Hooks` has been removed. This is technically a breaking change, but this function was a vestigal refactoring artifact that was never used by anything, so in practice it shouldn't be breaking. [PR #997](https://github.com/riverqueue/river/pull/997).
 - Periodic job records are upserted immediately through a pilot when a client is started rather than the first time their associated job would run. This doesn't mean they're run immediately (they'll only run if `RunOnStart` is enabled), but rather just tracked immediately. [PR #998](https://github.com/riverqueue/river/pull/998).
+- The job scheduler still schedules jobs in batches of 10,000, but when it encounters a series of consecutive timeouts it assumes that the database is in a degraded state and switches to doing work in a smaller batch size of 1,000 jobs. [PR #1013](https://github.com/riverqueue/river/pull/1013).
```
```suggestion
- The job scheduler still schedules jobs in batches of up to 10,000, but when it encounters a series of consecutive timeouts it assumes that the database is in a degraded state and switches to doing work in a smaller batch size of 1,000 jobs. [PR #1013](https://github.com/riverqueue/river/pull/1013).
```

bgentry (approved):
Excellent stab at improving things here, let's give it a shot!

internal/maintenance/job_scheduler.go
brandur: Yeah true. 10k might be too much.

However, we're a little inconsistent on batch sizes and a lot of the other maintenance services are using 1k which I'd say is too _small_. There's a brief sleep between batches, and I worry that in a really high frequency system doing only 1k at a time with a sleep might not be enough.

So maybe the right answer is in the middle there somewhere. I think probably closer to 10k than 1k, but it's pretty arbitrary.

There's also an argument that it should be configurable. Will have to consider that one too.

PR #1014, opened 2025-08-13
Start testing against Go 1.25

brandur:
Start testing against Go 1.25. I didn't change the `go.mod` version. I
figure it's probably fine to leave that as low as is convenient for us
to maintain as a courtesy to those at companies where it's harder to
upgrade.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,12 +27,20 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        # Run the 4 latest Postgres versions against the latest Go version:
-        go-version:
-          - "1.24"
-        postgres-version: [14, 15, 16, 17]
         include:
-          # Also run previous Go version against the latest Postgres version:
+          # Run the 4 latest Postgres versions against the latest Go version:
+          - go-version: "1.25"
+            postgres-version: 17
+          - go-version: "1.25"
+            postgres-version: 16
+          - go-version: "1.25"
+            postgres-version: 15
+          - go-version: "1.25"
+            postgres-version: 14
+
+          # Also run a couple previous Go versions against the latest Postgres version:
+          - go-version: "1.24"
+            postgres-version: 17
           - go-version: "1.23"
             postgres-version: 17
       fail-fast: false
--- a/cmd/river/rivercli/river_cli_test.go
+++ b/cmd/river/rivercli/river_cli_test.go
@@ -179,8 +179,9 @@ func TestBaseCommandSetIntegration(t *testing.T) {
 
 		buildInfo, _ := debug.ReadBuildInfo()
 
-		require.Equal(t, strings.TrimSpace(fmt.Sprintf(`
-River version (unknown)
+		// `devel` on 1.25, `unknown` on versions previous to that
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -27,12 +27,20 @@ jobs:
...
+          - go-version: "1.25"
+            postgres-version: 15
+          - go-version: "1.25"
+            postgres-version: 14
+
+          # Also run a couple previous Go versions against the latest Postgres version:
+          - go-version: "1.24"
+            postgres-version: 17
           - go-version: "1.23"
             postgres-version: 17
```
bgentry, line 45: Thoughts on removing the v1.23 from here? I don't think it adds much tbh.

bgentry (approved):

.github/workflows/ci.yaml
brandur: I'm kindaa thinking that we should lave that in for the time being. It's still the version of Go in all the `go.mod`s, and we should be testing against what our ostensible lowest version is.

We could bump all the `go.mod`'s to 1.24, but I don't think we should — only the last two versions of Go are supported in practice, but a lot of people get stuck behind, at least for a while. If there's a feature we really want in the latest or something we really wanted to drop, I can see the logic for an aggressive push for new Go versions, but given there isn't this time around ... may as well leave that lower version constraint as low as is convenient. At least for now.

Thoughts?

.github/workflows/ci.yaml
```diff
@@ -27,12 +27,20 @@ jobs:
...
-          # Also run previous Go version against the latest Postgres version:
+          # Run the 4 latest Postgres versions against the latest Go version:
+          - go-version: "1.25"
+            postgres-version: 17
+          - go-version: "1.25"
+            postgres-version: 16
+          - go-version: "1.25"
+            postgres-version: 15
+          - go-version: "1.25"
+            postgres-version: 14
```
brandur, line 39: I ended up flattening this matrix out because I don't think it was working as intended before, or if it was, it was only doing so accidentally.

GitHub Actions [matrixes operation quite unintuitively](https://github.com/riverqueue/homepage/pull/279#pullrequestreview-3118492160). I read those docs three times and still don't 100% understand it, but if I added only 1.24/17, it would buit it on all of Postgres 17/16/15/14, which is obviously not what was intended. Found it easier just to unroll the matrix like this which isn't too many more LOCs anwyay.

bgentry: I think the intention was to test the latest Go version against all recent supported Postgres versions, and then separately test each recent Go version against the latest Postgres. This was on the assumption that cross PG-Go version incompatibilities are super unlikely, whereas it's more likely that some dependency doesn't work as well in a particular Go version or against a certain PG version.

bgentry (approved):

.github/workflows/ci.yaml
brandur: Oh yeah, I totally understood that part of it. What I meant is just that the matrix functionality is kind of weirder than you'd expect. If you wanna play with it a bit, I'm sure there must be another way to get it working as intended, but my brain was too strained to figure it out.

PR #1025, opened 2025-08-26
don't double-log fetch errors

bgentry:
While digging into some Pro errors that were surfacing, I realized we were logging these fetch errors twice with the same message each time they occur (albeit with slightly different attrs). This removes one in favor of keeping all logging on the channel receive side.


bgentry:
FYI @callmemc, this and some other Pro-specific changes will be in the next bugfix release to significantly reduce the likelihood of those transaction collision errors you were seeing all over the place yesterday. Will ping about it when shipped!

PR #1030, opened 2025-09-07
stronger test coverage around JobDeleteMany

bgentry:
While researching a customer issue I wanted to ensure that we had strong coverage so we can be confident that jobs we expect to _not_ be deleted are in fact not deleted.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -2145,15 +2145,27 @@ func Test_Client_JobDeleteMany(t *testing.T) {
 			job1 = testfactory.Job(ctx, t, bundle.exec, &testfactory.JobOpts{Schema: bundle.schema})
 			job2 = testfactory.Job(ctx, t, bundle.exec, &testfactory.JobOpts{Schema: bundle.schema})
 			job3 = testfactory.Job(ctx, t, bundle.exec, &testfactory.JobOpts{Schema: bundle.schema})
+			job4 = testfactory.Job(ctx, t, bundle.exec, &testfactory.JobOpts{Schema: bundle.schema})
 		)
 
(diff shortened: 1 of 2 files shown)


defany:
Heya! I'm thinking it's going to be easier to continue our conversation here) I was trying to research code as well and it looks strange for me, that you have an `Where` field inside `JobListParams` struct, but you do not put it in here 

Screenshot 2025-09-07 at 19 52 28

For example, you are doing this in `JobList` method, so I guess it may be a problem

bgentry:
@defany yeah, I think that's intentional here. There are currently fewer filtering options exposed for `JobDeleteMany` compared to listing. The standard structured fields like IDs get handled directly in a safer way without having to rely on more free-form where string injection.

The thing I found when researching this is that we had pretty strong coverage already on these methods, with the exception being that we often weren't testing to see if the "survivor" rows were in fact being left around after others were deleted. That coverage was added in this PR and so far I've been unable to expose any situations where it fails.

In your case, I assume you've already confirmed that you're not dealing with some kind of cross transaction visibility issues where some of the queries are in different txns and can't see each others' changes? That's really the only other idea I had.

defany:
Heya! I think it's going to be easier to continue conversation here

So, I've found why tests are going normally and problem as well, so:
```go
// You are making an copy of struct every time when user calls an method, so:
params := river.NewJobDeleteManyParams()
params.IDs(putResult.Job.ID)

_, err = client.JobDeleteMany(ctx, params)
if err != nil {
	panic(err)
}
```

In this case p.ids is going to be empty 

```go
// You are making an copy of struct every time when user calls an method, so:
params := river.NewJobDeleteManyParams()
params = params.IDs(putResult.Job.ID)

_, err = client.JobDeleteMany(ctx, params)
if err != nil {
	panic(err)
}
```

Everything is going fine

PR #1033, opened 2025-09-09
Require use of `JobDeleteManyParams.UnsafeAll` to delete all jobs

### LINKED ISSUE #1031: Non-mutable builder pattern for JobDeleteMany might be dangerous
We reused a lot of the implementation from the `JobList` API when implementing `JobDeleteMany`. One fairly dangerous outcome of this impacted @defany in https://github.com/riverqueue/river/issues/1030#issuecomment-3263986233_

Specifically, if you either don't save the result of the `.IDs()` call or just omit the line altogether, you'll accidentally delete _all_ non-running jobs. While returning more jobs than expected is annoying but not critical when listing jobs, when deleting them it's potentially very bad.

We should consider what options we have for avoiding footguns with this API.

brandur:
Addresses #1031. It's come to light that it's relatively easy to
accidentally delete all non-running jobs by invoking `JobDeleteMany`
with `nil` parameters or parameters with all defaults.

Here, make it a little harder to do so by requiring an invocation to
`JobDeleteManyParams.UnsafeAll` before a blanket delete is allowed.
Calling `UnsafeAll` isn't necessary if any filters are applied.

Fixes #1031.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Set minimum Go version to Go 1.24. [PR #1032](https://github.com/riverqueue/river/pull/1032).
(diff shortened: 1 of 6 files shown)


bgentry (approved):
Yeah this is roughly what I had in mind! I think this is about as good as we can do with the param builder pattern.

@defany do you have any thoughts before we ship this?

PR #1037, opened 2025-09-11
Cleanly handle jobs being snoozed with zero duration

brandur:
As I was going through to see if we could recommend snoozing jobs with a
zero duration as a way of returning an error that'd cause them to be
retried immediately but without logging error noise, I found that
snoozes with zero duration do work, but had a bug in the subscription
manager that'd cause them to be emitted as `EventKindJobFailed` instead
of `EventKindJobSnoozed`

Here, go through and fix that problem, and making sure to add adequate
test coverage for it.

I also add a formal documentation to `river.JobSnooze` that suggests the
function can be used for a noise-less retry even in cases where no
snooze time is required.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -496,6 +496,37 @@ func Test_Client_Common(t *testing.T) {
 		require.WithinDuration(t, time.Now().Add(15*time.Minute), updatedJob.ScheduledAt, 2*time.Second)
 	})
 
+	t.Run("JobSnoozeWithZeroDurationSetsAvailableImmediately", func(t *testing.T) {
+		t.Parallel()
+
+		client, _ := setup(t)
+
+		type JobArgs struct {
+			testutil.JobArgsReflectKind[JobArgs]
+		}
+
+		AddWorker(client.config.Workers, WorkFunc(func(ctx context.Context, job *Job[JobArgs]) error {
+			return JobSnooze(0)
+		}))
+
+		subscribeChan := subscribe(t, client)
+		startClient(ctx, t, client)
+
+		insertRes, err := client.Insert(ctx, &JobArgs{}, nil)
+		require.NoError(t, err)
+
+		event := riversharedtest.WaitOrTimeout(t, subscribeChan)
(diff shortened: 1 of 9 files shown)


client_test.go
```diff
@@ -496,6 +496,37 @@ func Test_Client_Common(t *testing.T) {
...
+		}))
+
+		subscribeChan := subscribe(t, client)
+		startClient(ctx, t, client)
+
+		insertRes, err := client.Insert(ctx, &JobArgs{}, nil)
+		require.NoError(t, err)
+
+		event := riversharedtest.WaitOrTimeout(t, subscribeChan)
+		// require.Equal(t, EventKindJobSnoozed, event.Kind)
```
bgentry:
```suggestion
		require.Equal(t, EventKindJobSnoozed, event.Kind)
```

internal/jobcompleter/job_completer_test.go
```diff
@@ -740,15 +741,46 @@ func testCompleter[TCompleter JobCompleter](
...
+		// one item each while the batch completer will send both. Put
+		// otherwise, expect the sync and async completers to iterate this loop
+		// twice and batch completer to iterate it twice.
+		var jobUpdates []CompleterJobUpdated
+		for {
+			jobUpdates = append(jobUpdates, riversharedtest.WaitOrTimeout(t, bundle.subscribeCh)...)
+			if len(jobUpdates) >= 2 {
+				break
+			}
+		}
```
bgentry: Yeah I guess you essentially want to flatten or fan out the job updates and wait for 2 total jobs rather than a specific number of batches. Probably not worth building a special helper for since we may not need it elsewhere.

client_test.go
brandur: Ah, whoops.

internal/jobcompleter/job_completer_test.go
brandur: Yeah exactly. Seems fixable, but just not worth it for this particular thing.

PR #1046, opened 2025-09-27
Minor grammar fix on `Pilot.JobCleanerQueuesExcluded`

brandur:
There was an extra word before ("as") that made this doc's grammar
wrong, but it was also just written in such a way that it was difficult
to parse. Here, do a bit of a rewrite as a fix up.

### DIFF
--- a/rivershared/riverpilot/pilot.go
+++ b/rivershared/riverpilot/pilot.go
@@ -20,8 +20,10 @@ type Pilot interface {
 	PilotPeriodicJob
 
 	// JobCleanerQueuesExcluded returns queues that should be excluded from the
-	// main River client's JobCleaner. Empty omitted queues should return nil as
-	// instead of empty array.
+	// main River client's JobCleaner. If no queues should be omitted, this
+	// function should return nil as opposed to an empty array. (Because the
+	// underlying database query uses an `IS NULL` check, though this could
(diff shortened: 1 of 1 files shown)


rivershared/riverpilot/pilot.go
```diff
@@ -20,8 +20,8 @@ type Pilot interface {
 	PilotPeriodicJob
 
 	// JobCleanerQueuesExcluded returns queues that should be excluded from the
-	// main River client's JobCleaner. Empty omitted queues should return nil as
-	// instead of empty array.
+	// main River client's JobCleaner. If no queues should be omitted, this
+	// function should return nil as opposed to an empty array.
```
bgentry: What's the difference in behavior if I _do_ return an empty array? Maybe worth spelling out if anything but I would expect any logic depending on this to consider them equivalent?

bgentry (approved):

rivershared/riverpilot/pilot.go
brandur: K I added a little more text to that.

The reason is that we check on empty in the underlying database query and unlike Go, a `null` array and a zero-length array or _not_ equivalent in Postgres. We could check on both those conditions, but I figured it'd be better to constrain our expectations so we don't pollute the SQL that much.

PR #1047, opened 2025-09-29
JobRescuer: set the rescue count as an int in metadata

bgentry:
When rescuing a "stuck" job, set or increment a `river:rescued` metadata key on the job to indicate that it was rescued by the rescuer.

### DIFF
--- a/internal/rivercommon/river_common.go
+++ b/internal/rivercommon/river_common.go
@@ -24,6 +24,9 @@ const (
 	// them.
 	MetadataKeyPeriodicJobID = "river:periodic_job_id"
 
+	// MetadataKeyRescueCount records how many times the job has been rescued.
+	MetadataKeyRescueCount = "river:rescue_count"
+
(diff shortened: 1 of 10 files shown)


internal/rivercommon/river_common.go
```diff
@@ -24,6 +24,9 @@ const (
 	// them.
 	MetadataKeyPeriodicJobID = "river:periodic_job_id"
 
+	// MetadataKeyRescued records how many times the job has been rescued.
+	MetadataKeyRescued = "river:rescued"
```
brandur: Nit, but just given that it's a counter, I wonder if something like `num_rescues` would be a little clearer? ("rescued" sounds like a true/false bool to me anyway.)

bgentry: Good call, how about `river:rescue_count`? Feels slightly easier to read and keeps the primary noun first. Updated it with that.

PR #1048, opened 2025-09-30
extract composable riverdbtest.TestTxWithDriver helper

bgentry:
The existing `TestTxPgx` forced a specific non-Pro driver while also holding useful internal logic around setting a default search path. To make this more reusable, expose a far more flexible `TestTxWithDriver` variant (used by `TestTxPgx`) which takes a custom driver, allows custom opts, and returns both the tx and schema—all while still setting the default `search_path` so that others can avoid reimplementing those parts.

Being leveraged by riverqueue/riverui#435.

### DIFF
--- a/riverdbtest/riverdbtest.go
+++ b/riverdbtest/riverdbtest.go
@@ -428,22 +428,36 @@ func packageFromFunc(funcName string) string {
 	return packageName
 }
 
-// TestTx starts a test transaction that's rolled back automatically as the test
-// case is cleaning itself up.
+// TestTxPgx starts a test transaction that's rolled back automatically as the
+// test case is cleaning itself up.
 //
 // This variant starts a transaction for the standard pgx/v5 driver most
 // commonly used throughout most of River.
 func TestTxPgx(ctx context.Context, tb testing.TB) pgx.Tx {
 	tb.Helper()
+	tx, _ := TestTxPgxDriver(ctx, tb, riverpgxv5.New(riversharedtest.DBPool(ctx, tb)), nil)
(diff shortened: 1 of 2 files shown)


riverdbtest/riverdbtest.go
```diff
@@ -428,22 +428,36 @@ func packageFromFunc(funcName string) string {
...
+}
 
-	tx, schema := TestTx(ctx, tb, riverpgxv5.New(riversharedtest.DBPool(ctx, tb)), &TestTxOpts{
-		IsTestTxHelper: true,
-	})
+// TestTxWithDriver starts a test transaction that's rolled back automatically
+// as the test case is cleaning itself up. Unlike TestTxPgx, this variant takes
+// a driver and options for greater flexibility, including allowing for Pro
+// drivers, while still sharing common setup like schema search path.
+func TestTxWithDriver(ctx context.Context, tb testing.TB, driver riverdriver.Driver[pgx.Tx], opts *TestTxOpts) (pgx.Tx, string) {
```
brandur: What do you think about calling this `TestTxPgxDriver` instead?

The "raw" `TestTx` below is really the most generic sort of "test tx with driver". The function you're adding puts in a `search_path`, which is a Postgres-specific feature, so it's more like we're trying to start a test tx, but on a specific Postgres driver like `riverpgx` or its Pro variant.

bgentry: Heh, I actually had that before and then changed it because it felt weird to say “pgx but with an arbitrary driver”. Easy to change back!

PR #1051, opened 2025-10-01
Don't convert errors to strings before logging

bradleyjkemp:
Currently by calling `.Error()` on job worker errors before logging them, data is being thrown away that cannot be used by slog handlers.

I use a structured error library where errors can contain metadata like status codes or user IDs. In my slog handler I can extract these values and e.g. send this on to an error tracking service

But, this metadata isn't present in the basic `.Error()` output and so it's being thrown away by this line which converts the error to a string.

Instead I think you can use the `slog.Any` attribute type instead, by default this will have the same behaviour as the current code (printing out the error text), but allows custom slog handlers to extract the original error type


bgentry (approved):
Looks good, thanks!

PR #1062, opened 2025-10-22
Add `RollbackWithoutCancel` helper to have rollbacks supersede context cancellation

### LINKED ISSUE #1059: High deadlocking by River causing database to lock up
We are using River to process a relatively small number of jobs (maybe a few hundred an hour) and a bunch of them are periodic (e.g., every few hours, once a day, once a week, etc). After a few hours of operation, the database is eventually starved of connections, and the application becomes in accessible to users. Manually failing over the database appears to sever all active connections and remove all locks, allowing the application to reconnect and start working again. After a few hours (or days depending on the load), the locks will build up again and the application will become inaccessible.

Image

The "1" indicates when the database was failed over to this instance. You can see the locks build up until it reaches the max and starts blocking connections. The "2" indicates when the database was failed over again.

Image

RDS shows that a River-managed job is responsible for locks. I'm not an SQL expert so it's hard to figure out why this could be causing issues. It's always this "JobSchedule" statement, even when we're not actually scheduling anything. River UI shows no scheduled jobs and very few jobs running and it still locks up:

Image

The database is RDS Aurora Postgres (16.6) and has an RDS proxy sitting in front of it. We also have two instances of River running on the same database, separated by different schema names. I can't imagine this would cause this issue though.

I found the following issues which seem relevant:

* https://github.com/riverqueue/river/issues/59
* https://github.com/riverqueue/river/issues/995

I had a feeling that the auto-vacuum starvation was relevant because if I manually shell into the database and run `VACUUM` it hangs forever.

We are using River v0.25. I thought about upgrading to v0.26 in case it included a fix, but I couldn't see anything in the diff that seemed relevant.

Any advice would be much appreciated as at the moment we're having to frequently fail over the database to keep the application accessible.

brandur:
Here, add a `dbutil.RollbackWithoutCancel` helper. The purpose of this
is in situations where an operation is in a transaction but ends up
cancelling due to a context timeout. From there, there tends to be a
rollback in a `defer`, but the rollback ends up not running because its
context is already cancelled.

The new helper removes a cancelled context, then adds a new 5 second
context timeout to run the rollback command. This is mainly used for
`JobSchedule` in the hopes of coming up with a fix for #1059, but we use
the new function everywhere that we previously had a manual rollback,
and it also gets add to the `dbutil.WithTx*` helpers so that it gets
used anywhere those do (and these should probably be preferred most of
the time because it makes rollbacks and commits harder to forget).

Fixes #1059.

### DIFF
--- a/rivermigrate/river_migrate_test.go
+++ b/rivermigrate/river_migrate_test.go
@@ -27,10 +27,9 @@ import (
 
 const (
 	// The name of an actual migration line embedded in our test data below.
-	migrationLineAlternate                = "alternate"
-	migrationLineAlternateMaxVersion      = 6
-	migrationLineCommitRequired           = "commit_required"
-	migrationLineCommitRequiredMaxVersion = 3
+	migrationLineAlternate           = "alternate"
+	migrationLineAlternateMaxVersion = 6
+	migrationLineCommitRequired      = "commit_required"
 )
 
 //go:embed migration/*/*.sql
--- a/rivershared/util/dbutil/db_util.go
+++ b/rivershared/util/dbutil/db_util.go
@@ -3,12 +3,44 @@ package dbutil
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/riverqueue/river/riverdriver"
 )
 
+// RollbackWithoutCancel initiates a rollback, but one in which context is
+// overridden with context.WithoutCancel so that the rollback can proceed even
+// if a previous operation was cancelled. This decreases the chance that a
+// transaction is accidentally left open and in an ambiguous state.
+//
+// A rollback error is returned the same way most driver rollbacks return an
+// error, but given this is normally expected to be used in a defer statement,
+// it's unusual for the error to be handled.
+func RollbackWithoutCancel[TExec riverdriver.ExecutorTx](ctx context.Context, execTx TExec) error {
+	ctxWithoutCancel := context.WithoutCancel(ctx)
+
+	ctx, cancel := context.WithTimeout(ctxWithoutCancel, 5*time.Second)
+	defer cancel()
+
+	// It might not be the worst idea to log an unexpected error on rollback
+	// here instead of returning it. I had this in place initially, but there's
+	// a number of common errors that need to be ignored like "conn closed",
+	// `pgx.ErrTxClosed`, or `sql.ErrTxDone`. These all turn out to be
+	// driver-specific when this function is meant to be driver agnostic.
+	//
+	// It'd still be possible to make it happen, but we'd have to have a driver
+	// function like `ShouldIgnoreRollbackError` that'd need a lot of plumbing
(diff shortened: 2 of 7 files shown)


rivermigrate/river_migrate_test.go
```diff
@@ -27,10 +27,9 @@ import (
 
 const (
 	// The name of an actual migration line embedded in our test data below.
-	migrationLineAlternate                = "alternate"
-	migrationLineAlternateMaxVersion      = 6
-	migrationLineCommitRequired           = "commit_required"
-	migrationLineCommitRequiredMaxVersion = 3
```
brandur, line 33: Unrelated change, but linter started complaining this was unused.

rivershared/util/dbutil/db_util.go
```diff
@@ -3,12 +3,44 @@ package dbutil
...
+// transaction is accidentally left open and in an ambiguous state.
+//
+// A rollback error is returned the same way most driver rollbacks return an
+// error, but given this is normally expected to be used in a defer statement,
+// it's unusual for the error to be handled.
+func RollbackWithoutCancel[TExec riverdriver.ExecutorTx](ctx context.Context, execTx TExec) error {
+	ctxWithoutCancel := context.WithoutCancel(ctx)
+
+	ctx, cancel := context.WithTimeout(ctxWithoutCancel, 5*time.Second)
+	defer cancel()
```
bgentry, line 23: super nit but I might prefer to inline this as it tends to be what I see in Go code in the wild, up to you though

```suggestion
	ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 5*time.Second)
	defer cancel()
```

rivershared/util/dbutil/db_util.go
```diff
@@ -18,14 +50,20 @@ func WithTx[TExec riverdriver.Executor](ctx context.Context, exec TExec, innerFu
...
 	var defaultRes T
 
 	tx, err := exec.Begin(ctx)
 	if err != nil {
 		return defaultRes, fmt.Errorf("error beginning transaction: %w", err)
 	}
-	defer tx.Rollback(ctx)
+	defer func() {
+		_ = RollbackWithoutCancel(ctx, tx)
+	}()
```
bgentry: this seems to be the only one wrapped in a func closure, intentional?

bgentry (approved):

rivershared/util/dbutil/db_util.go
brandur: Hmm, I kinda like how having the two separate statements lets you easily see each step. It's trivially easy (and I'd say default even) to miss the cancel when they're about wrapped together like that.

rivershared/util/dbutil/db_util.go
brandur: Oh yeah, let me comment this. For some reason this is the one place where the linter requires handling the error of `RollbackWithoutCancel`. Maybe because it's in the same package? I don't 100% get it.

PR #1065, opened 2025-10-27
Replace `SlogMessageOnlyHandler` with attribute replacement scheme

brandur:
Here, replace with the utility `SlogMessageOnlyHandler` that we use in
example tests with a `ReplaceAttr` implementation instead so that we use
a standard `TextHandler` with a `ReplaceAttr`:

``` diff
-		Logger: slog.New(&slogutil.SlogMessageOnlyHandler{Level: slog.LevelWarn}),
+		Logger: slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelWarn, ReplaceAttr: slogutil.NoLevelTimeJobID})),
```

The major benefit of this approach is that in case of an error (where
the error detail gets put in a key of the output log), we don't end up
hiding the detail and making the error difficult to debug without
removing the message-only handler.

It also has the nominal advantage of being a little more standard, i.e.
Uses normal `slog` handlers instead of a custom one, and is a little
more succinct to write.

### DIFF
--- a/example_batch_insert_test.go
+++ b/example_batch_insert_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"os"
 
 	"github.com/jackc/pgx/v5/pgxpool"
 
@@ -45,7 +46,7 @@ func Example_batchInsert() {
 	river.AddWorker(workers, &BatchInsertWorker{})
 
 	riverClient, err := river.NewClient(riverpgxv5.New(dbPool), &river.Config{
-		Logger: slog.New(&slogutil.SlogMessageOnlyHandler{Level: slog.LevelWarn}),
+		Logger: slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelWarn, ReplaceAttr: slogutil.NoLevelTime})),
 		Queues: map[string]river.QueueConfig{
 			river.QueueDefault: {MaxWorkers: 100},
 		},
--- a/riverlog/example_new_middleware_test.go
+++ b/riverlog/example_new_middleware_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"os"
 
 	"github.com/jackc/pgx/v5/pgxpool"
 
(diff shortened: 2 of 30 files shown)


example_batch_insert_test.go
```diff
@@ -45,7 +46,7 @@ func Example_batchInsert() {
 	river.AddWorker(workers, &BatchInsertWorker{})
 
 	riverClient, err := river.NewClient(riverpgxv5.New(dbPool), &river.Config{
-		Logger: slog.New(&slogutil.SlogMessageOnlyHandler{Level: slog.LevelWarn}),
+		Logger: slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelWarn, ReplaceAttr: slogutil.NoLevelTime})),
```
bgentry, line 49: I might prefer to still wrap this in a helper to a) avoid potential drift as examples are added/updated and b) allow some space for a comment about _why_ this uses `ReplaceAttr` which most users won't want/need. Up to you though, I do see the advantage of not obscuring the example code behind more indirection.

riverlog/example_new_middleware_test.go
```diff
@@ -48,18 +49,18 @@ func ExampleNewMiddleware() {
...
 		Middleware: []rivertype.Middleware{
 			riverlog.NewMiddleware(func(w io.Writer) slog.Handler {
-				// We have to use a specialized handler without timestamps to
-				// make test output reproducible, but in reality this would as
-				// simple as something like:
+				// We have to use a specialized ReplacedAttr without level or
+				// timestamps to make test output reproducible, but in reality
+				// this would as simple as something like:
 				//
 				// 	return slog.NewJSONHandler(w, nil)
```
bgentry, line 62: yeah this is what I was getting at with context for why we have special attrs passed in, seems helpful here though it adds a few lines

example_batch_insert_test.go
brandur: Yeah, let's leave for now. I kinda like how this just shows raw options instead of hiding things in a helper, and it's really only three options it's configuring total, with 2/3 very standard. The comment on why `ReplaceAttr` is on `slogutil.NoLevelTime` in case anyone cares to jump-to and read it.

Issue #1017, opened 2025-08-15
Feature Request:`JobCount()` client function


danieljosephchambers:
There's a nice `JobList` method on the client that can be used to get back a slice of all jobs, according to some provided params (kind, state etc).

I've been using this - but just to get a count of the number of jobs of each kind in each state (I've made a prometheus metrics Collector that keeps track of the number of jobs by `kind` and `state`). 

This is quite a heavy query as it's returning the whole `JobRow` - which includes a bunch of fields I don't really care about for this narrow purpose.

It'd be great if there were a lighter weight `JobCount` method that would just return a count of the number of jobs according to the parameters.

PR #1068, opened 2025-10-30
Add JobCountByAllStates method to Client

jcass8695:
Addresses #1017. In our organisation we would like to be able to have a regular metric published with the count of jobs in each state for monitoring purposes.

### DIFF
--- a/client.go
+++ b/client.go
@@ -2275,6 +2275,18 @@ func (c *Client[TTx]) JobListTx(ctx context.Context, tx TTx, params *JobListPara
 	return res, nil
 }
 
+// JobCountByAllStates returns a map of counts for all job states.
+// The provided context is used for the underlying database query.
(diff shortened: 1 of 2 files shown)


brandur:
We don't normally expose driver functions like this without strong justification. In this case, counting can somewhat unintuitively become a bit of a problem because it can be very slow when dealing with enormous numbers of rows.

jcass8695:
> We don't normally expose driver functions like this without strong justification. In this case, counting can somewhat unintuitively become a bit of a problem because it can be very slow when dealing with enormous numbers of rows.

Hey @brandur, thanks for taking a look at this change so promptly. The motivation is that we would like to be able to publish a metric on the number of jobs in each state. If this isn't something that you and the other maintainers would like to expose through the client then it's not something critically important for us but it would be a nice to have. Lmk your thoughts.

PR #1069, opened 2025-11-08
Add `UnknownConfigure` function for durable periodic jobs

brandur:
Here, add an `UnknownConfigure` function to the pilot for use with
rescuing and reconfiguring unknown durable periodic jobs. Full
explanation in the counterpart pull request.


periodic_job_test.go
```diff
@@ -62,7 +63,7 @@ func TestPeriodicJobBundle(t *testing.T) {
 
 		periodicJob := NewPeriodicJob(
 			PeriodicInterval(15*time.Minute),
-			func() (JobArgs, *InsertOpts) {
+			func() (river.JobArgs, *river.InsertOpts) {
```
bgentry: Can you revert this diff so the file doesn't need a named import for its own package? Seems like that'll fix CI issues.

client.go
```diff
@@ -2709,3 +2736,70 @@ func defaultClientIDWithHost(startedAt time.Time, host string) string {
...
+
+// uniqueOptsIsEmpty returns true for an empty, uninitialized options struct.
+//
+// This is required because we can't check against `UniqueOpts{}` because slices
+// aren't comparable. Unfortunately it makes things a little more brittle
+// comparatively because any new options must also be considered here for things
+// to work.
+//
+// This is a function instance of an instance function so that it doesn't have
+// to be exported from `rivertype` and doesn't become part of the public API.
```
bgentry: Not sure if I'm just having trouble reading this in its current form, but did you mean for something like this?

```suggestion
// This is an unexported function in `river` so that it doesn't have
// to be exported from `rivertype` and doesn't become part of the public API.
```

client.go
```diff
@@ -2709,3 +2736,70 @@ func defaultClientIDWithHost(startedAt time.Time, host string) string {
...
+	rivertype.JobStateAvailable,
+	rivertype.JobStatePending,
+	rivertype.JobStateRunning,
+	rivertype.JobStateScheduled,
+}
+
+// uniqueOptsValidate validates the given rivertype.UniqueOpts.
+//
+// This is a function instance of an instance function so that it doesn't have
+// to be exported from `rivertype` and doesn't become part of the public API.
```
bgentry, line 2771: same comment re: wording on this

job.go
```diff
@@ -13,28 +13,12 @@ type Job[T JobArgs] struct {
...
-	// which worker should work the old kind. Job kinds can be renamed safely
-	// over multiple deploys using the JobArgsWithKindAliases interface.
-	Kind() string
-}
+type (
+	InsertOpts = rivertype.InsertOpts
+	UniqueOpts = rivertype.UniqueOpts
+)
+
+type JobArgs = rivertype.JobArgs
```
bgentry: Do we lose anything on editor-friendliness here, either with the alias or with the removal of doc comments? Went to confirm this myself and it looks like we should definitely keep type-level comments in both places, otherwise you see this:

Screenshot 2025-11-09 at 11 32 03 AM

So the aliased type has its internal details properly resolved, which is great, but there are no type-level comments to tell you what the type is or what it's for unless you keep them present here where the type alias is declared.

The same probably applies to the two types aliased above this too.

periodic_job.go
```diff
@@ -8,27 +8,16 @@ import (
...
-
-// PeriodicJobConstructor is a function that gets called each time the paired
-// PeriodicSchedule is triggered.
-//
-// A constructor must never block. It may return nil to indicate that no job
-// should be inserted.
-type PeriodicJobConstructor func() (JobArgs, *InsertOpts)
+type (
+	PeriodicJobConstructor = rivertype.PeriodicJobConstructor
+	PeriodicSchedule       = rivertype.PeriodicSchedule
```
bgentry: same re: preserving type-level doc comments on these for developer experience

client.go
brandur: Oops, yep. Thanks.

job.go
brandur: Hm, isn't the screenshot you posted exactly what you'd want? The LSP is showing you the alias, but since that's not very interesting, it also shows you the original type right inline. This DX strikes me as perfect — tells you everything you need to know in one place.

I added some additional comments to explain what happened with each of these:

```
	// JobArgs was originally in the top-level river package, but moved so it
	// could be accessible by subpackages. Prefer using the one in rivertype.
	JobArgs = rivertype.JobArgs
```

I'm kind of thinking that before release, we update all internal references in the project to `rivertype` (I didn't do that here because it'd make a huge diff), and add `Deprecated` to the aliases to encourage people to use the `rivertype` version (I didn't do that here because it'd cause lint to fail). Thoughts on that?

periodic_job_test.go
brandur: Yeah, good catch. Done.

PR #1070, opened 2025-11-08
Remove IDs from ID map when periodic jobs with IDs are removed

brandur:
While implementing another feature I found a little bug in the periodic
job enqueuer in which when removing periodic jobs that have IDs assigned
(i.e. for durable jobs), the IDs weren't being removed from the enqueuer's
internal ID map. Here, fix the problem.

### DIFF
--- a/internal/maintenance/periodic_job_enqueuer.go
+++ b/internal/maintenance/periodic_job_enqueuer.go
@@ -243,7 +243,9 @@ func (s *PeriodicJobEnqueuer) Remove(periodicJobHandle rivertype.PeriodicJobHand
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	delete(s.periodicJobs, periodicJobHandle)
+	if periodicJob, ok := s.periodicJobs[periodicJobHandle]; ok {
+		s.removeJobLockFree(periodicJob, periodicJobHandle)
+	}
 }
 
 // RemoveMany removes many periodic jobs from the enqueuer. Their current target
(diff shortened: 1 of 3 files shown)


internal/maintenance/periodic_job_enqueuer.go
```diff
@@ -243,7 +243,10 @@ func (s *PeriodicJobEnqueuer) Remove(periodicJobHandle rivertype.PeriodicJobHand
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	delete(s.periodicJobs, periodicJobHandle)
+	if periodicJob, ok := s.periodicJobs[periodicJobHandle]; ok {
+		delete(s.periodicJobs, periodicJobHandle)
+		delete(s.periodicJobIDs, periodicJob.ID)
```
bgentry: Although it creates some indirection, this is the kind of scenario with needing to update multiple things upon deletion is one where I'd typically extract these into a shared function. This guarantees the two call sites use the same logic for manipulating these maps. Otherwise I fear drift if one is updated and the other is overlooked. Something like:

```go

func (s *PeriodicJobEnqueuer) removeLocked(periodicJobHandle rivertype.PeriodicJobHandle) {
// both deletes
}
```

Which you can then call in either location. The compiler will probably inline it and IMO the reduced risk of concurrency bugs outweighs the small cost of indirection.

Your call though! The tests at least provide us some protection.

brandur: K, yeah the main reason I don't do this for trivial functions is that the indirection makes code harder to read/follow as you add another jump to things.

I suppose it doesn't matter too much here though. Added helper.

brandur:
Thanks!

PR #1078, opened 2025-11-13
Add client mechanism to tell leader to remove periodic from any client

brandur:
This is once again in pursuit of solving a particular user's use case.
We currently have mechanisms that to add and remove periodic jobs to a
client, but these functions are only effective on the client which is
elected leader. In a multi-node system this can make things difficult
because it means that all operations have to be run on every known
client.

Here, introduce a system whereby the periodic job enqueuer starts
listening on a Postgres channel and we give the client a way of
notifying.

When it comes to sending notification, the most obvious place to send a
removal would be when you do a `client.PeriodicJobs().RemoveByID(...)`.
However, I'm reluctant to add a notification there because Postgres
notifications have such specific semantics about being runnable in or
out of a transaction. If we sent it in `RemoveByID` then we'd also need
to add a `RemoveByIDTx`, and things really get messy quickly from there.

So here I'm proposing something much more generic, which is the addition
of a top level `Notify` to the River client. So we get:

    client.Notify(ctx, NotifyKindPeriodicJobRemove, "my_periodic_job")

The idea is that we make this generic enough so that we could add other
sorts of notifications in the future (brought in very carefully on a
case by case basis). The payload argument is of type `any` so that it
could take specific multi-field structs for more complex operations.

### DIFF
--- a/client.go
+++ b/client.go
@@ -903,6 +903,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 			periodicJobEnqueuer, err := maintenance.NewPeriodicJobEnqueuer(archetype, &maintenance.PeriodicJobEnqueuerConfig{
 				AdvisoryLockPrefix: config.AdvisoryLockPrefix,
 				Insert:             client.insertMany,
+				Notifier:           client.notifier, // may be nil for poll-only driver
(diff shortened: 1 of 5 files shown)


brandur:
Closing in favor of #1085.

Issue #1077, opened 2025-11-13
leadership.Elector: Error attempting reelection


fvbock:
I ran into an issue recently - i think i saw this around 5-7 days ago for the first time - that in a single node dev setup leader election would fail and then endlessly trying to attempt reelection.

In the DB i can see the leader and it happens right after its expiry time:

```
# select * from river_leader;
          elected_at           |          expires_at           |            leader_id            |  name   
-------------------------------+-------------------------------+---------------------------------+---------
 2025-11-13 02:18:05.071551+00 | 2025-11-13 04:57:12.851001+00 | tutu_2025_11_13T02_18_05_067641 | default
(1 row)
```

after a successful job run leader elections starts and fails and keeps trying:

```
time=2025-11-13T13:57:00.080+09:00 level=INFO msg="producer: Producer job counts" num_completed_jobs=160 num_jobs_running=1 queue=default
time=2025-11-13T13:57:07.853+09:00 level=ERROR msg="leadership.Elector: Error attempting reelection" attempt=1 client_id=tutu_2025_11_13T02_18_05_067641 err="error beginning transaction: context deadline exceeded" sleep_duration=1.000562712s
time=2025-11-13T13:57:18.856+09:00 level=ERROR msg="leadership.Elector: Error attempting reelection" attempt=2 client_id=tutu_2025_11_13T02_18_05_067641 err="error beginning transaction: context deadline exceeded" sleep_duration=2.022818933s
```

NB: the db time is UTC the logs are logged with UTC+9

After that all job scheduling fails and the election error keeps coming up. no other messages. 

Any pointers would be greatly appreciated.

Thanks for the great library/tool!

brandur:
> is there a recommended minimum pool size to reserve for river? this pool is also shared with the app.

I thought we'd written some docs on this but I couldn't find it. Rough guidelines are one connection per client for listen/notify, plus one connection for each queue for each client. So for 3 queues and 5 clients it'd be roughly:

(1 + 3) * 5 = 20

There are some other queries that come and go though, so this isn't exact number. It may briefly go higher or lower.

> after it failed once it would keep failing "for ever" (saw something like attempt 100-something... shouldn't it at least eventually go through?

Yeah, that is odd. It could be something else I suppose. Worth checking — are you using any non-Postgres technology like Supabase or something?

> i will check pg_stat_activity for any stuck txns the next time i run into the issue.

Yep, that'd be good info.

> should this be max the max conn size of the connection pool?

[...]

PR #1079, opened 2025-11-14
Decrease `serviceutil.MaxAttemptsBeforeResetDefault`

bgentry:
In [#1077](https://github.com/riverqueue/river/issues/1077#issuecomment-3528161568), I noticed this log line:

> ```
> time=2025-11-13T23:39:08.589+09:00 level=ERROR msg="leadership.Elector: Error attempting to elect" attempt=100 client_id=tutu_2025_11_13T09_03_12_779431 err="error beginning transaction: context deadline exceeded" sleep_duration=8m21.535638673s
> ```

Specifically the `sleep_duration=8m21.535638673s` part. I don't think it makes much sense for the leader elector to be this lax on its retry attempts. Looking at where else we use `serviceutil.ExponentialBackoff` with its default `MaxAttemptsBeforeDefault` of 10, I spotted other elector calls as well as the job completer and notifier—none of which really make sense to let sleep for >8 minutes between attempts.

To resolve this I decreased this default max attempts before reset from from 10 to 7 lowering the effective limit on most internal exponential backoffs from ~512 seconds to 64 seconds. Further lowered the leader elector's keep leadership backoff interval to cap out at 4 seconds since leadership without a successful heartbeat will be lost soon after that anyway.

While this will result in some more noise and load in prolonged severely degraded scenarios, it also means faster recovery.

### DIFF
--- a/rivershared/util/serviceutil/service_util.go
+++ b/rivershared/util/serviceutil/service_util.go
@@ -44,7 +44,9 @@ func CancellableSleepC(ctx context.Context, sleepDuration time.Duration) <-chan 
 // into a ridiculously distant future. This constant is typically injected into
 // the CancellableSleepExponentialBackoff function. It could technically take
 // another value instead, but shouldn't unless there's a good reason to do so.
-const MaxAttemptsBeforeResetDefault = 10
+//
+// The value of 7 corresponds to a max sleep duration of 64 seconds ±10% jitter.
+const MaxAttemptsBeforeResetDefault = 7
 
 // ExponentialBackoff returns a duration for a reasonable exponential backoff
 // interval for a service based on the given attempt number, which can then be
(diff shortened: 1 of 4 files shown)


bgentry:
@brandur let me know if you disagree on any of these, but at least for the leader elector it seemed like a bad idea to let it sleep for over 8 minutes between attempts.

rivershared/util/serviceutil/service_util.go
```diff
@@ -44,14 +44,23 @@ func CancellableSleepC(ctx context.Context, sleepDuration time.Duration) <-chan
 // into a ridiculously distant future. This constant is typically injected into
 // the CancellableSleepExponentialBackoff function. It could technically take
 // another value instead, but shouldn't unless there's a good reason to do so.
-const MaxAttemptsBeforeResetDefault = 10
+const MaxAttemptsBeforeResetDefault = 7
```
brandur, line 49: Might make sense to call out the max sleep here so the choice of 7 is somewhat justified.

PR #1080, opened 2025-11-14
Prepare v0.27.0

brandur:
We haven't done a release in a while, so here prepare v0.27.0 to go out
the door with the last few months worth of changes and bug fixes.

[skip ci]


brandur:
Thx. I pushed the latest couple bug fixes into here too.

PR #1081, opened 2025-11-15
Stop subservices and embedded `baseservice.Service` on periodic jobs enqueuer start error

brandur:
Just a small one where in the very rare event of a start error from the
periodic jobs enqueuer, make sure to stop all subservices and stop the
`baseservice.Service`. In practice this won't matter much because
there's only one subservice currently that can error, but it might
matter in the future in case more subservices are added.


bgentry (approved):
Good call :pray:

brandur:
ty.

PR #1082, opened 2025-11-15
Rename `context.go` to `client_context.go`

brandur:
Here, rename `context.go` to `client_context.go`. The premise here is to
give it a less generic name so it's easier to find via quick file find
when working from memory (something I currently have trouble doing). In
case there's anything that lands in here not related to client-specific
context in the future, we can find a different name for the file, but
it's been only client-specific since inception so far.


brandur:
@bgentry You okay with this change?

bgentry (approved):

PR #1083, opened 2025-11-15
Clear ID map on periodic jobs bundle clear

brandur:
This is another bug similar to the one fixed in #1070. When clearing
periodic jobs we were only clearing the handle -> job map and not the
secondary map of periodic job IDs.


bgentry (approved):
Good find, thanks!

brandur:
thx.

PR #1084, opened 2025-11-16
Add hook for when the periodic job enqueuer starts up

brandur:
This one's an alternative (and hopefully a better one) to #1078. #1078
works for removing a periodic job from a leader, but doesn't cover the
case where a new periodic job might need to be added dynamically from
any possible client.

Here, we add the new hook `HookPeriodicJobsStart` which runs when the
periodic job enqueuer starts up on a newly elected leader. This gives
the hook quite a bit of power to dynamically adjust all currently
configured periodic jobs on the enqueuer. For example, it may clear all
of them and then add them back from scratch using a canonical set of
periodic job records from elsewhere which it knows should exist.

This will need to be paired with another feature that would allow users
to force a leader reelection (which I haven't written yet). So when they
determine that a periodic job overhaul is necessary, they'll knock the
current leader offline and then use `HookPeriodicJobsStart` to configure
the right set of periodic jobs.

A modification we make here is to make it possible to get a client from
context in the hook similar to how it's possible in a worker right now.
This is because the hook implementation will often want access to a
client instance, but it's very awkward getting access to a client
otherwise because you define your hooks before a client is available,
thereby creating a chicken and egg problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `riverlog.LoggerSafely` which provides a non-panic variant of `riverlog.Logger` for use when code may or may not have a context logger available. [PR #1093](https://github.com/riverqueue/river/pull/1093).
+
 ## [0.27.1] - 2025-11-21
 
 ### Fixed
--- a/riverlog/river_log.go
+++ b/riverlog/river_log.go
@@ -27,12 +27,25 @@ type contextKey struct{}
 // Logger extracts a logger from context from within the Work body of a worker.
 // Middleware must be installed on either the worker or client for this function
 // to be usable.
+//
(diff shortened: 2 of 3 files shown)


client_test.go
```diff
@@ -4838,6 +4838,54 @@ func Test_Client_Maintenance(t *testing.T) {
...
 	})
 
+	t.Run("PeriodicJobEnqueuerStartHook", func(t *testing.T) {
+		t.Parallel()
+
+		config := newTestConfig(t, "")
+
+		var periodicJobsStartHookCalled bool
+		config.Hooks = []rivertype.Hook{
+			HookPeriodicJobsStartFunc(func(ctx context.Context, params *rivertype.HookPeriodicJobsStartParams) error {
```
bgentry: Should we add coverage for when this hook returns an error? I guess it should block client start—this is the first place I can think of where we're doing something like that so we may want to be cautious and ensure proper coverage.

brandur: Yep, good call. Added the new test case over in #1096.

PR #1085, opened 2025-11-16
`Notify().*` functions for client + force resign mechanism

brandur:
This one's the counterpart to #1084: a mechanism that requests that the
current River leader resign. Users can use this to force a resignation
and then `HookPeriodicJobsStart` from #1084 to dynamically reconfigure
all periodic jobs the next time a leader is elected.

Introduces two new functions to the client:

* `client.Notify().RequestResign(ctx)`
* `client.Notify().RequestResignTx(ctx)`

The idea is that the "notify bundle" will be used to force resign for now, 
but could be expanded to other types of manual notifications that need
to be sent.

The alternative would be add something like `LeaderResign`, but it'd be
nice to have some naming that implies that we're _requesting_ via
Postgres listen/notify that the leader resign rather than that we're
specifically requesting that _this_ client resigns (because this client
may not be the leader). So maybe something like `NotifyLeaderResign`
could be plausible, but that's not particularly great either. The other
thing to remember is that whatever we choose, we need to have a *tx
variant like `NotifyLeaderResignTx`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -2316,10 +2316,6 @@ func (c *ClientNotifyBundle[TTx]) RequestResignTx(ctx context.Context, tx TTx) e
 // notifyExecTx is a shared helper between Notify and NotifyTx that sends a
 // notification.
 func (c *ClientNotifyBundle[TTx]) requestResignTx(ctx context.Context, execTx riverdriver.ExecutorTx) error {
-	if !c.driver.SupportsListenNotify() {
-		return errors.New("notify is only supported for drivers that support listen/notify (e.g. pgx, but not database/sql)")
-	}
-
 	payloadStr, err := json.Marshal(&leadership.DBNotification{
 		Action: leadership.DBNotificationKindRequestResign,
 	})
--- a/rivershared/util/serviceutil/service_util.go
+++ b/rivershared/util/serviceutil/service_util.go
@@ -62,7 +62,7 @@ const MaxAttemptsBeforeResetDefault = 7
 // - 8s
 // - 16s
 // - 32s
-// - 64s.
+// - 64s
 func ExponentialBackoff(attempt, maxAttemptsBeforeReset int) time.Duration {
 	retrySeconds := exponentialBackoffSecondsWithoutJitter(attempt, maxAttemptsBeforeReset)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Added `Client.Notify` and `Client.NotifyTx` functions allowing any client to request that the current leader resign. [PR #1085](https://github.com/riverqueue/river/pull/1085).
+- Added `Client.Notify().RequestResign` and `Client.Notify().RequestResignTx` functions allowing any client to request that the current leader resign. [PR #1085](https://github.com/riverqueue/river/pull/1085).
 
 ## [0.28.0] - 2025-11-23
 
--- a/client_test.go
+++ b/client_test.go
@@ -1127,25 +1127,6 @@ func Test_Client_Common(t *testing.T) {
 		client.config.Logger.InfoContext(ctx, "Client was elected leader after forced resignation")
 	})
 
-	t.Run("NotifyRequestResignNoListenNotifyError", func(t *testing.T) {
-		t.Parallel()
-
-		var (
-			dbPool = riversharedtest.DBPoolClone(ctx, t)
-			driver = NewDriverWithoutListenNotify(dbPool)
-			schema = riverdbtest.TestSchema(ctx, t, driver, nil)
-			config = newTestConfig(t, schema)
-		)
-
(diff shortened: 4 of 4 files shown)


client.go
```diff
@@ -2275,6 +2275,77 @@ func (c *Client[TTx]) JobListTx(ctx context.Context, tx TTx, params *JobListPara
 	return res, nil
 }
 
+// NotifyKind is a kind of notification to be sent through Client.Notify.
+type NotifyKind string
+
+// NotifyKindLeaderRequestResign sends a notification for the current leader to
+// resign.
+const NotifyKindLeaderRequestResign NotifyKind = "leader_request_resign"
```
bgentry: We have these constants defined in `notifier.go` already. Maybe we should keep this one's string value defined there and just alias it here? I don't think users would generally need to see the actual topic string anyway so I'm less concerned about the indirection of jumping to another package to see the raw value.

https://github.com/riverqueue/river/blob/3559f4d584191b88e8e563efb8177bb70d622ee1/internal/notifier/notifier.go#L22-L28

client.go
```diff
@@ -2275,6 +2275,77 @@ func (c *Client[TTx]) JobListTx(ctx context.Context, tx TTx, params *JobListPara
...
+
+// Notify sends a notification of a known kind along with a compatible payload.
+// The payload type is any because it varies based on notification kind.
+//
+// This function should only be used with a driver that supports listen/notify.
+// Sending a notification with no listener available will have no effect.
```
bgentry: Even if the driver doesn't support `LISTEN`, it may still support `NOTIFY`, right? Although if you're using the same driver everywhere then the distinction is not going to matter.

Applicable to the tx variant as well.

rivershared/util/serviceutil/service_util.go
```diff
@@ -62,7 +62,7 @@ const MaxAttemptsBeforeResetDefault = 7
 // - 8s
 // - 16s
 // - 32s
-// - 64s
+// - 64s.
```
bgentry: mistakenly added?

internal/leadership/elector.go
```diff
@@ -247,34 +257,44 @@ func (e *Elector) handleLeadershipNotification(ctx context.Context, topic notifi
...
+		isLeader := e.isLeader
+		e.mu.Unlock()
+
+		if !isLeader {
+			return
+		}
+
+		select {
+		case <-ctx.Done():
+		case e.forceResignChan <- struct{}{}:
```
bgentry: If we're not currently the leader, is this channel prone to filling up and causing this notification handler to block and buffer indefinitely? In the case of the resignation channel, we explicitly drain these while attempting to gain leadership, but AFAICT you didn't add an equivalent for this change:

https://github.com/riverqueue/river/blob/3559f4d584191b88e8e563efb8177bb70d622ee1/internal/leadership/elector.go#L224-L239

Are we also missing test coverage on whether or not a burst of events is handled cleanly (a sort of stress test)?

bgentry (commented):
Mostly good, except I believe there's one bug around this new channel potentially filling up when not the leader. Lmk if you think I've misidentified that! It's definitely a little tricky adding more channels into this model.

This would also be great to add to the UI on the client table. In fact we should also be able to indicate the leader there, which we don't currently do.

client.go
brandur: Yeah, good call.

Actually, I was thinking about this today, and I think I came up with a better API than what I had. I just pushed a change to make "bundle style" like `PeriodicJobs()` so that notifications work like this:

``` go
client.Notify().RequestResign(ctx)
client.Notify().RequestResignTx(ctx, tx)
```

I think this is the best of both worlds because we don't need to have the "generic" `Notify` API with kinds and `any` payload which is awkward for people to use, but we also don't make the client API crowded because all notification-related functions are consigned to the `Notify()`.

Want to take a look at the latest and see what you think?

client.go
brandur: Okay I just pushed a change to take this error out. I thought it might help people accidentally send a notify that's not going to be received, but I don't feel too strongly about it.

rivershared/util/serviceutil/service_util.go
brandur: Yeah, my local golangci-lint adds this for some reason (need to investigate). I took it back out.

internal/leadership/elector.go
brandur: Okay I made a couple tweaks here:

* Made the channel async of size of 1.
* When sending into the channel, added a `default` case so in case of many notifications received at once, just drop the additional ones.
* When losing leadership, the current leader drains the channel before leaving the function.

Also note that when receiving the notification, it's ignored unless the elector is currently leader:

``` go
	switch notification.Action {
	case DBNotificationKindRequestResign:
		e.mu.Lock()
		isLeader := e.isLeader
		e.mu.Unlock()

		if !isLeader {
			return
		}
```

I added a stress test case as suggested to try to demonstrate that this is working as expected. Want to take a look and LMK what you think?

brandur:
Great. Thanks!

PR #1089, opened 2025-11-21
Remove impossible `nil` checks

brandur:
I just noticed while reading the unique field code that we seemed to
have a bug:

    if typ == nil || typ.Kind() != reflect.Struct {
            return nil, fmt.Errorf("expected struct, got %T", typ.Name())
    }

If the first condition of `typ == nil` were to be taken, the line below
would panic on `typ.Name()` because `typ` was `nil`.

The fact that nothing is panicking and no panics have been reported
seems to suggest that this case never actually happens. Indeed, reading
code that calls into this function, there's no place where a `nil` type
could ever be returned (`field.Type` which will always have a type and
`reflect.TypeOf(...)` which would return `nil` in the case of
`reflect.TypeOf(nil)`, but because we've already called `args.Kind()`
before ever getting to this function, that will never be possible).


brandur:
thx!

PR #1093, opened 2025-11-23
Add `riverlog.LoggerSafely` for access outside work body without panic

### LINKED ISSUE #1092: Safe logger
Hello! It would be nice to have a `SafeLogger` method inside riverlog package

For example, I have method which I'm calling inside and outside of river job in my repository. I may want to log some info in river to improve debug or smth like this. For now I can't do this in any way because I have no access to the key inside the context and it leads to panic.

I am thinking about having noop logger as a default value when logger is not inside the context

This is not going to break back compatibility and only improves UX

brandur:
Here, add a "safe" variant of `riverlog.Logger`: `rigerlog.LoggerSafely`.

While it'd be rare to not know whether you had a logger or not from
inside a `Work` function, `Work` functions might call into other Go
modules in your project, those modules might be reused in non-`Work`
contexts, and it might not always be clear whether a logger should be
available or not.

This function lets users do a conditional check before logging:

    if logger, ok := riverlog.LoggerSafely(ctx); ok {
        logger.InfoContext(ctx, "Hello from logger")
    }

I went with the "safely" naming according to our conventions elsewhere,
but one thing I'm not sure about is that `ClientFromContextSafely` which
is a similar function, returns an error instead of `ok` boolean. Here we
break with that convention somewhat, but I think the boolean version is
more ergonomic. Another option might be to give this a slightly
alternative naming convention like `LoggerOK` or `LoggerMaybe`.

Fixes #1092.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `riverlog.LoggerSafely` which provides a non-panic variant of `riverlog.Logger` for use when code may or may not have a context logger available. [PR #1093](https://github.com/riverqueue/river/pull/1093).
+
 ## [0.27.1] - 2025-11-21
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


brandur:
The other option is that we could just get rid of the panic and have `Logger` return a simple `nil` in case it wasn't available in context. Not sure which version is better, although that one would be a little more risky in causing nil pointer exceptions.

PR #1097, opened 2025-11-25
Basic stuck job detection

brandur:
Here, try to make some inroads on a feature we've been talking about for
a while: detection of stuck jobs.

Unfortunately in Go it's quite easy to accidentally park a job by using
a `select` on a channel that won't return and forgetting a separate
branch for `<-ctx.Done()` so that it won't respect job timeouts either.

Here, add in some basic detection for that case. Eventually we'd like to
give users some options for what to do in case jobs become stuck, but
here we do only the simplest things for now: log when we detect a stuck
job and count the number of stuck jobs in a producer's stats loop.

In the future we may want to have some additional intelligence like
having producers move stuck jobs to a separate bucket up to a certain
limit before crashing (the next best option because it's not possible to
manually kill goroutines).

### DIFF
--- a/internal/jobexecutor/job_executor.go
+++ b/internal/jobexecutor/job_executor.go
@@ -112,12 +112,17 @@ type JobExecutor struct {
 	ErrorHandler             ErrorHandler
 	HookLookupByJob          *hooklookup.JobHookLookup
 	HookLookupGlobal         hooklookup.HookLookupInterface
-	InformProducerDoneFunc   func(jobRow *rivertype.JobRow)
 	JobRow                   *rivertype.JobRow
 	MiddlewareLookupGlobal   middlewarelookup.MiddlewareLookupInterface
-	SchedulerInterval        time.Duration
-	WorkerMiddleware         []rivertype.WorkerMiddleware
-	WorkUnit                 workunit.WorkUnit
+	ProducerCallbacks        struct {
+		JobDone func(jobRow *rivertype.JobRow)
+		Stuck   func()
+		Unstuck func()
+	}
+	SchedulerInterval      time.Duration
+	StuckThresholdOverride time.Duration
+	WorkerMiddleware       []rivertype.WorkerMiddleware
+	WorkUnit               workunit.WorkUnit
 
 	// Meant to be used from within the job executor only.
 	start time.Time
@@ -159,7 +164,7 @@ func (e *JobExecutor) Execute(ctx context.Context) {
 		}
 	}
 
-	e.InformProducerDoneFunc(e.JobRow)
+	e.ProducerCallbacks.JobDone(e.JobRow)
 }
 
 // Executes the job, handling a panic if necessary (and various other error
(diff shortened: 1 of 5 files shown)


internal/jobexecutor/job_executor.go
```diff
@@ -171,6 +176,51 @@ func (e *JobExecutor) execute(ctx context.Context) (res *jobExecutorResult) {
...
+					slog.Int64("job_id", e.JobRow.ID),
+					slog.String("kind", e.JobRow.Kind),
+					slog.Duration("timeout", e.ClientJobTimeout),
+				)
+
+				// In case the executor ever becomes unstuck, inform the
+				// producer. However, if we got all the way here there's a good
+				// chance this will never happen (the worker is really stuck and
+				// will never return).
+				defer e.ProducerCallbacks.Unstuck()
```
bgentry, line 219: Doesn't this just run immediately after the above warning log? It's deferring within the inner `go func()` closure which merely exits after both these defers are added to the stack, and there's nothing to block on the jobs actually becoming unstuck. Or am I missing something?

internal/jobexecutor/job_executor.go
```diff
@@ -171,6 +176,51 @@ func (e *JobExecutor) execute(ctx context.Context) (res *jobExecutorResult) {
...
+	// to context cancellation (unfortunately, quite an easy error to make in
+	// Go).
+	//
+	// Currently we don't do anything if we notice a job is stuck. Knowing about
+	// stuck jobs is just used for informational purposes in the producer in
+	// generating periodic stats.
+	if e.ClientJobTimeout > 0 {
+		ctx, cancel := context.WithCancel(ctx)
+		defer cancel()
+
```
bgentry: AFAICT `ctx` is the main job context, which could be cancelled under a variety of circumstances (aggressive client shutdown, manual cancellation attempt via UI, etc). That would lead the `case <- ctx.Done()` below to exit even if the job is actually stuck.

Am I misunderstanding this?

internal/jobexecutor/job_executor.go
brandur: Yeah, there was another `<- ctx.Done()` that was missing here. I've added that in, and also improved the test case so that it'll fail in the event that wait is missing.

I also put in some more logging in the test case so you can run it and verify manually (in case you want/need to) that it's working as expected. e.g.:

```
$ go test ./internal/jobexecutor -run TestJobExecutor_Execute/StuckDetectionActivates -test.v
=== RUN   TestJobExecutor_Execute
=== PAUSE TestJobExecutor_Execute
=== CONT  TestJobExecutor_Execute
=== RUN   TestJobExecutor_Execute/StuckDetectionActivates
=== PAUSE TestJobExecutor_Execute/StuckDetectionActivates
=== CONT  TestJobExecutor_Execute/StuckDetectionActivates
    riverdbtest.go:216: Dropped 1 expired postgres schema(s) in 14.537458ms
    riverdbtest.go:293: TestSchemaOpts.disableReuse is set; schema not checked in for reuse
    job_executor_test.go:715: Generated postgres schema "jobexecutor_2025_12_08t09_03_56_schema_01" with migrations [1 2 3 4 5 6] on line "main" in 63.787208ms [1 generated] [0 reused]
    job_executor_test.go:715: TestTx using postgres schema: jobexecutor_2025_12_08t09_03_56_schema_01
    job_executor_test.go:724: Job executor reported stuck
    logger.go:256: time=2025-12-08T09:03:56.218-05:00 level=WARN msg="jobexecutor.JobExecutor: Job appears to be stuck" job_id=1 kind=jobexecutor_test timeout=5ms
    job_executor_test.go:739: Job executor still stuck after wait (this is expected)
    logger.go:256: time=2025-12-08T09:03:56.229-05:00 level=INFO msg="jobexecutor.JobExecutor: Job became unstuck" duration=17.011ms job_id=1 kind=jobexecutor_test
    job_executor_test.go:728: Job executor reported unstuck (after being stuck)
--- PASS: TestJobExecutor_Execute (0.00s)
    --- PASS: TestJobExecutor_Execute/StuckDetectionActivates (0.12s)
PASS
ok      github.com/riverqueue/river/internal/jobexecutor        0.299s
```

internal/jobexecutor/job_executor.go
brandur: Yeah, on second thought, it makes sense to have a `context.WithoutCancel(...)` on that thing. Added, and put in a new test case that checks it's doing the right thing.

internal/jobexecutor/job_executor.go
```diff
@@ -171,6 +176,57 @@ func (e *JobExecutor) execute(ctx context.Context) (res *jobExecutorResult) {
...
+	//
+	// Currently we don't do anything if we notice a job is stuck. Knowing about
+	// stuck jobs is just used for informational purposes in the producer in
+	// generating periodic stats.
+	if e.ClientJobTimeout > 0 {
+		// We add a WithoutCancel here so that this inner goroutine becomes
+		// immune to all context cancellations _except_ the one where it's
+		// cancelled because we leave JobExecutor.execute.
+		ctx, cancel := context.WithCancel(context.WithoutCancel(ctx))
+		defer cancel()
```
bgentry, line 194: May be worth calling out the intentional shadowing here too or using an alternate `ctx` var name

internal/jobexecutor/job_executor_test.go
```diff
@@ -696,6 +704,94 @@ func TestJobExecutor_Execute(t *testing.T) {
...
+		}, nil).MakeUnit(bundle.jobRow)
+
+		executor.Execute(ctx)
+		_ = riversharedtest.WaitOrTimeout(t, bundle.updateCh)
+
+		riversharedtest.WaitOrTimeout(t, informProducerUnstuckReceived)
+	})
+
+	// Checks that even if a work's context is cancelled immediately, stuck
+	// detection still works as expected.
```
bgentry: Maybe a typo? "work" or "work function's"?

```suggestion
	// Checks that even if a work context is cancelled immediately, stuck
	// detection still works as expected.
```

internal/jobexecutor/job_executor.go
brandur: Works for me. Added comment.

internal/jobexecutor/job_executor_test.go
brandur: Fixed!

PR #1098, opened 2025-12-02
Introduce `Client.JobUpdate` function that can store output incrementally

brandur:
Here, take a stab at a solution for #1064. We introduce a new client
function `Client.JobUpdate` that takes any output currently recorded in
context and stores it to the given job row. `JobUpdate` currently only
sets output, but the idea is that it could be expanded in the future in
case it's useful to do so.

The reason that we don't have a `river.PersistOutput` in line with
`river.RecordOutput` is that once we're talking about storing data,
we're dealing with the usual persistence semantics. We need a client
instance with access to an executor, and we probably want to have a
`*Tx` variant like we have for all other functions like this one.

Fixes #1064.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Basic stuck detection after a job's exceeded its timeout and still not returned after the executor's initiated context cancellation and waited a short margin for the cancellation to take effect. [PR #1097](https://github.com/riverqueue/river/pull/1097).
+- Added `Client.JobUpdate` which can be used to persist job output partway through a running work function instead of having to wait until the job is completed. [PR #1098](https://github.com/riverqueue/river/pull/1098).
 - Add a little more error flavor for when encountering a deadline exceeded error on leadership election suggesting that the user may want to try increasing their database pool size. [PR #1101](https://github.com/riverqueue/river/pull/1101).
 
 ## [0.29.0-rc.1] - 2025-12-04
+
+### Added
 
 - Added `HookPeriodicJobsStart` that can be used to run custom logic when a periodic job enqueuer starts up on a new leader. [PR #1084](https://github.com/riverqueue/river/pull/1084).
 - Added `Client.Notify().RequestResign` and `Client.Notify().RequestResignTx` functions allowing any client to request that the current leader resign. [PR #1085](https://github.com/riverqueue/river/pull/1085).
--- a/client.go
+++ b/client.go
@@ -17,6 +17,7 @@ import (
 	"github.com/riverqueue/river/internal/dbunique"
 	"github.com/riverqueue/river/internal/hooklookup"
 	"github.com/riverqueue/river/internal/jobcompleter"
+	"github.com/riverqueue/river/internal/jobexecutor"
 	"github.com/riverqueue/river/internal/leadership"
 	"github.com/riverqueue/river/internal/maintenance"
 	"github.com/riverqueue/river/internal/middlewarelookup"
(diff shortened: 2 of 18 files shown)


riverdriver/riversqlite/internal/dbsqlc/river_job.sql
```diff
@@ -503,7 +503,9 @@ SET
     errors = CASE WHEN cast(@errors_do_update AS boolean) THEN @errors ELSE errors END,
     finalized_at = CASE WHEN cast(@finalized_at_do_update AS boolean) THEN @finalized_at ELSE finalized_at END,
     max_attempts = CASE WHEN cast(@max_attempts_do_update AS boolean) THEN @max_attempts ELSE max_attempts END,
-    metadata = CASE WHEN cast(@metadata_do_update AS boolean) THEN json(cast(@metadata AS blob)) ELSE metadata END,
+    metadata = CASE WHEN cast(@metadata_do_update AS boolean) THEN json(cast(@metadata AS blob))
+                    WHEN cast(@metadata_do_merge AS boolean) THEN json_patch(metadata, json(cast(@metadata AS blob)))
+                    ELSE metadata END,
```
brandur: Actually, I'll probably have to move this to a `SELECT ... FOR UPDATE` version of this function.

brandur: Alright, updated this with a locking version.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `Client.JobUpdate` which can be used to persist job output partway through a running work function instead of having to wait until the job is completed. [PR #1093](https://github.com/riverqueue/river/pull/1093).
```
```suggestion
- Added `Client.JobUpdate` which can be used to persist job output partway through a running work function instead of having to wait until the job is completed. [PR #1098](https://github.com/riverqueue/river/pull/1098).
```

riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
```diff
@@ -712,3 +712,18 @@ SET
...
+    WHERE river_job.id = @id
+    FOR UPDATE
+)
+UPDATE /* TEMPLATE: schema */river_job
+SET
+    metadata = CASE WHEN @metadata_do_merge::boolean THEN metadata || @metadata::jsonb ELSE metadata END
+FROM
+    locked_job
+WHERE river_job.id = locked_job.id
+RETURNING river_job.*;
```
bgentry: The name `JobUpdateFast` doesn't seem right—it's not really faster than an alternative option and in its current form it merely updates metadata. Maybe we will expand its purpose, I'm not sure, but in the mean time we could probably still pick a better suffix.

One idea is to rename the current `JobUpdate` to `JobUpdateFull` or something of that sort and let this one take the `JobUpdate` name.

client.go
```diff
@@ -1504,6 +1505,55 @@ func (c *Client[TTx]) jobRetry(ctx context.Context, exec riverdriver.Executor, i
...
+type JobUpdateParams struct{}
+
+// JobUpdate updates the job with the given ID.
+//
+// Currently, no fields are explicitly updatable via this method, but if called
+// inside a work function, it will set output recorded with RecordOutput, if
+// any.
+func (c *Client[TTx]) JobUpdate(ctx context.Context, id int64, params *JobUpdateParams) (*rivertype.JobRow, error) {
+	return c.jobUpdate(ctx, c.driver.GetExecutor(), id, params)
+}
```
bgentry: I'm fully onboard with providing an API to make it easy to write metadata updates, particularly for output. At the same time I'm wondering if we want to couple this functionality for "store the recorded output immediately instead of waiting for job completion" to something that's named as a general-purpose `JobUpdate`. People will probably want to use this to make arbitrary updates to jobs, including those that aren't even the job currently being executed. That'd lead to this method having some surprising and potentially undesired behavior just depending on whether it's called for the current running job or another one.

It also feels a little weird/confusing/unnatural to have to first call the `river.RecordOutput()` function (which buffers up or lazily records the output) followed by a subsequent call to some client method to immediately save the already-buffered output.

Despite all this, I understand why you went down this route. A special-purpose API like `Client.JobRecordOutput` / `Client.JobRecordOutputNow` feels quite specific. And if you exposed metadata as a field on a generalized `JobUpdateParams` it'd be unclear whether it fully replaces or merges with metadata. Additionally, the nicety of `river.RecordOutput` is that it both marshals for you and enforces length restrictions that would have to be reproduced elsewhere.

I don't really have a strong preference on an approach here, but I do think it's important to raise these points and hear your perspective before we commit to one. I think I have a slight preference for a single client method that'd allow directly passing output or other metadata updates to be merged in (while applying the same rules & restrictions we do in `RecordOutput` / with generalized metadata updates from middleware), primarily for two reasons:

1. Avoid the two-step process of `river.RecordOutput` + `client.JobUpdate` without params, in favor of a single explicit call. Let `river.RecordOutput` be clearly documented to do so lazily, and cross-reference w/ the new method documentation to explain when to use each.
2. Avoid behavioral differences depending on _which_ job you're updating (the currently running one or another, say a dependency).

Thoughts?

CHANGELOG.md
brandur: Fixed, thx.

riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
brandur: K, I changed this to `JobUpdate` + `JobUpdateFull`.

client.go
brandur: Yeah, I can think of a good half dozen alternatives to this, but I couldn't think of any that are _better_ than the proposed route, just different. For example:

* Specific `Client.JobRecordOutput` functions: definitely thought about this, but just feels too specific and unnecessary given it wouldn't have a major advantage over `JobUpdate` and we might want to expand `JobUpdate`.
* `Client.JobUpdate` (or `Client.JobRecordOutput`) _instead of_ `river.RecordOutput` + `Client.JobUpdate`: I see this as a possibility, but it definitely felt to me that it'd be surprising if the normal `river.RecordOutput` didn't work as expected when doing a mid-work output update.

I've pushed a new change in which I augment the new `JobUpdate` so it now has two modes of operation:

* If `river.RecordOutput` has been used, it uses the output out of that to save to the job row.
* Adds a `JobUpdateParams.Output` that can be used to override output directly, thereby allowing a one-step operation.

In case both context output and `JobUpdateParams.Output` is set, we error as it's ambiguous which one was intended.

This feels like a pretty good compromise to me: you get shorter usage if you want it, but we don't introduce any API footguns.

I still need to make a few augmentation like checking output length and such, but do you want to take a look and see if you're okay with this direction?

client.go
```diff
@@ -1514,6 +1515,87 @@ func (c *Client[TTx]) jobRetry(ctx context.Context, exec riverdriver.Executor, i
...
 
+type JobUpdateParams struct {
+	Output any
+}
+
+// .JobUpdateFull updates the job with the given ID.
+//
+// Currently, no fields are explicitly updatable via this method, but if called
+// inside a work function, it will set output recorded with RecordOutput, if
+// any.
```
bgentry: Typo at the start and slightly outdated given the new behavior here. Tx variant below also needs updating.

client.go
```diff
@@ -1514,6 +1515,87 @@ func (c *Client[TTx]) jobRetry(ctx context.Context, exec riverdriver.Executor, i
...
+		if val, ok := metadataUpdates[rivertype.MetadataKeyOutput]; ok {
+			return val.(json.RawMessage) //nolint:forcetypeassert
+		}
+
+		return nil
+	}()
+
+	var (
+		metadataDoMerge      bool
+		metadataUpdatesBytes = []byte("{}") // even in the even of no update, still valid jsonb
```
```suggestion
		metadataUpdatesBytes = []byte("{}") // even in the event of no update, still valid jsonb
```

client.go
```diff
@@ -1514,6 +1515,87 @@ func (c *Client[TTx]) jobRetry(ctx context.Context, exec riverdriver.Executor, i
...
+	)
+	if outputFromWorkContext != nil && params.Output != nil {
+		return nil, errors.New("should not set job output both from work context (via RecordOutput) and in JobUpdateParams")
```
bgentry: Hmm, feels a bit surprising, but then so is the alternative of quietly allowing `JobUpdate` to take precedence over the previously recorded output. I think I lean slightly in the direction of allowing it with documentation instead of erroring as done here since that would only happen if you're explicitly updating the job with an `Output` _after_ you've already called `RecordOutput`, but I could go either way as long as clearly documented.

For reference we already make the last-write-wins behavior clear in `RecordOutput` so IMO allowing it here too would be consistent with that:

```go
// Only one output can be stored per job. If this function is called more than
// once, the output will be overwritten with the latest value. The output also
// must be recorded _before_ the job finishes executing so that it can be stored
// when the job's row is updated.
```

bgentry (commented):
Minor comments only, let's ship this! I think this is a good approach generally, and I like that it gives you flexibility.

We should probably also tweak docs on `RecordOutput` to make its lazy-write behavior clear and could call out `JobUpdate` / `JobUpdateTx` for cases where you want an immediate extra write.

client.go
brandur: Yep, good call. Fixed.

client.go
brandur: Thx, fixed.

client.go
brandur: Alright, works for me. I tweaked the code so it's no longer an error and that params output always takes precedence.

PR #1106, opened 2025-12-21
When operating outside of transaction, add/remove versions immediately

brandur:
We had a problem reported recently wherein a user who was executing
migrations in tight contention with each other hit a problem where
because a transaction is only in operation right around the immediate
SQL of each migration, it's easily possible to get all the side effects
of a transaction without actually adding/removing any versions. In the
event of such a failure, the database is left in an irreconcilable
state. Neither migrating up or down can fix the problem, with the only
option being manual intervention.

Here, when running without an outer transaction, run version insertion
or deletion immediately so that it happens immediately inside the inner
transaction, thereby preventing this problem.

All `MigrateTx` test exercise was completely stripped out so I added a
couple basic tests back in so that this code has some coverage.

### DIFF
--- a/rivermigrate/river_migrate.go
+++ b/rivermigrate/river_migrate.go
@@ -26,10 +26,17 @@ import (
 	"github.com/riverqueue/river/rivershared/util/sliceutil"
 )
 
-// The migrate version where the `line` column was added. Meaningful in that the
-// migrator has to behave a little differently depending on whether it's working
-// with versions before or after this boundary.
-const migrateVersionLineColumnAdded = 5
+const (
+	// The migrate version where the `line` column was added. Meaningful in that
+	// the migrator has to behave a little differently depending on whether it's
+	// working with versions before or after this boundary.
+	migrateVersionLineColumnAdded = 5
+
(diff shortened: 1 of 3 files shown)


rivermigrate/river_migrate.go
```diff
@@ -414,36 +421,19 @@ func (m *Migrator[TTx]) migrateDown(ctx context.Context, exec riverdriver.Execut
...
-		} else {
-			if _, err := exec.MigrationDeleteByLineAndVersionMany(ctx, &riverdriver.MigrationDeleteByLineAndVersionManyParams{
-				Line:     m.line,
-				Schema:   m.schema,
-				Versions: versions,
-			}); err != nil {
-				return nil, fmt.Errorf("error deleting migration rows for versions %+v on line %q: %w", res.Versions, m.line, err)
-			}
+	// When operating with an outer transaction, add/remove all versions at once
+	// so we can save a few database operations.
```
bgentry: This is only _removing_, not _adding_, right?

brandur: Ah good catch. Corrected this comment.

Issue #1105, opened 2025-12-17
Worker is trying to rescue jobs from queues that are not assigned to it


NanoBjorn:
# A bit of context

We have two kinds of workers at the moment, sharing the same database, let's say `worker-1` is running jobs from `queue-1` and `worker-2` is running jobs from `queue-2`. However, we are using the same binary for the worker and we recently tried registering jobs conditionally, basically registering only those jobs that are supposed to be executed by this worker based on queue selection.

Another thing to mention is that our `RescueStuckJobsAfter` is configured to be pretty small (30s) whereas timeouts on the jobs are minutes.

# An issue
After introducing this conditional registration, we observed the following behaviour:
- `worker-1` is fetching all jobs (for both queues) to rescue: https://github.com/riverqueue/river/blob/eb0b9854baa90b073783ae76841df154c311b7dd/riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql#L257-L263
- amongst others, it can fetch jobs dedicated for `queue-2` and they are not registered in the `worker-1`, so `makeRetryDecision` discards the job: https://github.com/riverqueue/river/blob/eb0b9854baa90b073783ae76841df154c311b7dd/internal/maintenance/job_rescuer.go#L302-L307

Checked documentation if we are missing something, and both pages that could have mention any details of this behaviour are not saying anything about our weird case:
- [Inserting and working jobs](https://riverqueue.com/docs/inserting-and-working-jobs) does not mention if all jobs from a single database have to be registered in all workers
- [Multiple queues](https://riverqueue.com/docs/multiple-queues) only says _but workers will only select jobs to work for queues that they're configured to handle_, but still it is unclear if we should register all jobs or not

A very bad side effect of this for us was that discarded job continued execution (context wasn't cancelled) and another job with the same `unique_key` was scheduled, violating unique constraint that we are relying on for correctness, but I am not blaming this on river, this is a consequence of our ignorance of rescue mechanics nuances. 

---

Short-term we fixed our issue by again registering all jobs in the worker binary regardless its configuration, but it would be valuable to have an answer if this is a designed behaviour or a bug?

NanoBjorn:
>  "orphaned" will not get any attention from any clients

Yeah, that makes sense, I was also thinking about it.

Then I think action items on this one could be:
1. Mention this in both documentation pages on registering jobs and multiple queues
2. Cancel context by the job itself if the job was marked as discarded in the db (I am not sure though if there is any active polling from the side of running job now)

[...]

Issue #342, opened 2024-05-07
Periodic Job enqueuing & leader election with multiple different river clients


magaldima:
I've observed (expected) behavior while developing with River (v0.4.1) that periodic jobs are only enqueued by an elected leader which I now understand to be part of the [maintenance service responsibilities of an elected leader](https://riverqueue.com/docs/maintenance-services#periodic-enqueuer).

The "problem" that I'm experiencing is that I'm attempting to use river (and initialize River Clients) across a few different services that all have different workers and different river periodic job configurations and are mostly isolated by different queues. So if I have service A with a periodic job that executes every hour and service B with a different periodic job, if service B is the elected leader for more than 1 hour, service A's periodic job will not be enqueued per its configured schedule.

This is less of an issue and more of a question around how River was designed and if this use case was considered. Presumably, several services sharing a background job queue is a common pattern and we'd like for each service to contain the logic of processing jobs specific to that service. Does this necessitate that scheduled jobs should be declared and managed by a single service or that all the different applications that create a River client need to use the same River config (with the same set of periodic jobs including those that are not in scope for the specific service)?

The other workaround currently under discussion at the moment is creating separate DB schemas per service and instantiating the River tables in each schema so that each service has it's own elected leader. On the surface this seems like something River should be able to handle and I'm worried about the increased overhead with maintaining separate DB schemas for each application.

I was hoping you would be able to share insights/context into River's design and if this was something that was considered at any point in time and how you thought about it and if this is on your radar/roadmap. Curious what your recommendation is for a solution: consolidating job definition vs. separate DB schemas (or something else entirely) ? 

Another last idea that I had was to create a new DB table in River `river_periodic_job` where the definition/configuration for periodic jobs could live and be shared across all services in a way that the river client configuration state can be pushed into the DB layer. In this way, the elected leader would have visibility into all periodic jobs and be able to enqueue jobs based on the periodic jobs defined in that DB table. Curious your thoughts on this idea as well - happy to help put forth a proposal and contribute time/work if you think this feature would be doable. 

Thanks for the time for reading that through! I've been using River for a few months now and have loved it. If anything, this "problem" has arisen out of my exuberance to extend River's capabilities to many of our backend services.

elee1766:
> that all the different applications that create a River client need to use the same River config (with the same set of periodic jobs including those that are not in scope for the specific service)?

not the exact same river config, but yes, sharing the same scheduled_job across both workers and insert only client is what we are doing

PR #1113, opened 2025-12-23
Add leadership "domains" so multiple Rivers can operate in one schema

brandur:
We've gotten a couple requests so far (see #342 and #1105) to be able to
start multiple River clients targeting different queues within the same
database/schema, and giving them the capacity to operate independently
enough to be functional. This is currently not possible because a single
leader is elected given a single schema and it handles all maintenance
operations including non-queue ones like periodic job enqueuing.

Here, add the idea of a `LeaderDomain`. This lets a user set the
"domain" on which a client will elect its leader and allowing multiple
leaders to be elected in a single schema. Each leader will run its own
maintenance services.

Setting `LeaderDomain` causes the additional effect of having
maintenance services start to operate only on the queues that their
client is configured for. The idea here is to give us backwards
compatibility in that the default behavior (in case of an unset
`LeaderDomain`) is the same, but providing a path for multiple leaders
to be interoperable with each other.

There are still a few edges: for example, reindexing is not queue
specific, so multiple leaders could be running a reindexer. I've
provided guidance in the config documentation that ideally, all clients
but one should have their reindexer disabled.

### DIFF
--- a/client.go
+++ b/client.go
@@ -9,6 +9,7 @@ import (
 	"log/slog"
 	"os"
 	"regexp"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -219,6 +220,48 @@ type Config struct {
 	//
 	// Jobs may have their own specific hooks by implementing JobArgsWithHooks.
 	Hooks []rivertype.Hook
+
+	// LeaderDomain is an optional "domain" string to use for leader election.
+	// Different clients sharing the same River schema can elect multiple
+	// leaders as long as they're using different domains, with one leader
+	// elected per domain. This is an advanced feature that will almost never be
+	// needed. Don't set this unless you know what you're doing.
+	//
+	// Setting this value also triggers the related behavior that maintenance
+	// services start to only operate on the queues they're configured on. So
+	// for example, given client1 handling queue_a and queue_b and client2
+	// handling queue_c and queue_d, whichever client is elected leader will end
+	// up running all maintenance services for all queues (queue_a, queue_b,
+	// queue_c, and queue_d). But if client1 is using domain "domain1" and
+	// client2 is using domain "domain2", then client1 (elected in domain1) will
+	// only run maintenance services on queue_a and queue_b, while client2
+	// (elected in domain2) will run maintenance services on queue_c and
+	// queue_d.
+	//
+	// A warning though that River *does not protect against configuration
+	// mistakes*. If client1 on domain1 is configured for queue_a and queue_b,
+	// and client2 on domain2 is *also* configured for queue_a and queue_b, then
+	// both clients may end up running maintenance services on the same queues
+	// at the same time. It's the caller's responsibility to ensure that doesn't
+	// happen.
+	//
+	// Left empty or use of the special value "default" causes the client to
+	// operate on all queues. When setting this value to non-empty
+	// non-"default", no other clients should be left empty or use "default"
+	// because the default client(s) will infringe on the domains of the
+	// non-default one(s).
+	//
+	// Certain maintenance services that aren't queue-related like the reindexer
+	// will continue to run on all leaders regardless of domain. If using this
+	// feature, it's a good idea to configure ReindexerTimeout on all but a
+	// single leader domain to river.NeverSchedule().
+	//
+	// In general, most River users should not need LeaderDomain, and when
+	// running multiple Rivers may want to consider using multiple databases and
+	// multiple schemas instead.
+	//
+	// Defaults to "default".
+	LeaderDomain string
 
 	// Logger is the structured logger to use for logging purposes. If none is
 	// specified, logs will be emitted to STDOUT with messages at warn level
@@ -445,6 +488,7 @@ func (c *Config) WithDefaults() *Config {
 		Hooks:                       c.Hooks,
 		JobInsertMiddleware:         c.JobInsertMiddleware,
 		JobTimeout:                  cmp.Or(c.JobTimeout, JobTimeoutDefault),
+		LeaderDomain:                c.LeaderDomain,
 		Logger:                      logger,
 		MaxAttempts:                 cmp.Or(c.MaxAttempts, MaxAttemptsDefault),
 		Middleware:                  c.Middleware,
@@ -873,6 +917,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 
 		client.elector = leadership.NewElector(archetype, driver.GetExecutor(), client.notifier, &leadership.Config{
 			ClientID: config.ID,
+			Domain:   config.LeaderDomain,
 			Schema:   config.Schema,
 		})
 		client.services = append(client.services, client.elector)
@@ -890,6 +935,12 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 			client.services = append(client.services, pluginPilot.PluginServices()...)
 		}
 
+		var queuesIncluded []string
+		if config.LeaderDomain != "" && config.LeaderDomain != leadership.DomainDefault && len(config.Queues) > 0 {
+			queuesIncluded = maputil.Keys(config.Queues)
+			slices.Sort(queuesIncluded)
+		}
+
 		//
 		// Maintenance services
 		//
@@ -902,6 +953,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 				CompletedJobRetentionPeriod: config.CompletedJobRetentionPeriod,
 				DiscardedJobRetentionPeriod: config.DiscardedJobRetentionPeriod,
 				QueuesExcluded:              client.pilot.JobCleanerQueuesExcluded(),
+				QueuesIncluded:              queuesIncluded,
 				Schema:                      config.Schema,
 				Timeout:                     config.JobCleanerTimeout,
 			}, driver.GetExecutor())
@@ -912,6 +964,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		{
 			jobRescuer := maintenance.NewRescuer(archetype, &maintenance.JobRescuerConfig{
 				ClientRetryPolicy: config.RetryPolicy,
+				QueuesIncluded:    queuesIncluded,
 				RescueAfter:       config.RescueStuckJobsAfter,
 				Schema:            config.Schema,
 				WorkUnitFactoryFunc: func(kind string) workunit.WorkUnitFactory {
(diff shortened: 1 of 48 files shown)


brandur:
Okay, the tests for this should be in good shape now.

riverdriver/riverpgxv5/migration/main/007_river_leader_non_default_name.down.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE /* TEMPLATE: schema */river_leader
```
brandur, line 1: Since we're adding a migration here, before shipping I'd also go through the existing "needs migration" issues and pull them in.

One nice thing about this migration as currently written is that if you did *not* run it, it wouldn't be a problem as long as you didn't try to use `LeaderDomain`. So it's safer than your average migration.

client.go
```diff
@@ -209,6 +210,47 @@ type Config struct {
...
+	// at the same time. It's the caller's responsibility to ensure that doesn't
+	// happen.
+	//
+	// Left empty or use of the special value "default" causes the client to
+	// operate on all queues. When setting this value to non-empty
+	// non-"default", no other clients should be left empty or use "default"
+	// because the default client(s) will infringe on the domains of the
+	// non-default one(s).
+	//
+	// Certain maintenance services that aren't queue-related like the indexer
```
```suggestion
	// Certain maintenance services that aren't queue-related like the reindexer
```

client.go
```diff
@@ -209,6 +210,47 @@ type Config struct {
...
+	// will continue to run on all leaders regardless of domain. If using this
+	// feature, it's a good idea to configure ReindexerTimeout on all but a
+	// single leader domain to river.NeverSchedule().
+	//
+	// In general, most River users should not need LeaderDomain, and when
+	// running multiple Rivers may want to consider using multiple databases and
+	// multiple schemas instead.
```
bgentry, line 261: I would probably try to get some version of this warning in the first paragraph to dissuade people from using this feature. Could describe it as "an advanced option" or something like that. I would just like to ensure people don't start using this setting automatically just bc it's there because it has some major footguns.

client.go
```diff
@@ -209,6 +210,47 @@ type Config struct {
...
+	// only run maintenance services on queue_a and queue_b, while client2
+	// (elected in domain2) will run maintenance services on queue_c and
+	// queue_d.
+	//
+	// A warning though that River *does not protect against configuration
+	// mistakes*. If client1 on domain1 is configured for queue_a and queue_b,
+	// and client2 on domain2 is *also* configured for queue_a and queue_b, then
+	// both clients may end up running maintenance services on the same queues
+	// at the same time. It's the caller's responsibility to ensure that doesn't
+	// happen.
```
bgentry, line 246: Gotta say this definitely feels dangerous. I'm wondering what else can break if we allow for breaking one of the main promises of leader election (there can only be one), including Pro features.

client.go
```diff
@@ -860,6 +904,14 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 			client.services = append(client.services, pluginPilot.PluginServices()...)
 		}
 
+		// It's important for queuesIncluded to be `nil` in case it's not in use
+		// for the various driver queries to work correctly.
```
bgentry: maybe the driver layer should handle `nil` and `[]string{}` equivalently to avoid needing to deal with that concern at this level?

internal/leadership/elector.go
```diff
@@ -143,9 +147,9 @@ func (e *Elector) Start(ctx context.Context) error {
 
 	var sub *notifier.Subscription
 	if e.notifier == nil {
-		e.Logger.DebugContext(ctx, e.Name+": No notifier configured; starting in poll mode", "client_id", e.config.ClientID)
+		e.Logger.DebugContext(ctx, e.Name+": Resigned leadership successfully", "client_id", e.config.ClientID, "domain", e.config.Domain)
```
bgentry: was this log text altered by mistake?

internal/leadership/elector.go
```diff
@@ -383,10 +388,11 @@ func (e *Elector) keepLeadershipLoop(ctx context.Context) error {
 			// Reelect timer expired; attempt reelection below.
 		}
 
-		e.Logger.DebugContext(ctx, e.Name+": Current leader attempting reelect", "client_id", e.config.ClientID)
+		e.Logger.InfoContext(ctx, e.Name+": Current leader received forced resignation", "client_id", e.config.ClientID, "domain", e.config.Domain)
```
bgentry: the text here was also changed, I think accidentally

internal/leadership/elector.go
```diff
@@ -424,7 +430,7 @@ func (e *Elector) keepLeadershipLoop(ctx context.Context) error {
 // always surrendered in a timely manner so it can be picked up quickly by
 // another client, even in the event of a cancellation.
 func (e *Elector) attemptResignLoop(ctx context.Context) {
-	e.Logger.DebugContext(ctx, e.Name+": Attempting to resign leadership", "client_id", e.config.ClientID)
+	e.Logger.InfoContext(ctx, e.Name+": Current leader received forced resignation", "client_id", e.config.ClientID, "domain", e.config.Domain)
```
bgentry: I don't think this text is necessarily true, there are other reasons this could be called besides a forced resignation.

riverdriver/riversqlite/internal/dbsqlc/river_job.sql
```diff
@@ -92,34 +92,19 @@ RETURNING *;
...
-    AND (
-        cast(@queues_excluded_empty AS boolean)
-        OR river_job.queue NOT IN (sqlc.slice('queues_excluded'))
-    );
+        )
+        AND (/* TEMPLATE_BEGIN: queues_excluded_clause */ true /* TEMPLATE_END */)
+        AND (/* TEMPLATE_BEGIN: queues_included_clause */ true /* TEMPLATE_END */)
+    ORDER BY id
+    LIMIT @max
+);
```
bgentry, line 107: This comment applies to all drivers but I'm leaving it here because it doesn't seem like the pgx driver's `JobDeleteBefore` was updated in this PR yet. This query is currently targeted at the following index:

```
"river_job_state_and_finalized_at_index" btree (state, finalized_at) WHERE finalized_at IS NOT NULL
```

That will break if we're also filtering by queue, and this may turn into a sequential scan.

I haven't checked the other queries you've modified here, but this is an important consideration we'd need to be very careful of for all of them.

riverdriver/riversqlite/migration/main/007_river_leader_non_default_name.up.sql
```diff
@@ -0,0 +1,16 @@
+--
+-- Alter `river_leader` to remove check constraint that `name` must be
+-- `default`. SQLite doesn't allow schema modifications, so this redefines the
+-- table entirely.
```
bgentry, line 4: > SQLite doesn't allow schema modifications, so this redefines the table entirely.

🤯

rivershared/sqlctemplate/sqlc_template.go
```diff
@@ -188,8 +188,8 @@ func (r *Replacer) RunSafely(ctx context.Context, argPlaceholder, sql string, ar
 	}
 
 	updatedSQL := sql
-	updatedSQL = replaceTemplate(updatedSQL, templateBeginEndRE)
 	updatedSQL = replaceTemplate(updatedSQL, templateRE)
+	updatedSQL = replaceTemplate(updatedSQL, templateBeginEndRE)
```
bgentry, line 191: What's going on with the reordering here?

client.go
```diff
@@ -209,6 +210,47 @@ type Config struct {
...
+	// Left empty or use of the special value "default" causes the client to
+	// operate on all queues. When setting this value to non-empty
+	// non-"default", no other clients should be left empty or use "default"
+	// because the default client(s) will infringe on the domains of the
+	// non-default one(s).
+	//
+	// Certain maintenance services that aren't queue-related like the indexer
+	// will continue to run on all leaders regardless of domain. If using this
+	// feature, it's a good idea to configure ReindexerTimeout on all but a
+	// single leader domain to river.NeverSchedule().
```
NanoBjorn, line 257: Now we are running many pods of the same worker, do we treat as "leader" a pod here or a group of workers? If we are saying that only one pod should be reindexing, then what do we do when the pod dies? Or having let's say 3 reindexer pods and many more that don't would be also alright?

client.go
brandur: Fixed! Thx.

client.go
brandur: Yep, great point. Fixed.

internal/leadership/elector.go
brandur: Yep, thx. Fixed.

internal/leadership/elector.go
brandur: Yep, thx. Fixed.

internal/leadership/elector.go
brandur: Yeah, I think it's wrong. Reverted.

riverdriver/riversqlite/migration/main/007_river_leader_non_default_name.up.sql
brandur: Yeah :( Awful.

rivershared/sqlctemplate/sqlc_template.go
brandur: Just accidental again :/ Reverted.

riverdriver/riversqlite/internal/dbsqlc/river_job.sql
brandur: @bgentry I ended up having the LLM run some minor benchmarking and checking index use. Here's what it produced:

```
  When QueuesIncluded is NULL (the default, no LeaderDomain set): The query uses
  river_job_state_and_finalized_at_index as a single index scan, exactly as before. No change in behavior.

  When QueuesIncluded is a concrete array (non-default LeaderDomain): The planner switches to a BitmapOr
  strategy with three legs:
  - cancelled leg: uses river_job_state_and_finalized_at_index on (state, finalized_at)
  - completed leg: uses river_job_prioritized_fetching_index on (state, queue)
  - discarded leg: uses river_job_state_and_finalized_at_index on (state, finalized_at)

  Then finalized_at and queue are applied as heap filters.

  This is actually reasonable planner behavior. For the completed leg, when filtering by specific queues, the
  planner correctly identifies that (state, queue) from the fetching index is more selective than (state,
  finalized_at) — it narrows to both the state and the small set of queues before checking finalized_at. The
  total estimated cost (21.76) is lower than the NULL-case cost (194.13).
```

Maybe read that over to make sure you're satisfied, but I think even with the added queues, we should still be good here.

brandur:
Okay, I rebased this one to make the proposal a bit more current again in pursuit of coming up with a potential solution for #1105.

I had my LLM try to work some alternative solutions with me, and though it came up with a variety of alternatives, it didn't come up with anything substantially better. It would be better if you could have the client pick a value for leader domain based on each client's configured queues, but as we discussed yesterday, this doesn't work well operationally in case a queue is added or removed from a domain.

It did suggest renaming `LeaderDomain` to a less-abstract alternative like `ClientGroup`, and I think that's worth considering, but even that I'm not sure is substantially better. ("Client group" is okay in that it suggests that client's might be grouped together, but I kind of like that "leader domain" suggests concretely that this is related to leadership elections specifically. Maybe something like "leader group" would be better though.)

client.go
brandur: Yeah, good call. Modified the drivers to do that and added tests to each query in question.

client.go
brandur: Yeah, it's admittedly not super great. I think the only alternatives though involving storing something to the database so clients can compare notes with one another, which is also not great for other reasons.

Issue #1012, opened 2025-08-12
`river_job.max_attempts` (the database column) should have a default value of 25


brandur:
While writing documentation for how to insert a job using raw SQL, I came across the fact that while almost every `river_job` field gets a reasonable default, `max_attempts` is the odd man out and must have a value specified despite having a clear default of 25 in code.

Having a default on this column would be a slight improvement in that it'd make a manually inserted job a little easier to insert. Also, I think not having a default on this property may have been a mistake anyway since all other properties except the highly variable ones like `kind` and `args` have one.

Not worth its own migration, but let's ship the change opportunistically during the next migration that goes out.

Issue #875, opened 2025-05-03
Drop `river_client` + `river_client_queue`


brandur:
These were never put to work, and unless we find a reason they should exist soon, we should drop them on the next migration, especially now that `river_producer` is a thing.

PR #1115, opened 2025-12-26
Miscellaneous database cleanup

### LINKED ISSUE #874: `river_queue.updated_at` should be `NOT NULL` with a default, or nullable
An oddity I noticed while working on #870 in `river_queue`:

``` sql
CREATE TABLE river_queue (
    name text PRIMARY KEY NOT NULL,
    created_at timestamptz NOT NULL DEFAULT now(),
    metadata jsonb NOT NULL DEFAULT '{}' ::jsonb,
    paused_at timestamptz,
    updated_at timestamptz NOT NULL
);
```

`updated_at` is non-nullable, but also doesn't apply a default. That means that if you inserted a row without including a value for it, the insert would fail.

This is pretty weird. It should be one of:

1. Non-nullable but with a default, similar to `created_at`. So like: `updated_at timestamptz NOT NULL DEFAULT now()`.
2. Nullable (you could make the case that `updated_at` shouldn't be set on insert, and rather than only the first time it's updated after insert).

Probably (1) out of these two options.

This also applies to `river_client` and `river_client_queue` where the problem was copy/pasta'ed from `river_queue`, but these ones matter less because we'll probably end up dropping them anyway.

brandur:
Here, some miscellaneous database cleanup that we've been tracking for a
while. Nothing here is worth sending out by itself, but the next time we
do a database migration for something else, we can bundle these changes
in and send them out for free.

Fixes #874, #875, and #1012.


bgentry (approved):

brandur:
Rebased on #1224 and #1275.

brandur:
Awesome. Got Codex to combine all the migrations into a single 007 and it's looking pretty good. It's fairly well organized so you can see each distinct concern in its own section.

PR #1117, opened 2025-12-30
Initialize `nil` migration opts earlier

brandur:
The migration functions are supposed to allow you to call them with
nil options, but I noticed in some cases it's not working properly for
`MigrateTx` and it's possible to raise a nil pointer panic.

Here, initialize a nil `opts` earlier to prevent the problem. This makes
the code more obvious/conventional anyway in that it's easier to see
that nil options is supposed to be supported.

### DIFF
--- a/rivermigrate/river_migrate.go
+++ b/rivermigrate/river_migrate.go
@@ -385,6 +385,10 @@ func (m *Migrator[TTx]) ValidateTx(ctx context.Context, tx TTx) (*ValidateResult
 
 // migrateDown runs down migrations.
 func (m *Migrator[TTx]) migrateDown(ctx context.Context, exec riverdriver.Executor, direction Direction, opts *MigrateOpts, inOuterTx bool) (*MigrateResult, error) {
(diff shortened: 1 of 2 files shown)


rivermigrate/river_migrate.go
```diff
@@ -345,6 +349,10 @@ func (m *Migrator[TTx]) Migrate(ctx context.Context, direction Direction, opts *
 // Deprecated: Use Migrate instead. Certain migrations cannot be batched together
 // in a single transaction, so this method is not recommended.
 func (m *Migrator[TTx]) MigrateTx(ctx context.Context, tx TTx, direction Direction, opts *MigrateOpts) (*MigrateResult, error) {
+	if opts == nil {
+		opts = &MigrateOpts{}
+	}
```
brandur: Alternatively these could be right at the beginning of `migrateUp` and `migrateDown`, but I figure the earlier the better and other places in the code we have that do this tend to come right away.

PR #1120, opened 2026-01-11
Add shared `structtag` package for use across features

brandur:
Here, introduce a new `rivershared/structtag` package, extracted from
some of the previous implementation of `dbunique`. We modify the
previous implementation so that it can handle `river` tag values with
any arbitrary value (previously, only `unique` worked) so that we can
reuse the implementation or other features.

### DIFF
--- a/rivershared/structtag/struct_tag.go
+++ b/rivershared/structtag/struct_tag.go
@@ -1,4 +1,4 @@
-package dbunique
+package structtag
 
 import (
 	"fmt"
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -21,7 +21,10 @@ require (
 )
 
 require (
+	github.com/alecthomas/assert/v2 v2.11.0 // indirect
```
bgentry: ah, trying something other than testify!

brandur: Hah, oops.

Yeah, I was experimenting with other assert packages to see if we should be using something else for <secret work project> since we're right at the beginning of the project and it'd easy to make that change now (it seems like it should be possible to have something exactly like Testify except with a few niceties like generics??). I wanted to try things out in River first though to see how it'd look in a large-ish project.

I put it in, but then forgot to put it in a branch and mixed up a couple things with these `structtag` changes. I've reverted the `go.mod` changes to keep this change lean.

rivershared/structtag/struct_tag.go
```diff
@@ -29,23 +22,66 @@ func extractUniqueValues(encodedArgs []byte, uniqueKeys []string) []string {
...
+var (
+	// uniqueFieldsCache caches the unique fields for each JobArgs type. These are
+	// global to ensure that each struct type's tags are only extracted once.
+	uniqueFieldsCache = make(map[uniqueFieldCacheKey][]string) //nolint:gochecknoglobals
+	cacheMutex        sync.RWMutex                             //nolint:gochecknoglobals
+)
+
+// SortedFieldsWithTag retrieves unique fields with caching to avoid
+// extracting fields from the same struct type repeatedly.
+func SortedFieldsWithTag(args rivertype.JobArgs, tagValue string) ([]string, error) {
```
brandur, line 49: Also renamed this from `GetSortedTagFieldsCached` to `SortedFieldsWithTag`. Seems a little better for a public API name.

PR #1126, opened 2026-01-18
Have stuck detection account for worker-level timeouts as well as client-level

### LINKED ISSUE #1125: Stuck job evaluation doesn't take into account worker-level job timeouts
Jobs are considered "stuck" based on the timeout defined at the client level only, and not the timeout that can be set by individual workers (per [Worker-level job timeouts](https://riverqueue.com/docs/reliable-workers#worker-level-job-timeouts)).

## Repro

1. Create a worker that defines a timeout function greater than the client-level timeout
2. Start a job on that worker that runs for a time greater than the client-level timeout

```go
// configure river client to have a job timeout of 1 minute

type StuckJobArgs struct {}

func (StuckJobArgs) Kind() string {
	return "stuck"
}

type StuckWorker struct {
	river.WorkerDefaults[StuckJobArgs]
}

func (w *StuckWorker) Timeout(job *river.Job[StuckJobArgs]) time.Duration {
    // this job has a higher timeout that the default
	return 2 * time.Minute
}

func (w *StuckWorker) Work(ctx context.Context, job *river.Job[StuckJobArgs]) error {
	time.Sleep(2 * time.Minute)
}
```

In logs - note the timeout is not the custom one defined above:

```
WRN jobexecutor.JobExecutor: Job appears to be stuck source=river job_id=577 kind=stuck timeout=1m0s
INF producer: Producer job counts source=river num_completed_jobs=0 num_jobs_running=1 num_jobs_stuck=1 queue=default
```

## Expected

When a worker-level timeout is defined, the job should be considered stuck once it has exeeded the worker-level timeout.

## Actual

The worker-level timeout is not taken into account, and the job is considered stuck after it exceeds client-level timeout.

---

Thanks for your work on River!

brandur:
As reported by #1125, the stuck job detection mechanism currently only
considers client-level timeouts and without properly accounting for the
possibility of worker-level overrides.

Here, make sure to account for worker-level timeouts as well by hosting
the job timeout calculation further up in the executor logic.

Fixes #1125.

### DIFF
--- a/internal/execution/execution.go
+++ b/internal/execution/execution.go
@@ -2,7 +2,6 @@ package execution
 
 import (
 	"context"
-	"time"
 
 	"github.com/riverqueue/river/rivertype"
 )
@@ -12,19 +11,6 @@ import (
 type ContextKeyInsideTestWorker struct{}
 
 type Func func(ctx context.Context) error
-
-// MaybeApplyTimeout returns a context that will be cancelled after the given
-// timeout. If the timeout is <= 0, the context will not be timed out, but it
-// will still have a cancel function returned. In either case the cancel
-// function should be called after execution (in a defer).
-func MaybeApplyTimeout(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
-	// No timeout if a -1 was specified.
-	if timeout > 0 {
-		return context.WithTimeout(ctx, timeout)
-	}
-
-	return context.WithCancel(ctx)
-}
 
 // MiddlewareChain chains together the given middleware functions, returning a
 // single function that applies them all in reverse order.
--- a/internal/jobexecutor/job_executor.go
+++ b/internal/jobexecutor/job_executor.go
@@ -177,59 +177,6 @@ func (e *JobExecutor) execute(ctx context.Context) (res *jobExecutorResult) {
 	metadataUpdates := make(map[string]any)
 	ctx = context.WithValue(ctx, ContextKeyMetadataUpdates, metadataUpdates)
 
-	// Watches for jobs that may have become stuck. i.e. They've run longer than
-	// their job timeout (plus a small margin) and don't appear to be responding
-	// to context cancellation (unfortunately, quite an easy error to make in
-	// Go).
-	//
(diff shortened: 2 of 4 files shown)


internal/execution/execution.go
```diff
@@ -13,19 +12,6 @@ type ContextKeyInsideTestWorker struct{}
...
-// will still have a cancel function returned. In either case the cancel
-// function should be called after execution (in a defer).
-func MaybeApplyTimeout(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
-	// No timeout if a -1 was specified.
-	if timeout > 0 {
-		return context.WithTimeout(ctx, timeout)
-	}
-
-	return context.WithCancel(ctx)
-}
```
brandur, line 27: This was doing very little (it's the same as a `if jobTimeout > 0 {` check) and I needed the conditional for something else, so I just pulled it into `execute`.

internal/jobexecutor/job_executor.go
```diff
@@ -278,8 +225,15 @@ func (e *JobExecutor) execute(ctx context.Context) (res *jobExecutorResult) {
...
-		ctx, cancel := execution.MaybeApplyTimeout(ctx, jobTimeout)
-		defer cancel()
+
+		if jobTimeout > 0 {
+			var timeoutCancel context.CancelFunc
+			ctx, timeoutCancel = context.WithTimeout(ctx, jobTimeout)
+			defer timeoutCancel()
+
+			watchStuckCancel := e.watchStuck(ctx)
+			defer watchStuckCancel()
```
bgentry, line 235: IMO this is so much more readable than the large inline block before. I don't feel like I lose the context and flow of the surrounding code, while the fairly complex stuck job watching logic is nicely contained in a helper. 👏

bgentry (approved):
LGTM, great quick fix. Probably worthy of a quick `.1` release?

internal/jobexecutor/job_executor.go
brandur: Yeah +1. Should have done this from the beginning.

PR #1137, opened 2026-01-30
rivertest: enable metadata assertions

bgentry:
This enables basic assertions against metadata in `rivertest.RequireX` methods. The approach here isn't anything special: it allows a `map[string]any` to be specified and considers it a failure if all keys & values in it are not also in the metadata.

This should make it possible to build workflow-related assertions and other metadata assertions much more easily.

### DIFF
--- a/rivertest/rivertest.go
+++ b/rivertest/rivertest.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"slices"
 	"strings"
 	"testing"
@@ -49,6 +50,13 @@ type RequireInsertedOpts struct {
 	//
 	// No assertion is made if left the zero value.
 	MaxAttempts int
+
+	// Metadata is a subset of job metadata to assert against. Only the keys and
+	// values provided are compared, and any extra metadata on the job is
+	// ignored.
+	//
+	// No assertion is made if left nil or empty.
+	Metadata map[string]any
 
 	// Priority is the expected priority for the inserted job.
 	//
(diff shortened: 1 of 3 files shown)


rivertest/rivertest.go
```diff
@@ -50,6 +51,13 @@ type RequireInsertedOpts struct {
 	// No assertion is made if left the zero value.
 	MaxAttempts int
 
+	// Metadata is a subset of job metadata to assert against. Only the keys and
+	// values provided are compared, and any extra metadata on the job is
+	// ignored.
+	//
+	// No assertion is made if left nil or empty.
+	Metadata map[string]any
```
brandur, line 59: Just related to our convo last weekend, wanted to quickly get your thoughts on allowing people to do something like specify an exact match versus partial match.

If this became:

``` go
	Metadata MetadataMatcher
```

And you had something like:

``` go
type MetadataMatcher interface {
	CheckMatch(t testingT, actualBytes []byte, requireNotInserted bool) (bool, []string)
}

type metadataContainsMatcher struct {
	expected map[string]any
}

// MetadataContains checks that the given key-value pairs are present in job
// metadata, ignoring any extra keys that may also be there.
func MetadataContains(expected map[string]any) metadataContainsMatcher {
	return metadataContainsMatcher{expected: expected}
}

type metadataExactlyMatcher struct {
	expected map[string]any
}

// MetadataExactly checks that the given key-value pairs are present in job
// metadata, and that no extra keys are present.
func MetadataExactly(expected map[string]any) metadataExactlyMatcher {
	return metadataExactlyMatcher{expected: expected}
}
```

You get a pretty clean check that's a bit more readable than what's in here right now like:

``` go
			opts.Metadata = MetadataContains(map[string]any{
				"key": "value",
			})
```

The downside is that sometimes these interface-based checks can be a little hard to figure out how to use.

Thoughts?

PR #1138, opened 2026-01-31
split riverdrivertest.go

bgentry:
@brandur hit a point where I felt like this file is just too unwieldy. Instead of a giant 5000 line file, this splits them into smaller files grouped by entity. For job queries where we have many tests, break them into files by query type.

Let me know if you think this would be an improvement broadly. It's easy to tweak as it's a mostly AI-driven refactor. FWIW I also spun up a separate session and had it cross check every removed line to verify it was moved without being altered (other than setup details)—all good there. ✅


riverdriver/riverdrivertest/schema_introspection.go
```diff
@@ -0,0 +1,329 @@
...
+			// Postgres runs the drop with `CONCURRENTLY` so this must use a full
+			// schema rather than a transaction block.
+			driver, schema := driverWithSchema(ctx, t, nil)
+
+			err := driver.GetExecutor().IndexDropIfExists(ctx, &riverdriver.IndexDropIfExistsParams{
+				Index:  "does_not_exist",
+				Schema: schema,
+			})
+			require.NoError(t, err)
+		})
```
bgentry, line 118: @brandur hmm, this test was actually just an empty block before the split and Codex attempted to fix it. However it can't work in a tx as the other tests do because you can't run `DROP INDEX CONCURRENTLY` inside a transaction.

I filled it out using the same pattern as the above test to grab a full schema instead of txn.

brandur (approved):
Cool. Definitely nice to have the AI review as it'd be really hard to tell from the diff here if something was lost in the refactor.

I'm up for trying it! My only small concern is that you now have job-related operations broken into files that aren't necessarily suggestive from their name (i.e. you'd have to think for a bit to see that something like `JobCount` would be found in `job_read.go`). We should maybe look at an alternate scheme if that gets annoying.

PR #1144, opened 2026-02-11
use Go 1.25 + 1.26, bump golangci-lint

bgentry:
Keeping with our policy of supporting the latest major Go release as well as the prior one, this upgrades the project to a minimum of Go 1.25 now that 1.26 has been released.

In the process, I adopted a workaround for golangci-lint to pin it to run with 1.25 since its current build fails/panics against Go 1.26


cmd/river/go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/riverqueue/river/cmd/river
 
-go 1.24.0
+go 1.25.0
```
brandur, line 3: Similar to comment on other PR — should we just bump this to 1.26?

brandur (approved):

cmd/river/go.mod
bgentry: Per offline convo, this turns out to be the minimum supported Go version and users can't user install the package on a Go version lower than this.

PR #1148, opened 2026-02-13
Fix missing ticker.Stop() in pollForSettingChanges

GiGurra:
## Summary

- `pollForSettingChanges` creates a `time.NewTicker` but never calls `Stop()` on it when the context is cancelled, leaking the ticker's internal runtime timer
- Other functions in the same file (`heartbeatLogLoop`, `reportQueueStatusLoop`) correctly use `defer ticker.Stop()` — this aligns with that pattern
- Note: `pollForSettingChanges` only runs when no notifier is configured (poll-only mode), so this only affects non-LISTEN/NOTIFY setups

## Test plan

- [x] Existing `TestProducer_PollOnly` exercises this code path
- [x] Existing test suite passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1151, opened 2026-02-15
Add changelog entries for recent contributions + a couple minor tweaks

brandur:
Adds changelog entries for #1147, #1148, #1149, and #1150.

Also make a couple small tweaks: one to alphabetize a newly added
property, and one to move a success case test to above the error case,
which is a little more conventional.


bgentry (approved):

brandur:
thx!

PR #1155, opened 2026-02-22
fix JSON params in pgx text query modes

### LINKED ISSUE #1153: Can't use pgx5's QueryExecModeSimpleProtocol to avoid prepared statements when using connection pooling
Hi there,

We use PgBouncer as our connection pooler in transaction mode, which means we cannot use prepared statements. In our application, we disable prepared statements by using [QueryExecModeSimpleProtocol](https://pkg.go.dev/github.com/jackc/pgx/v5#QueryExecMode). I believe this is a common setup given the limitation of transaction mode. 

The problem is running the River client produces:
```
level=ERROR
msg="producer: Error fetching initial queue settings"
err="ERROR: invalid input syntax for type json (SQLSTATE 22P02)"
```

Which I believe is likely related to `river_queue.metadata` database column which is JSONB. According to pgx's comment on QueryExecModeSimpleProtocol:

```
 This is especially significant for []byte values. []byte values are encoded as PostgreSQL bytea. string must be used
instead for text type values including json and jsonb.
```

This means I'm pretty sure the following line is the issue:
https://github.com/riverqueue/river/blob/6b9b6223ffc8507ca25e13d90db3e6c4157fefe5/producer.go#L292

I appreciate this isn't a direct problem with River, though I think River's pgx driver would ideally know which exec mode is being used and use the according types?

Any advice would be much appreciated! Thank you.

bgentry:
River passes marshaled JSON query inputs as `[]byte` into sqlc-generated `pgx` calls. In `QueryExecModeSimpleProtocol` and `QueryExecModeExec` (including PgBouncer transaction-pooling setups), pgx treats `[]byte` bind args as `bytea`, so `json`/`jsonb` inputs can fail with invalid JSON syntax.

This change adapts JSON bind arguments from `[]byte` to `string` at wrapper `Exec`/`Query`/`QueryRow` boundaries, but only for pgx text execution modes. Query option parsing now mirrors pgx option semantics so per-query `QueryExecMode` overrides are honored. When a `pgx.QueryRewriter` is present, the driver wraps it so adaptation runs after rewrite against final SQL and bind args.

Explicit binary placeholders cast as `$n::bytea` or `CAST($n AS bytea)` are excluded from adaptation so intentional `bytea` inputs keep working in both extended and simple protocol paths. `defaultQueryExecMode` stays alongside `templateReplaceWrapper`, and the test-only `SharedTx.Conn()` now returns `nil` for capability probing instead of forcing panic matching.

Coverage includes driver tests for per-query mode overrides, `QueryRewriter` post-rewrite adaptation, `Exec`-path behavior, explicit `bytea` cast protection, and nil-conn fallback. `riverdrivertest` now exercises pgx endpoints in default, simple-protocol, and exec-mode configurations.

Fixes #1153.

### DIFF
--- a/riverdriver/riverdrivertest/driver_test.go
+++ b/riverdriver/riverdrivertest/driver_test.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
 	"github.com/jackc/pgx/v5/stdlib"
 	"github.com/lib/pq"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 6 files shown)


riverdriver/riverdrivertest/driver_test.go
```diff
@@ -92,9 +93,34 @@ func TestDriverRiverDatabaseSQLPgx(t *testing.T) {
...
 	t.Parallel()
 
+	// Default/primary pgx path with prepared statement caching.
+	t.Run("DefaultMode", func(t *testing.T) {
+		t.Parallel()
+
+		exerciseDriverRiverPgxV5WithMode(t, pgx.QueryExecModeCacheStatement)
+	})
+
+	// PgBouncer transaction-pooling compatibility path (`simple protocol`).
```
brandur: Nit, but is "simple protocol" in backticks like this intended? It seems like maybe it should've been just simple protocol without quotes, or something like `pgx.QueryExecModeSimpleProtocol`. It's repeated twice more below too which is what drew it to my attention.

bgentry: Thanks for catching these odd artifacts, I missed in my own review :pray: fixed

PR #1157, opened 2026-02-25
riverlog total size cap, perf and allocation optimizations

bgentry:
A user reported high transient memory usage and slow append cycles when using `riverlog` when the total size of its jsonb array grew large (especially in snooze-heavy jobs). It turns out we were doing a lot of highly inefficient stuff in here with lots of low-hanging fruit for optimization. To start, I added a benchmark to validate the perf troubles and potential solutions.

First, the biggest user-facing change is the new total history cap: `MiddlewareConfig.MaxTotalBytes` (default 8 MB). When appending a new `river:log` entry would exceed that limit, the middleware now prunes oldest entries first while always keeping the newest entry. That directly addresses unbounded log-history growth for long-lived/snoozed jobs.

Internally, the implementation was first refactored from its original struct-unmarshal/append logic to `json.RawMessage`; however this turned out to still be quite slow and allocation-heavy, so I did another pass leveraging `gjson/sjson` which was far better (results below).

A final round of optimization improved this significantly more using a one-pass capped-array builder which marshals only the new entry, computes exactly how much existing suffix can be kept, and emits the final JSON array once instead of append-then-repeated-delete loops. It also truncates log bytes before converting to string, so oversized buffers are not copied in full first.

Coverage was expanded with focused unit tests around append/prune semantics (missing key, non-array metadata, minimal pruning, newest-entry retention when over cap, and no-cap behavior).

In the end, we dramatically reduced the duration of these operations as well as the number of allocations, at the cost of a small increase in the total memory overhead per operation.

Stage | Change | 256KB ns/op | 256KB B/op | 256KB allocs/op | 2MB ns/op | 2MB B/op | 2MB allocs/op
-- | -- | -- | -- | -- | -- | -- | --
Baseline (23921c2) | Original append path + benchmark harness | 1,331,725 | 703,634 | 294 | 8,348,696 | 6,441,894 | 2,100
Interim (ecb718b) | Internal []json.RawMessage path | 1,562,610 (+17.3%) | 674,228 (-4.2%) | 292 (-0.7%) | 12,557,009 (+50.4%) | 6,433,251 (-0.1%) | 2,097 (-0.1%)
Interim (c87f47e) | gjson/sjson append+prune | 627,022 (-52.9%) | 1,166,917 (+65.8%) | 24 (-91.8%) | 4,397,872 (-47.3%) | 6,480,502 (+0.6%) | 30 (-98.6%)
Final (5896050) | One-pass suffix selection + single-build output | 163,320 (-87.7%) | 884,066 (+25.6%) | 31 (-89.5%) | 1,207,252 (-85.5%) | 7,006,968 (+8.8%) | 46 (-97.8%)

### DIFF
--- a/riverlog/river_log.go
+++ b/riverlog/river_log.go
@@ -13,7 +13,6 @@ import (
 	"log/slog"
 
 	"github.com/tidwall/gjson"
-	"github.com/tidwall/sjson"
 
 	"github.com/riverqueue/river/internal/jobexecutor"
 	"github.com/riverqueue/river/rivershared/baseservice"
@@ -28,7 +27,7 @@ const (
 	// Hard ceiling to prevent pathological allocations from extreme configs.
 	maxTotalSizeMaxMB = 64
 	maxTotalBytesMax  = maxTotalSizeMaxMB * 1024 * 1024
-	metadataKey    = "river:log"
+	metadataKey       = "river:log"
 )
 
 type contextKey struct{}
@@ -155,10 +154,7 @@ func defaultConfig(config *MiddlewareConfig) *MiddlewareConfig {
 	}
 
 	config.MaxSizeBytes = cmp.Or(config.MaxSizeBytes, maxSizeBytes)
-	config.MaxTotalBytes = cmp.Or(config.MaxTotalBytes, maxTotalBytes)
-	if config.MaxTotalBytes > maxTotalBytesMax {
-		config.MaxTotalBytes = maxTotalBytesMax
-	}
+	config.MaxTotalBytes = min(cmp.Or(config.MaxTotalBytes, maxTotalBytes), maxTotalBytesMax)
 
 	return config
 }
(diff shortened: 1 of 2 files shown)


riverlog/river_log.go
```diff
@@ -165,46 +178,182 @@
...
+		}
+	}
+
+	return len(bounds)
+}
+
+func buildAppendedArray(existingArrayBytes []byte, bounds []arrayElementBounds, keepStart int, newEntryBytes []byte) []byte {
+	var totalLen int
+	if keepStart >= len(bounds) {
+		totalLen = 2 + len(newEntryBytes)
```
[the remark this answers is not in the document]
bgentry: Heh, I think we turned this on at some point as part of SOC2. Unsurprisingly, like the rest of Copilot it's not quite right. Codex 5.2 summary:

> On the overflow feedback: I corroborated the code paths, and this looks like a generic static-analysis warning rather than a practical bug in our case.
> 
> The flagged math is in [river_log.go:333](app://-/index.html#) and [river_log.go:337](app://-/index.html#).
> newEntryBytes is bounded by truncation at [river_log.go:197](app://-/index.html#).
> keptContentLen is derived from parsed offsets at [river_log.go:336](app://-/index.html#) based on array bytes from metadata; this is not unbounded arithmetic on attacker-sized integers.
> We also already document expected JSONB size constraints at [river_log.go:195](app://-/index.html#).

That being said, we didn't have test coverage for the case where a very large payload comes back from the DB, so I added that.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `riverlog.Middleware` now supports `MiddlewareConfig.MaxTotalBytes` (default 8 MB) to cap total persisted `river:log` history per job. When the cap is exceeded, oldest log entries are dropped first while retaining the newest entry. Values over 64 MB are clamped to 64 MB. [PR #1157](https://github.com/riverqueue/river/pull/1157).
```
bgentry, line 12: @brandur I should have called this out earlier, does this 8MB default seem reasonable to you? Do you feel it should be called out as "breaking" or is "changed" sufficient?

We can tweak it in a follow up before release if needed.

brandur: Yeah, I saw it in there so I can't complain. I could see it potentially on the low side, but it really depends on the use case. How did you pick 8 MB anyway?

Just re-reading this, we might want to update this comment on `MaxSizeBytes` and call out that you'll likely hit `MaxTotalBytes` way before you hit something like 50 MB:

``` go
	// Be careful with this number because the maximum total log size is equal
	// to maximum number of attempts multiplied by this number (each attempt's
	// logs are kept separately). For example, 25 * 2 MB = 50 MB maximum
	// theoretical log size. Log data goes into metadata which is a JSONB field,
	// and JSONB fields have a maximum size of 255 MB, so any number larger than
	// 255 divided by maximum number of attempts may cause serious operational
	// problems.
```

PR #1166, opened 2026-03-13
Add `make bench` target, run in CI, fix failure

bgentry:
Fixing three separate issues here:

* We don't have a `make bench` task in this project's `Makefile`, so I added one.
* Configured it to run in CI with low `ITERATIONS=1` to ensure it works but isn't slow.
* Lo and behold, a benchmark was broken because we weren't running them in CI. Fix it by properly setting the query's schema.

### DIFF
--- a/riverdriver/riverdrivertest/driver_test.go
+++ b/riverdriver/riverdrivertest/driver_test.go
@@ -441,6 +441,7 @@ func BenchmarkDriverRiverPgxV5_Executor(b *testing.B) {
 				ClientID:  clientID,
 				MaxToLock: 100,
 				Queue:     river.QueueDefault,
+				Schema:    bundle.schema,
 			}); err != nil {
 				b.Fatal(err)
 			}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -17,6 +17,52 @@ on:
...
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 2s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+
+    steps:
+      - uses: actions/checkout@v4
```
brandur: Probably doesn't matter much, but it's up to v6 these days.

brandur (approved):

.github/workflows/ci.yaml
bgentry: Ahhh thank you, fixing! :pray:

bgentry: Actually we had a few other outdated major versions in here, I fixed them all 🚀

PR #1175, opened 2026-03-15
Quote schema name in operations to support special characters and spaces

### LINKED ISSUE #1170: Using upper case schemas do not correctly
During tests we generate random schema names so that we can run tests in parallel. We generated them based on the regex in river, which allows upper case letters. The problem is that river migrations do not quote schemas (nor table names) so PostgreSQL internally converts them all to lowercase. Which can lead to conflicting schemas where we would expect that they would be independent. Resulting in flaky tests.

Or migration scripts should quote schemas or only lower case schemas should be allowed.

brandur:
Came up in #1170: we don't quote schemas in SQL operations, and that can
lead to problems in use of spaces, special characters, and uppercase
characters. It's not the end of the world given use of the above can be
considered a sizable anti-pattern anyway, but use of quoting is good for
general correctness.

Fixes #1170.

### DIFF
--- a/riverdriver/riverdatabasesql/river_database_sql_driver.go
+++ b/riverdriver/riverdatabasesql/river_database_sql_driver.go
@@ -24,6 +24,7 @@ import (
 	"github.com/riverqueue/river/riverdriver/riverdatabasesql/internal/dbsqlc"
 	"github.com/riverqueue/river/rivershared/sqlctemplate"
 	"github.com/riverqueue/river/rivershared/uniquestates"
+	"github.com/riverqueue/river/rivershared/util/dbutil"
 	"github.com/riverqueue/river/rivershared/util/ptrutil"
 	"github.com/riverqueue/river/rivershared/util/savepointutil"
 	"github.com/riverqueue/river/rivershared/util/sliceutil"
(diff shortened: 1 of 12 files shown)


riverdriver/riverdatabasesql/river_database_sql_driver.go
```diff
@@ -143,7 +143,8 @@ func (e *Executor) Exec(ctx context.Context, sql string, args ...any) error {
 func (e *Executor) IndexDropIfExists(ctx context.Context, params *riverdriver.IndexDropIfExistsParams) error {
 	var maybeSchema string
 	if params.Schema != "" {
-		maybeSchema = params.Schema + "."
+		// quote the schema name to support uppercase schema names and odd characters
+		maybeSchema = `"` + params.Schema + `".`
```
mitar: Isn't proper quoting that you have to replace also any `"` with `""` inside schema name?

I think defining a quoting function and then using it everywhere would be the easiest.

riverdriver/riverdrivertest/riverdrivertest.go
```diff
@@ -36,6 +36,7 @@ func Exercise[TTx any](ctx context.Context, t *testing.T,
 	exerciseSQLFragments(ctx, t, executorWithTx)
 	exerciseExecutorTx(ctx, t, driverWithSchema, executorWithTx)
 	exerciseSchemaIntrospection(ctx, t, driverWithSchema, executorWithTx)
+	exerciseSchemaName(ctx, t, driverWithSchema)
```
brandur, line 39: This new category doesn't feel super great. That said, I started with adding a specific test case for `JobInsert` (since I don't want to have to test this for every single operation), and that didn't feel great either.

riverdriver/riverdatabasesql/river_database_sql_driver.go
brandur: Good point. Replaced these instances with a new `dbutil.SafeIdentifier` function. I thought about using `pgx.Identifier` too, but it's a tad awkward to use, and I didn't want to pull in the dependency for the non-Pgx drivers. I checked its implementation and luckily it's quite simple.

riverdriver/riverdrivertest/riverdrivertest.go
bgentry: Yeah, I would say it could be grouped with other structural / foundational stuff but we don't really have a group for that yet 🤔

PR #1203, opened 2026-04-12
Add hook `HookQueueStateCount` + read middleware as hooks and vice versa

brandur:
Here, add a new hook called `HookQueueStateCount` which gets invoked to
produce job queue count statistics. We do this by adding a new
maintenance service which like other maintenance services, runs only on
the leader, so we only have one client performing counts at any given
time. Furthermore, in order to not introduce a potential operational
problem without opt-in from River users, the counts only run if a
`HookQueueStateCount` hook/middleware has been added to the client.

The reason we do all this to to implement a feature requested by one of
users: for `otelriver` in contrib to be able to emit queue count
metrics, which seems like a pretty reasonable ask for the package to be
able to do, and something that every River user would likely want access
to in their ops charts.

A slight oddity, but which I think is _probably_ okay, is that the new
hook ideally stays a hook, but the existing `otelriver` middleware is a
middleware. It'd be nice not to have to put `otelriver.Middleware` into
both a client's `Hooks` and `Middleware` configuration, so we modify
client to allow for hooks that middleware and middleware which are
hooks. This lets `otelriver.Middleware` continue doing what it was
already doing, but also to start producing new counts as a hook.

### DIFF
--- a/client.go
+++ b/client.go
@@ -9,6 +9,7 @@ import (
 	"log/slog"
 	"os"
 	"regexp"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -657,6 +658,7 @@ type clientTestSignals struct {
 	periodicJobEnqueuer   *maintenance.PeriodicJobEnqueuerTestSignals
 	queueCleaner          *maintenance.QueueCleanerTestSignals
 	queueMaintainerLeader *maintenance.QueueMaintainerLeaderTestSignals
+	queueStateCounter     *maintenance.QueueStateCounterTestSignals
 	reindexer             *maintenance.ReindexerTestSignals
 }
 
@@ -678,6 +680,9 @@ func (ts *clientTestSignals) Init(tb testutil.TestingTB) {
 	}
 	if ts.queueMaintainerLeader != nil {
 		ts.queueMaintainerLeader.Init(tb)
+	}
+	if ts.queueStateCounter != nil {
+		ts.queueStateCounter.Init(tb)
 	}
 	if ts.reindexer != nil {
 		ts.reindexer.Init(tb)
@@ -759,7 +764,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 		config:               config,
 		driver:               driver,
 		hookLookupByJob:      hooklookup.NewJobHookLookup(),
-		hookLookupGlobal:     hooklookup.NewHookLookup(config.Hooks),
+		hookLookupGlobal:     nil, // initialized below after cross-referencing with middleware
 		producersByQueueName: make(map[string]*producer),
 		testSignals:          clientTestSignals{},
 		workCancel:           func(cause error) {}, // replaced on start, but here in case StopAndCancel is called before start up
@@ -780,9 +785,9 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 	// the more abstract config.Middleware for middleware are set, but not both,
 	// so in practice we never append all three of these to each other.
 	{
-		middleware := config.Middleware
+		middlewares := config.Middleware
 		for _, jobInsertMiddleware := range config.JobInsertMiddleware {
-			middleware = append(middleware, jobInsertMiddleware)
+			middlewares = append(middlewares, jobInsertMiddleware)
 		}
 	outerLoop:
 		for _, workerMiddleware := range config.WorkerMiddleware {
@@ -798,16 +803,44 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 				}
 			}
 
-			middleware = append(middleware, workerMiddleware)
+			middlewares = append(middlewares, workerMiddleware)
 		}
 
-		for _, middleware := range middleware {
+		for _, middleware := range middlewares {
 			if withBaseService, ok := middleware.(baseservice.WithBaseService); ok {
 				baseservice.Init(archetype, withBaseService)
 			}
 		}
 
-		client.middlewareLookupGlobal = middlewarelookup.NewMiddlewareLookup(middleware)
+		// Cross-reference hooks and middleware: any middleware that also
+		// implements Hook is added to hooks, and any hook that also implements
+		// Middleware is added to middleware. Deduplication prevents double
+		// registration when the same struct is passed to both Config.Hooks and
+		// Config.Middleware.
+		hooks := config.Hooks
+
+		for _, middleware := range middlewares {
+			if hook, ok := middleware.(rivertype.Hook); ok {
+				// Only add if this middleware isn't already in hooks (it may
+				// have been passed to both config properties).
+				alreadyInHooks := slices.Contains(hooks, hook)
+				if !alreadyInHooks {
+					hooks = append(hooks, hook)
+				}
+			}
+		}
+
+		for _, hook := range config.Hooks {
+			if middleware, ok := hook.(rivertype.Middleware); ok {
+				alreadyInMiddleware := slices.Contains(middlewares, middleware)
+				if !alreadyInMiddleware {
+					middlewares = append(middlewares, middleware)
+				}
+			}
+		}
+
+		client.hookLookupGlobal = hooklookup.NewHookLookup(hooks)
+		client.middlewareLookupGlobal = middlewarelookup.NewMiddlewareLookup(middlewares)
 	}
 
 	pluginDriver, _ := driver.(driverPlugin[TTx])
@@ -959,6 +992,16 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 			}, driver.GetExecutor())
 			maintenanceServices = append(maintenanceServices, queueCleaner)
 			client.testSignals.queueCleaner = &queueCleaner.TestSignals
+		}
+
+		{
+			queueStateCounter := maintenance.NewQueueStateCounter(archetype, &maintenance.QueueStateCounterConfig{
+				HookLookupGlobal: client.hookLookupGlobal,
+				QueueNames:       maputil.Keys(config.Queues),
+				Schema:           config.Schema,
+			}, driver.GetExecutor())
+			maintenanceServices = append(maintenanceServices, queueStateCounter)
+			client.testSignals.queueStateCounter = &queueStateCounter.TestSignals
 		}
 
 		{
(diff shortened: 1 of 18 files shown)


brandur:
@codex review

client.go
```diff
@@ -807,6 +811,26 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
...
+		// implements Hook is added to hooks, and any hook that also implements
+		// Middleware is added to middleware. It may have been better to have
+		// Client only take one combined set, but we only realized that we may
+		// need a combined hook/middleware implementation after that ship sailed.
+		hooks := config.Hooks
+
+		for _, mw := range middleware {
+			if hook, ok := mw.(rivertype.Hook); ok {
+				hooks = append(hooks, hook)
+			}
```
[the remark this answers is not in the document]
brandur: Reasonable to check. Fixed.

riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
```diff
@@ -86,42 +86,14 @@ FROM /* TEMPLATE: schema */ river_job
...
-LEFT JOIN
-    running_job_counts ON all_queues.queue = running_job_counts.queue
-LEFT JOIN
-    available_job_counts ON all_queues.queue = available_job_counts.queue
-ORDER BY all_queues.queue ASC;
+    queue,
+    state,
+    COUNT(*) AS count
+FROM /* TEMPLATE: schema */river_job
+WHERE (COALESCE(cardinality(@queue_names::text[]), 0) = 0 OR queue = ANY(@queue_names::text[]))
```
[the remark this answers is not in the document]
brandur: Took the empty check out as it's not needed in the implementation anymore.

riverdriver/river_driver_interface.go
```diff
@@ -202,7 +202,7 @@ type Executor interface {
 
 	JobCancel(ctx context.Context, params *JobCancelParams) (*rivertype.JobRow, error)
 	JobCountByAllStates(ctx context.Context, params *JobCountByAllStatesParams) (map[rivertype.JobState]int, error)
-	JobCountByQueueAndState(ctx context.Context, params *JobCountByQueueAndStateParams) ([]*JobCountByQueueAndStateResult, error)
```
bgentry, line 205: This is a breaking change for the UI. It has a few callers in https://github.com/riverqueue/riverui/blob/fa202780c19504cb8510d0b2c8d92f8ff31d8dc9/handler_api_endpoint.go

I was going to hold off on asking about this, but in light of this break, I'm wondering if we should give any thought to whether or not the UI runs its own queries for this purpose or somehow reuses logic in River so that we don't have multiple redundant queriers. But I think it probably is going to be too hard to try and share that state among backend nodes and UI frontend nodes.

bgentry (commented):
It should probably be straightforward to update riverui to work with these changes, but we should make sure before merging it since it breaks an API it uses.

riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
```diff
@@ -86,42 +86,14 @@ FROM /* TEMPLATE: schema */ river_job
...
-LEFT JOIN
-    available_job_counts ON all_queues.queue = available_job_counts.queue
-ORDER BY all_queues.queue ASC;
+    queue,
+    state,
+    COUNT(*) AS count
+FROM /* TEMPLATE: schema */river_job
+WHERE queue = ANY(@queue_names::text[])
+GROUP BY queue, state
+ORDER BY queue, state;
```
bgentry, line 96: Also I believe this might break the prior query's efficient usage of the prioritized fetch index; I'm running an agent w/ benchmarks now to get some more details on whether this is a significant perf regression.

bgentry: Yeah so I opened #1211 with a perf benchmark showing this is a huge regression. Essentially the old split format was able to effectively leverage the index whereas this variant does not:

GPT 5.4 high explanation:

> The new query is inefficient for the current schema because it filters on `queue` and groups on `(queue, state)`, but the important existing btree on `river_job` is keyed `(state, queue, priority, scheduled_at, id)`. PostgreSQL btree indexes are most useful when the query constrains the leftmost columns first, and this query no longer does that. Because `state` is not fixed in the predicate, the planner cannot use the index as a selective access path in the same way, so it tends to fall back to a parallel sequential scan and aggregate after reading a large slice of the table.
> 
> The old query shape was more appropriate for the existing index because it split the work into separate branches with constant `state` predicates, specifically `state = 'available'` and `state = 'running'`. That matches the left edge of the `(state, queue, ...)` index, and once Postgres has a fixed `state`, it can efficiently narrow by `queue` within the same index. In practice that gave us index-only or bitmap/index-driven plans instead of scanning the table broadly, which is why the old query performed much better on selective queue lists.
> 
> If we want to fix this without adding indexes, the main option is to reshape the new query so it restores a constant-`state` predicate per branch. Concretely, that means counting each state in its own grouped subquery, then `UNION ALL`ing those results together or joining them back to the requested queue list. That is more verbose than the current grouped query, but it realigns the SQL with the index we already have and gives the planner a much better chance of using it efficiently.

The benchmark is probably useful regardless of how we proceed on this PR but I'll let you decide how to proceed on it ✌️

brandur: Commented here [1]. If it's indeed faster to iterate an index N times for N different states in the system than it is just to iterate the table once and count things, let's show it in the bench.

[1] https://github.com/riverqueue/river/pull/1211#issuecomment-4260217301

riverdriver/river_driver_interface.go
brandur: I changed this back to be a little more like it was before (uses `JobCountByQueueAndStateResult` still), but yeah, it's still breaking. It'll be really easy to fix though, so I think we should just stick to not worrying about breaking changes in the drivers that much.

And +1. I thought about if there might be some way to share this with the UI as well ... but unfortunately couldn't come up with a great way to do this. The obvious thing to do would be to write the results to some kind of temporary table somewhere, but obviously that's a much larger change and would probably produce some objections.

Another possibility would be to try and leverage listen/notify: the counter counts periodically, then blasts the counts off every so often to all listeners who'd cache them and expose via UI. I think this could work, but would need a fallback in case the leader was currently down or in cases where listen/notify isn't available.

riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
brandur: Alright, I ended up changing the query to this:

```
SELECT queue, 'available'::text AS state, COUNT(*) AS count FROM /* TEMPLATE: schema */river_job WHERE queue = ANY(@queue_names::text[]) AND state = 'available' GROUP BY queue
UNION ALL
SELECT queue, 'cancelled', COUNT(*) FROM /* TEMPLATE: schema */river_job WHERE queue = ANY(@queue_names::text[]) AND state = 'cancelled' GROUP BY queue
UNION ALL
SELECT queue, 'completed', COUNT(*) FROM /* TEMPLATE: schema */river_job WHERE queue = ANY(@queue_names::text[]) AND state = 'completed' GROUP BY queue
UNION ALL
SELECT queue, 'discarded', COUNT(*) FROM /* TEMPLATE: schema */river_job WHERE queue = ANY(@queue_names::text[]) AND state = 'discarded' GROUP BY queue
UNION ALL
SELECT queue, 'pending', COUNT(*) FROM /* TEMPLATE: schema */river_job WHERE queue = ANY(@queue_names::text[]) AND state = 'pending' GROUP BY queue
UNION ALL
SELECT queue, 'retryable', COUNT(*) FROM /* TEMPLATE: schema */river_job WHERE queue = ANY(@queue_names::text[]) AND state = 'retryable' GROUP BY queue
UNION ALL
SELECT queue, 'running', COUNT(*) FROM /* TEMPLATE: schema */river_job WHERE queue = ANY(@queue_names::text[]) AND state = 'running' GROUP BY queue
UNION ALL
SELECT queue, 'scheduled', COUNT(*) FROM /* TEMPLATE: schema */river_job WHERE queue = ANY(@queue_names::text[]) AND state = 'scheduled' GROUP BY queue
ORDER BY queue, state;
```

It's still a bit ugly, but not as ugly as the previous one. It generally performs a little better according to the benchmark (though I modified the benchmark so the previous count query counts all states). Here's the notes from Claude:

```
⏺ At 1k rows it's faster than CTE at 1-2 queues and roughly comparable at 10-20 queues. Summary across both scales:

  ┌────────┬─────────────────────────────┬───────────────────────────────┐
  │ Queues │  1k rows (UNION ALL / CTE)  │  100k rows (UNION ALL / CTE)  │
  ├────────┼─────────────────────────────┼───────────────────────────────┤
  │ 1      │ 172µs / 715µs (4.2x faster) │ 326µs / 796µs (2.4x faster)   │
  ├────────┼─────────────────────────────┼───────────────────────────────┤
  │ 2      │ 182µs / 612µs (3.4x faster) │ 606µs / 1061µs (1.8x faster)  │
  ├────────┼─────────────────────────────┼───────────────────────────────┤
  │ 10     │ 243µs / 118µs (0.5x)        │ 1838µs / 2351µs (1.3x faster) │
  ├────────┼─────────────────────────────┼───────────────────────────────┤
  │ 20     │ 184µs / 153µs (0.8x)        │ 3530µs / 3874µs (1.1x faster) │
  └────────┴─────────────────────────────┴───────────────────────────────┘

  The UNION ALL approach wins at every scale at 100k rows and at low queue counts at 1k rows. It only loses slightly to the CTE
  at 1k rows with many queues (where both are already sub-millisecond). It's also half the line count of the CTE approach. Each
  UNION ALL branch gets its own targeted index scan on (state, queue), same as the CTE, but without the overhead of
  materializing 8 CTEs and 8 LEFT JOINs.
```

brandur: I tried to make the SQLite version closer to this as well, but sqlc is still buggy as heck when it comes to SQLite. It probably won't matter as much for a local SQLite DB, so probably okay to leave for now. Easy to fix up in case sqlc fixes ever come through.

brandur:
@bgentry Alright, want to take a look at the latest changes here and see if you're happier with them? We do add a count that happens more frequently, but it should be as performant as the old version if all was equal (though we count more states in this one). Keep in mind too that the count only activates if there's someone listening for it (currently, this would require that the latest version of the OTEL middleware be plugged in).

PR #1205, opened 2026-04-12
Make `leaderResignedChan` buffered to protect against hang

brandur:
This one's aimed at fixing an intermittently failing tests that we're
occasionally seeing in CI:

https://github.com/riverqueue/river/actions/runs/24312334093/job/70984031771?pr=1203

It was possible for the test case to end to `leaderResignedChan` before
`keepLeadershipLoop` entered its select, which with an unbuffered
channel, could block forever. Changing this to a buffered channel lets
`keepLeadershipLoop` consume the value as its leisure while making sure
the send never blocks.

### DIFF
--- a/internal/leadership/elector.go
+++ b/internal/leadership/elector.go
@@ -136,8 +136,10 @@ func (e *Elector) Start(ctx context.Context) error {
 		return nil
 	}
 
-	// We'll send to this channel anytime a leader resigns on the key with `name`
-	e.leaderResignedChan = make(chan struct{})
+	// We'll send to this channel anytime a leader resigns on the key with `name`.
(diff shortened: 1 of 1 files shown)


brandur:
@codex review

Admittedly, Claude did all the debugging work on this one. I figure we can cross-reference the proposed solution with Codex and make sure it's all above bar.

brandur:
Man, have hit three intermittently failing tests in a row now ... and they're all different lol. I'm not sure if we regressed recently on this front or what.

brandur:
Actually, two of these are related to the leadership changes I made recently, so they are pretty new.

bgentry (approved):
Sorry, I thought I already stamped this one 🤦‍♂️

PR #1207, opened 2026-04-12
Buffered `requestResignChan`

brandur:
Aims to fix another intermittent test failure:

https://github.com/riverqueue/river/actions/runs/24312716933/job/70985012980?pr=1205

    --- FAIL: TestElector_WithNotifier (0.00s)
        --- FAIL: TestElector_WithNotifier/RequestResignStress (20.41s)
            elector_test.go:350: Generated postgres schema "leadership_2026_04_12t17_52_43_schema_04" with migrations [1 2 3 4 5 6] on line "main" in 229.626141ms [4 generated] [0 reused]
            elector_test.go:352: Starting test_client_id
            elector_test.go:375: Requesting leadership resign
            elector_test.go:375: Requesting leadership resign
            elector_test.go:375: Requesting leadership resign
            elector_test.go:375: Requesting leadership resign
            elector_test.go:375: Requesting leadership resign
            test_signal.go:95: timed out waiting on test signal after 10s
            test_signal.go:95: timed out waiting on test signal after 10s
            riverdbtest.go:293: Checked in postgres schema "leadership_2026_04_12t17_52_43_schema_04"; 1 idle schema(s) [5 generated] [3 reused]
    FAIL
    FAIL	github.com/riverqueue/river/internal/leadership	20.948s

The problem is that when `requestResignChan` is unbuffered, if
`keepLeadershipLoop` hasn't yet entered its `select`, then the `default`
statement on the `select` below will cause all senders (we have 5
competing senders in the `RequestResignStress` test case) to fall
through without sending anything:

    select {
    case <-ctx.Done():
    case e.requestResignChan <- struct{}{}:
    default:
            // if context is not done and requestResignChan has an item in it
            // already, do nothing
    }

By making the channel buffered, we guarantee at least one sender gets a
message through, and we don't end up hanging the test.

### DIFF
--- a/internal/leadership/elector.go
+++ b/internal/leadership/elector.go
@@ -139,7 +139,9 @@ func (e *Elector) Start(ctx context.Context) error {
 	// We'll send to this channel anytime a leader resigns on the key with `name`
 	e.leaderResignedChan = make(chan struct{})
 
-	e.requestResignChan = make(chan struct{})
+	// Buffered to 1 so a send from handleLeadershipNotification doesn't block
+	// if keepLeadershipLoop hasn't entered its select yet.
+	e.requestResignChan = make(chan struct{}, 1)
 
 	var sub *notifier.Subscription
 	if e.notifier == nil {
--- a/internal/leadership/elector_test.go
+++ b/internal/leadership/elector_test.go
@@ -344,6 +344,31 @@ func testElector[TElectorBundle any](
 		startstoptest.Stress(ctx, t, elector)
 	})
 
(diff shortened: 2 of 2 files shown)


brandur:
@codex review

brandur:
@bgentry Again, kind of hoping that we can rely on Claude/Codex to cross reference their work and root out any tricky concurrency problems here.

bgentry (approved):
Good fix. Not something we can reasonably add a test for, right?

brandur:
> Good fix. Not something we can reasonably add a test for, right?

Yeah, actually turns out we can write a test for this! Added.

PR #1208, opened 2026-04-13
fix race condition in batch completer stats

bgentry:
Batch completion could race when multiple job completions reused the
same `JobStatistics` pointer. Short excerpt from the race detector
failure in a recent CI run:

    WARNING: DATA RACE
    Write at ... BatchCompleter.handleBatch.func4()
    .../internal/jobcompleter/job_completer.go:482
    Previous read ... jobStatisticsFromInternal()
    .../event.go:75

`BatchCompleter` queued the caller's stats pointer directly and later
mutated `CompleteDuration` before publishing updates. Batch job error
paths can reuse one stats object across multiple completions, so a later
batch could mutate stats that were already visible to subscribers.

Take a snapshot of `JobStatistics` when queuing the completion so each
published update owns an immutable copy. Add regression coverage that
queues two completions with a shared stats pointer and verifies that
each published update has a distinct snapshot.

### DIFF
--- a/internal/jobcompleter/job_completer.go
+++ b/internal/jobcompleter/job_completer.go
@@ -513,7 +513,9 @@ func (c *BatchCompleter) JobSetStateIfRunning(ctx context.Context, stats *jobsta
 	c.setStateParamsMu.Lock()
 	defer c.setStateParamsMu.Unlock()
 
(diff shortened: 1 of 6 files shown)


brandur (approved):
Cool. Yep, makes sense.

PR #1210, opened 2026-04-13
Reindexer: Mock reindex function to reduce parallel contention

brandur:
This one's aimed at fixing an intermittently failing test case:

https://github.com/riverqueue/river/actions/runs/24322049420/job/71009998068?pr=1208

    --- FAIL: TestReindexer (0.00s)
        --- FAIL: TestReindexer/ReindexesConfiguredIndexes (10.07s)
            reindexer_test.go:219: Reusing idle postgres schema "maintenance_2026_04_13t01_54_14_schema_04" [user facing: "maintenance_2026_04_13t01_54_14_schema_04"] after cleaning in 26.436456ms [4 generated] [7 reused]
            test_signal.go:95: timed out waiting on test signal after 10s
            logger.go:256: time=2026-04-13T01:54:27.581Z level=INFO msg="maintenance.Reindexer: Signaled to stop during index build; attempting to clean up concurrent artifacts"
            riverdbtest.go:293: Checked in postgres schema "maintenance_2026_04_13t01_54_14_schema_04"; 1 idle schema(s) [4 generated] [10 reused]
        --- FAIL: TestReindexer/ReindexesMinimalSubsetofIndexes (10.14s)
            reindexer_test.go:183: Reusing idle postgres schema "maintenance_2026_04_13t01_54_14_schema_01" [user facing: "maintenance_2026_04_13t01_54_14_schema_01"] after cleaning in 28.042877ms [4 generated] [10 reused]
            test_signal.go:95: timed out waiting on test signal after 10s
            reindexer_test.go:211:
                    Error Trace:	/home/runner/work/river/river/internal/maintenance/reindexer_test.go:211
                    Error:      	Should be false
                    Test:       	TestReindexer/ReindexesMinimalSubsetofIndexes
            logger.go:256: time=2026-04-13T01:54:28.444Z level=INFO msg="maintenance.Reindexer: Signaled to stop during index build; attempting to clean up concurrent artifacts"
            riverdbtest.go:293: Checked in postgres schema "maintenance_2026_04_13t01_54_14_schema_01"; 1 idle schema(s) [5 generated] [24 reused]
    FAIL
    FAIL	github.com/riverqueue/river/internal/maintenance	18.764s

I'm diagnosing with Claude's help here, but what appears to be happening
is that although a reindex operation in Postgres is often fast, it is
still a heavy operation, and can slow down even further when there's a
lot of concurrent activity hammering a database.

Many reindexer test cases run in parallel, and it appears that was
happening here is that we got a reindex that exceeded our maximum
timeout of 10x. We have some evidence this during the test run from the
runtime of 10.07s and the line:

    Signaled to stop during index build; attempting to clean up concurrent artifacts

Here, I'm putting forward a solution proposed by Claude, which is to
mock out the reindex operation, especially where we have a number of
reindexes going in parallel. The tests `ReindexOneSuccess` and
`ReindexSkippedWithReindexArtifact` still put load on the real reindex
operation, so we're not going full mock here, and should see our
intermittency considerably reduced while still being confident that
everything still works.

### DIFF
--- a/internal/maintenance/reindexer_test.go
+++ b/internal/maintenance/reindexer_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
@@ -24,6 +25,7 @@ type reindexerExecutorMock struct {
 	indexesExistCalls  atomic.Int32
 	indexesExistFunc   func(ctx context.Context, params *riverdriver.IndexesExistParams) (map[string]bool, error)
 	indexesExistSignal chan struct{}
+	indexReindexFunc   func(ctx context.Context, params *riverdriver.IndexReindexParams) error
 }
 
 func newReindexerExecutorMock(exec riverdriver.Executor) *reindexerExecutorMock {
@@ -31,6 +33,7 @@ func newReindexerExecutorMock(exec riverdriver.Executor) *reindexerExecutorMock 
 		Executor:           exec,
 		indexesExistFunc:   exec.IndexesExist,
 		indexesExistSignal: make(chan struct{}, 10),
+		indexReindexFunc:   exec.IndexReindex,
 	}
 }
 
@@ -43,6 +46,10 @@ func (m *reindexerExecutorMock) IndexesExist(ctx context.Context, params *riverd
 	}
 
 	return m.indexesExistFunc(ctx, params)
+}
+
+func (m *reindexerExecutorMock) IndexReindex(ctx context.Context, params *riverdriver.IndexReindexParams) error {
+	return m.indexReindexFunc(ctx, params)
 }
 
 func TestReindexer(t *testing.T) {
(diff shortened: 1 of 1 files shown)


internal/maintenance/reindexer_test.go
```diff
@@ -182,6 +189,20 @@ func TestReindexer(t *testing.T) {
 
 		svc, bundle := setup(t)
 
+		// Mock IndexReindex so the test doesn't depend on the speed of real
+		// REINDEX CONCURRENTLY operations on the shared CI database. Track
+		// which indexes got reindexed so we can verify the expected set.
+		mockExec := newReindexerExecutorMock(bundle.exec)
+		var reindexedNames []string
+		var reindexedMu sync.Mutex
```
bgentry: you'd normally stack these in a `var` block, right?

bgentry (approved):
Yeah this seems appropriate; we can maintain coverage that the reindexing actually works while mocking out where needed to avoid timing issues in tests.

internal/maintenance/reindexer_test.go
brandur: Totally. Thanks, fixed!

PR #1213, opened 2026-04-14
refactor leader election around DB-issued "terms"

bgentry:
The old elector treated leadership as a lease held by one client ID and renewed over time. That kept the happy path simple, but it left too much implicit. One leadership term was not clearly separated from the next, so reelection and resignation were not scoped to a specific term. That made same-client reacquisition and stale cleanup harder to reason about than they needed to be.

This rewrite makes the database issue explicit leadership terms using the columns we already have. `leader_id` still identifies the client, while `(leader_id, elected_at)` identifies one specific term. Elect, reelect, and resign now all operate on that exact term and return the leader row directly from the database.

The elector still keeps a bounded local trust window for its last successful confirmation, but that window is now anchored to the attempt that produced it rather than to when a response happened to arrive. That keeps slow successful reelections from stretching leadership past the real lease budget, while still avoiding direct app-vs-database clock comparisons in the state machine.

This branch also adds an explicit `Time.Now` API at the base-service layer. The old shared API only exposed `NowUTC`, even for code doing local deadline and duration math. For the elector that was the wrong abstraction, because these comparisons need to preserve Go's monotonic `time.Time` reading rather than normalize through UTC. Making that API explicit lets the elector use the right local clock for trust-window calculations while keeping `NowUTCOrNil` for the existing test path that passes stubbed wall-clock timestamps through to database code.

The tests were rewritten around the final model instead of the intermediate mechanics. The shared driver suite now covers term-scoped elect, reelect, and resign behavior across PostgreSQL and both SQLite backends, including same-client term replacement and stale-term rejection. The elector tests focus on observable behavior such as leadership handoff, resign requests, ordered subscriber notifications, and stepping down cleanly when the trust window expires.

This also rolls up the branch's earlier flake investigation and keeps the original CI reference for the shared-schema failures that led to the redesign:
https://github.com/riverqueue/river/actions/runs/24406465152
https://github.com/riverqueue/river/actions/runs/24537567759/job/71736101198

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed leader election to track explicit database-issued leadership terms, reducing handoff flakiness and same-client reacquisition edge cases while making reelection and resign target the current leadership lease instead of a stale one. [PR #1213](https://github.com/riverqueue/river/pull/1213).
+
 ## [0.34.0] - 2026-04-08
 
 ### Added
--- a/internal/leadership/doc.go
+++ b/internal/leadership/doc.go
@@ -1,0 +1,117 @@
+// Package leadership implements leader election for River clients sharing a
+// database schema.
+//
+// The database records at most one current leadership term at a time. The
+// elected client runs distributed maintenance work such as queue management,
+// job scheduling, and reindexing that should not be duplicated across clients.
+//
+// # Overview
+//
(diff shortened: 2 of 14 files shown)


bgentry:
@codex review

bgentry:
@brandur took a whole bunch of turns and iteration to get this to where I want it but I think it's pretty solid now! Pared back a lot of the added complexity from earlier just by going all in on "DB is source of truth including for time".

brandur (approved):
Cool! Okay actually looked at this again, and it becomes all pretty unobjectionable without the expansions to the time API. Thanks for taking a look at it!

bgentry:
@brandur awesome, thank you!

PR #1215, opened 2026-04-16
Rename baseservice.NowUTC to Now

bgentry:
After opening #1213 I thought it'd be better to instead fix the underlying time API.

The shared `baseservice` clock exposed `NowUTC`, `NowUTCOrNil`, and `StubNowUTC`, but those names overstated what the production path was actually doing. The unstubbable implementation already returned `time.Now()` rather than normalizing through `UTC()`, so the API was implying a UTC guarantee that callers were not actually getting. This renames the clock APIs to `Now`, `NowOrNil`, and `StubNow` so the abstraction matches its real job: providing a process-local clock for deadline and duration math, plus an optional stubbed wall-clock value for database-facing test paths.

This also makes the monotonic-clock compatibility explicit in the documentation. If the production implementation were changed to normalize through `UTC()`, it would strip Go’s monotonic reading and make the API a bad fit for in-process elapsed-time calculations. The relevant Go documentation is here: [Monotonic Clocks in the `time` package](https://pkg.go.dev/time#hdr-Monotonic_Clocks):

> Because t.In, t.Local, and t.UTC are used for their effect on the interpretation of the wall time, they also strip any monotonic clock reading from their results. The canonical way to strip a monotonic clock reading is to use t = t.Round(0).

Fortunately we weren't actually relying on the UTC behavior since it wasn't being applied in prod, so it's an easy switch. Only two small test updates were needed beyond the simple bulk renames.


bgentry:
@codex review

brandur (approved):
WFM!

PR #1216, opened 2026-04-17
Ignore TLS error when closing listener from the notifier

### LINKED ISSUE #256: Intermittent listener close issue popping up again
We definitely were getting hit with the previous hot loop issue mentioned in issue [#248](https://github.com/riverqueue/river/issues/248). After upgrading to `0.0.25` we did see the hot loop issue resolved with PR [#246](https://github.com/riverqueue/river/pull/246), but it seems ~15 minutes or so we're still getting the `failed to send closeNotify alert` issue sporadically. While the listener is able to reconnect after a while, was wondering if anyone had any updates on a resolution.

Example log:
Screenshot 2024-03-08 at 2 28 10 PM

brandur:
Here, address a request coming from #256 in which it's fairly common to
see an like this one going to the error logs:

> `tls: failed to send closeNotify alert (but connection was closed anyway): write tcp 10.0.5.131:40192->10.0.7.96:5432: i/o timeout`

If you read the error message carefully, it indicates that this really
isn't a problem. The connection is closed already, and the TLS package
just wasn't able to close it as cleanly as it wanted to.

Here, modify the site that calls listener close in the notifier so that
it ignores these types of errors, thereby cleaning up peoples' error
logs somewhat.

Fixes #256.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Ignore errors like `tls: failed to send closeNotify alert (but connection was closed anyway)` when closing listeners. [PR #1216](https://github.com/riverqueue/river/pull/1216).
+
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


bgentry (approved):
Looks good and definitely worth trying, thanks!

brandur:
thx!

PR #1228, opened 2026-04-25
Add "table repacker" maintenance services that runs `REPACK (CONCURRENTLY)`

brandur:
Here, add a maintenance service modeled on the reindexer that runs
`REPACK (CONCURRENTLY)`, a new feature slated for Postgres 19 [1]. This
is an exciting new feature that can fully reset a table without needing
to take an exclusive lock the way that `VACUUM FULL` does.

This feature may be on the riskier side at first, so I've disabled it by
default, requiring an explicit `RepackerEnabled: true` to activate.

Similar to the reindexer it runs once a day. I've staggered its start
time to 1 AM UTC so it's offset from the reindexer's start time of
midnight.

This relies on Postgres 19 to work, so it's not really suitable for
release yet, but I've put in fallback code that causes it to run `VACUUM
FULL` for Postgres prior to 19. (This is obviously not suitable for use,
but it gets the test matrix running.)

[1] https://github.com/postgres/postgres/commit/28d534e2ae0ac888b5460f977a10cd9bb017ef98

### DIFF
--- a/client.go
+++ b/client.go
@@ -280,6 +280,27 @@ type Config struct {
 	// than working them. If it's specified, then Workers must also be given.
 	Queues map[string]QueueConfig
 
+	// RepackerEnabled enables the table repacker maintenance service, which
+	// periodically runs REPACK (CONCURRENTLY) on the river_job table to reclaim
+	// space from dead tuples. This is not enabled by default.
+	//
+	// REPACK (CONCURRENTLY) requires Postgres >= 19. On older versions, the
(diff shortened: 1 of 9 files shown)


brandur:
Actually, two addendums upon further thought:

* If the repacker is on, we'd want to disable the reindexer for that table. Indexes are rebuilt on repack, so it'd be added work for now good reason.
* We may just want to make the repacker + reindexer the same service since they're fairly interrelated. It'd also guarantee only one of them is doing anything at any given time.

PR #1231, opened 2026-04-26
Fix accidentally inverted logic in listener log check

brandur:
Follows up #1216 with a fix whereby the conditional on the check on
whether to log an error was accidentally inverted.


bgentry (approved):
Whoops, sorry for not catching in review

brandur:
Nope, no worries. Thanks!

PR #1235, opened 2026-04-29
Add QueueBundle.Remove() to stop and remove queues

tigrato:
Adds dynamic queue removal at runtime. When removed, the queue's producer is stopped and jobs can no longer be worked on that queue. Queues can be re-added after removal.

### DIFF
--- a/client.go
+++ b/client.go
@@ -768,6 +768,7 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 
 	client.queues = &QueueBundle{
 		addProducer:             client.addProducer,
+		removeProducer:          client.removeProducer,
 		clientFetchCooldown:     config.FetchCooldown,
 		clientFetchPollInterval: config.FetchPollInterval,
 		clientWillExecuteJobs:   config.willExecuteJobs(),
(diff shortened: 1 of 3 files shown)


tigrato:
cc @brandur @magaldima

brandur:
Cool! Thx. A few small asks on this one:

* Can you add some additional validation similar to what's found in the `Add` function? For example, check that the client will run jobs:

    ``` go
    if !b.clientWillExecuteJobs {
            return errors.New("client is not configured to execute jobs, cannot add queue")
    }
    ```

    And also, return an error in case the queue doesn't exist.

* Can you make it explicit in the dockblock that the function will block until all jobs in the given queue have been worked? And furthermore that this blocking may affect other operations like shutdown.

* Can you drop the trivial comments found in the test suite? e.g.

    ``` go
    // Remove the queue
    err = client.Queues().Remove(queueName)
    ```

* Just since it looks like the code is LLM-generated anyway, could you see if you can get it to write the tests without the sleeps?

Anything else you want to add @bgentry?

brandur:
And actually, while looking at this discovered a minor concurrency bug around producer map access. I patched that in #1236 — can you rebase, and make sure to lock that mutex in the `removeProducer` function in `Client`? (Thanks!)

tigrato:
@brandur thank your for the review. changes pushed.
the tests were indeed LLM generated but the code lives in one of our forks since 2022. It's now time to push it upstream to reduce the maintaining cost

brandur:
@tigrato Thanks! Going to pull this, but I'm going to make a few tweaks on top of it before release. I think the biggest one is that since `QueuesBundle.Remove` is a blocking operation with an uncertain deadline, it should take a context like `Stop` does already which allows it to be cancelled out of. Hopefully this won't be too big of a problem in terms of an update on your end!

brandur (approved):

brandur:
@tigrato Actually, just before I do, would you mind signing our CLA over at https://github.com/riverqueue/rivercla?

PR #1236, opened 2026-05-01
Patch unsafe producer map access

brandur:
While doing a little work on #1235, my LLM flagged that the way that
we're accessing `producersByQueueName` today is already a little unsafe
as it may be read in multiple places concurrently. There's a mutex to
protect it somewhat in the `QueueBundle`, but it may still race between
a change there a "notify producer" send.

Here, add one additional `RWMutex` that makes sure to synchronize read
access on the map.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


bgentry (approved):
ah, yeah map this was originally not being mutated outside initialization but that changed with the ability to add a new queue dynamically at runtime. Clearly we missed the concurrency safety around this! 🤖

brandur:
ty!

PR #1240, opened 2026-05-05
A few tweaks to `QueueBundle.Remove` implementation

brandur:
This builds on #1235 to bring in a few tweaks:

* Removing a queue is a blocking operation because it needs to wait for
  the producer to finish up its jobs and shut down. It'd be better to
  provide a way for this not to block forever, so here we add a context
  parameter to `QueueBundle.Remove` similar to the one taken by
  `Client.Stop`. If the context becomes done before the producer
  resolves, `QueueBundle.Remove` falls through with the error.

* Add a "stress" test case for `QueueBundle.Remove`. It's meant to
  detect a deadlock or other concurrency bug in case there is one and
  gives us a little more confidence that what we have here is right.

* Renamed `addProducer` and `removeProducer` to `producerAdd` and
  `producerRemove` so they sort more nicely against each other.

* Add changelog entry.

### DIFF
--- a/client.go
+++ b/client.go
@@ -767,11 +767,11 @@ func NewClient[TTx any](driver riverdriver.Driver[TTx], config *Config) (*Client
 	}
 
 	client.queues = &QueueBundle{
-		addProducer:             client.addProducer,
-		removeProducer:          client.removeProducer,
 		clientFetchCooldown:     config.FetchCooldown,
 		clientFetchPollInterval: config.FetchPollInterval,
 		clientWillExecuteJobs:   config.willExecuteJobs(),
+		producerAdd:             client.producerAdd,
+		producerRemove:          client.producerRemove,
 	}
 
 	baseservice.Init(archetype, &client.baseService)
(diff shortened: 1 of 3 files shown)


tigrato (commented):
LGTM. Thank you!

brandur:
@bgentry Ah doh, forgot to add you as a reviewer here. Seem okay to you?

client.go
```diff
@@ -2219,7 +2219,17 @@ func (c *Client[TTx]) removeProducer(queueName string) error {
...
+	shouldStop, stopped, finalizeStop := producer.StopInit()
+	if shouldStop {
+		select {
+		case <-ctx.Done():
+			finalizeStop(false)
+			return ctx.Err()
+		case <-stopped:
+			finalizeStop(true)
+		}
+	}
```
bgentry, line 2231: I'm not totally clear on some of the edge cases here. What happens if the producer is stopped, but the context also gets cancelled? Does the producer actually keep fully running, or do we go into an indeterminate state?

bgentry (approved):
LGTM aside from the major question, I just want to make sure this is bulletproof and clearly documented.

client.go
brandur: @bgentry Hm, so `StopInit` will only return `shouldStop = true` if the producer is actually started. So in the case of a stopped producer, this block gets skipped, and the context is never checked. This shouldn't really happen though because the producer should get removed from `c.producersByQueueName` when it's stopped, so subsequent calls to `Remove` will return `QueueNotFoundError`. So I think not checking the cancelled context is probably okay here given it's an invalid call. Does that answer your question?

brandur: I modified the docs to make it clear what a user should try to do in this case and added a test case that verifies what happens in this circumstance (`Queues_Remove_ContextCancelledThenRetry`).

brandur:
Thanks!

PR #1242, opened 2026-05-06
Mark schema template replacements as Stable to enable Replacer cache

### LINKED ISSUE #1241: schemaTemplateParam should mark schema replacement as Stable: true to enable Replacer cache
Hey folks! We've been tracking down a memory leak in our production Go service that uses River v0.33+ with `riverpgxv5`, and we traced it back to the `sqlctemplate.Replacer` cache never being used for schema replacements.

## What's happening

The `schemaTemplateParam` function in `riverpgxv5` passes `Stable: false` (the default) on the schema replacement:

```go
// riverdriver/riverpgxv5/river_pgx_v5_driver.go
func schemaTemplateParam(ctx context.Context, schema string) context.Context {
    return sqlctemplate.WithReplacements(ctx, map[string]sqlctemplate.Replacement{
        "schema": {Value: schema},  // Stable defaults to false
    }, nil)
}
```

Because of this, `replacerCacheKeyFrom` always returns `cacheEligible = false`, and `ReplaceAllStringFunc` runs on **every single query** — even though the schema value never changes for the lifetime of a River client.

The caching mechanism from #802 exists and works correctly, it's just never activated because `Stable` isn't set.

## Impact we observed

We run River in two deployments (an API that enqueues jobs and a worker that processes them). After profiling with `go tool pprof` in production:

- **Worker** (runs `river.Start()` with continuous polling): `regexp.ReplaceAllStringFunc` grew **+29MB in 16 minutes**, accounting for 99% of all memory growth
- **API** (insert-only client): `regexp.ReplaceAllStringFunc` grew **+16.9MB in 20 minutes**, accounting for 89% of all memory growth

The strings allocated by the regex are short-lived but under production throughput, the allocation rate outpaces GC and downstream consumers (like pgx tracers or OTel span attributes) can hold references long enough for the memory to accumulate.

We confirmed the fix by patching `schemaTemplateParam` locally via a `go.mod` replace directive — after deploying with `Stable: true`, memory graphs in our monitoring flatlined.

## Suggested fix

One-line change in `schemaTemplateParam`:

```go
"schema": {Value: schema, Stable: true},
```

The same applies to the `ColumnExists` method which also passes a schema replacement without `Stable`.

The schema value is derived from the client's config and is constant for the lifetime of the River client — it's the textbook case for `Stable: true`.

This wouldn't affect queries that use dynamic replacements like `where_clause` and `order_by_clause` (e.g., `JobList`, `JobDeleteMany`) — those correctly remain uncached since their replacements don't set `Stable`.

## Environment

- River: v0.33.0 (also confirmed on v0.35.1)
- riverpgxv5
- Go 1.26.2
- pgx v5.9.2

Happy to open a PR if you'd like!

## Screenshot of our pods in sandbox environment before and after the fix

Image

IgorDeo:
schemaTemplateParam (and the schema replacement in ColumnExists) never set Stable: true, so the sqlctemplate.Replacer cache introduced in #802 is bypassed on every query. This causes regexp.ReplaceAllStringFunc to run on every SQL query, allocating a new string each time.

The schema value comes from the client config and is constant for the lifetime of a River client, so marking it Stable is safe. With this change, the regex runs once per unique SQL string and then serves from cache.

Fixes all three drivers: riverpgxv5, riverdatabasesql, riversqlite.

Fixes #1241

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Mark schema replacements as `Stable` in sqlc templates, preventing query SQL from having to be reallocated over and over again.. [PR #1242](https://github.com/riverqueue/river/pull/1242).
 - Fix unsafe concurrent producer map access in client. [PR #1236](https://github.com/riverqueue/river/pull/1236).
 - Fix bug in `sqltemplate` cached path in order in which named args are passed to a query (previously, the order was unstable). [PR #1243](https://github.com/riverqueue/river/pull/1243).
 
--- a/riverdriver/riverdatabasesql/river_database_sql_driver.go
+++ b/riverdriver/riverdatabasesql/river_database_sql_driver.go
@@ -126,7 +126,7 @@ func (e *Executor) ColumnExists(ctx context.Context, params *riverdriver.ColumnE
 		schema = "'" + params.Schema + "'"
 	}
 	ctx = sqlctemplate.WithReplacements(ctx, map[string]sqlctemplate.Replacement{
-		"schema": {Value: schema},
+		"schema": {Value: schema, Stable: true},
 	}, nil)
 
 	exists, err := dbsqlc.New().ColumnExists(ctx, e.dbtx, &dbsqlc.ColumnExistsParams{
@@ -1247,6 +1247,6 @@ func schemaTemplateParam(ctx context.Context, schema string) context.Context {
 	}
 
 	return sqlctemplate.WithReplacements(ctx, map[string]sqlctemplate.Replacement{
-		"schema": {Value: schema},
+		"schema": {Value: schema, Stable: true},
 	}, nil)
 }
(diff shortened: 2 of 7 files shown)


brandur (commented):
Thanks Igor! Looks great.

Would you mind adding an entry under the "fixed" section in the CHANGELOG. Maybe just something like:

``` md
- Mark schema replacements as `Stable` in sqlc templates, preventing query SQL from having to be reallocated over and over again.. [PR #1242](https://github.com/riverqueue/river/pull/1242).
```

IgorDeo:
> Thanks Igor! Looks great.
> 
> Would you mind adding an entry under the "fixed" section in the CHANGELOG. Maybe just something like:
> 
> ```
> - Mark schema replacements as `Stable` in sqlc templates, preventing query SQL from having to be reallocated over and over again.. [PR #1242](https://github.com/riverqueue/river/pull/1242).
> ```

Done -> [6d208db](https://github.com/riverqueue/river/pull/1242/commits/6d208dba63264942899810640bfa9d3566bbd724)

brandur (approved):
Great. Thanks again!

IgorDeo:
I'm trying to run the tests locally to debug these failures. I see that one of them passed on the third try

brandur:
@IgorDeo Actually, this looks like it revealed an sqlctemplate bug by accident. I'll take the fix.

brandur:
Opened #1243.

brandur:
@IgorDeo Hey, just brought in #1243. Would you mind pulling master and rebasing to see if that fixes it?

PR #1243, opened 2026-05-08
Fix bug in `sqltemplate` cache path around args order

brandur:
Related to #1242. Activating caching everywhere revealed a bug in that
we were adding named args using `maputil.Values`. The problem with this
is that `maputil.Values` returns values in an arbitrary order, resulting
in all kinds of odd things potentially happening during a query as args
are switched around.

Here, fix the problem by appending args in the same order as incoming
map keys.


brandur:
@bgentry Mind taking a look at this one? Small bug fix.

bgentry (approved):

PR #1249, opened 2026-05-11
Add `initTestConfig` helper for more succinct example tests

brandur:
Here, add an `initTestConfig` helper for example tests that wraps up a
lot of the common testing logic in each. This leaves the example tests
more succinct, takes a lot of test-only code out of the equation to keep
the signal higher for readers, and makes broadly refactoring example
tests easier as changes may need to only be made in one place. This is
very similar to the change we already made in River Pro.

### DIFF
--- a/example_batch_insert_test.go
+++ b/example_batch_insert_test.go
@@ -3,16 +3,12 @@ package river_test
 import (
 	"context"
 	"fmt"
-	"log/slog"
-	"os"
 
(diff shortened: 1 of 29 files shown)


bgentry (approved):
I was just thinking of how useful it would be to pull this change over while I was reviewing the Resumable Jobs PR. Thank you for doing so!

brandur:
Awesome man. Thanks!

PR #1251, opened 2026-05-13
Add 30s timeout to PeriodicJobEnqueuer insertBatch

NanoBjorn:
Previously, `PeriodicJobEnqueuer.insertBatch` passed the raw service context directly to `Begin`/`Insert`/`Commit`, with no per-batch deadline (unlike all other calls to the db). If the underlying connection stalled, the enqueuer would hang indefinitely, halting all periodic job insertion until the process was restarted. **We have observed this issue multiple times during RW-RO failovers or temporary network unavailability, stalls in our case could last from minutes to hours**

Wrapped the DB calls in `context.WithTimeout(ctx,
riversharedmaintenance.TimeoutDefault)` (30s), matching the pattern already used by `JobScheduler` and `JobRescuer`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Wrap `PeriodicJobEnqueuer.insertBatch` database calls in a 30-second timeout. Previously, a stalled pgx `Begin`/`Insert`/`Commit` could hang the periodic enqueuer indefinitely, halting all periodic job insertion until the process was restarted or leader re-elected.
+
 ## [0.37.0] - 2026-05-11
 
 ### Added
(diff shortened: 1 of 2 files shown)


NanoBjorn:
@brandur I would also like to hotfix this into older version we use (v0.18.0), can we do this after merge?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Wrap `PeriodicJobEnqueuer.insertBatch` database calls in a 30-second timeout. Previously, a stalled pgx `Begin`/`Insert`/`Commit` could hang the periodic enqueuer indefinitely, halting all periodic job insertion until the process was restarted or leader re-elected.
```
brandur: Could you add a link to the PR here like the other entries have?

NanoBjorn:
cc @bgentry

brandur (commented):
Oops, forgot to hit the submit button on my review.

CHANGELOG.md
NanoBjorn: added

brandur (approved):

PR #1259, opened 2026-05-27
Add options to migrator `Validate` functions + option for `TargetVersion`

### LINKED ISSUE #1257: Allow specifying target version when validating migrations
We use the [`(*Migrator[TTx]).Validate`](https://pkg.go.dev/github.com/riverqueue/river@v0.37.1/rivermigrate#Migrator.Validate) to check on startup if the database migrations have been applied.

We want to roll out River Pro v0.24.0, which means we want to stop at migration v6 for the `pro` line.

We can pass a [`MigrateOpts`](https://pkg.go.dev/github.com/riverqueue/river@v0.37.1/rivermigrate#MigrateOpts) to the [`Migrate`](https://pkg.go.dev/github.com/riverqueue/river@v0.37.1/rivermigrate#Migrator.Migrate) method to actually perform the migration, but there doesn't seem to be a way to do the same when validating, so presumably we're going to get an error because v7 hasn't been applied yet.

As a workaround I think we can compare the results of [`AllVersions`](https://pkg.go.dev/github.com/riverqueue/river@v0.37.1/rivermigrate#Migrator.AllVersions) and [`ExistingVersions`](https://pkg.go.dev/github.com/riverqueue/river@v0.37.1/rivermigrate#Migrator.ExistingVersions) but it is a bit of a pain having to reimplement the validation logic.

brandur:
This one in response to #1257. It's useful to validate the current state
of the database's migrations, but you may not be fully on the latest
version and could get an error. The included example is for Pro v0.24.0
which includes two migrations: one to be rolled out before deploy and
one to be rolled out after. `Validate` would fail until the post-deploy
migrations were run, which isn't flexible enough.

Here, have `Validate` functions take a `ValidateOpts` similar to the
`MigrateOpts` taken by the `Migrate` functions. `ValidateOpts` includes
`TargetVersion`, which lets you specify a maximum version bound.

Notably, this is a breaking change in that we add a new parameter to
`Validate`. I'm a little on the fence about this, but doing it via
breaking (1) avoids an extra pair of functions that pollute the API and
need testing, (2) should be *mostly* low impact because I doubt all that
many people are using `Validate`, and (3) the fix is super easy in that
you just need to add a `nil` in the last parameter place if you don't
want to use any new options.

Fixes #1257.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+⚠️ **Breaking API change:** `rivermigrate.Migrator.Validate` and `rivermigrate.Migrator.ValidateTx` now take a `*rivermigrate.ValidateOpts` parameter. Pass `nil` to preserve previous behavior. We normally endeavor not to make any breaking API changes, but this one will keep the API in a much nicer state, and is on an ancillary function that most installations won't be using. [PR #1259](https://github.com/riverqueue/river/pull/1259)
+
+### Added
+
+- Added `rivermigrate.ValidateOpts.TargetVersion` so validation can check migrations up to a specific target version, matching the target-version behavior available on `Migrate` and `MigrateTx`. Notably, this is a breaking API change as the validate functions previously didn't take any options. [PR #1259](https://github.com/riverqueue/river/pull/1259)
+
 ## [0.38.0] - 2026-05-22
 
 ### Added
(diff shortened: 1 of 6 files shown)


bgentry (approved):
Yeah I think this is the best choice 👍 Does it require a matching pro change or paired release?

brandur:
@bgentry Yeah, doh, we might need to do a paired release just because the Pro project is affixed to a specific River version, so Pro users won't be able to upgrade to our latest unless that's also bumped.

brandur:
Thanks!

PR #1261, opened 2026-05-30
Make up migrations with `TargetVersion` idempotent

### LINKED ISSUE #1260: Migration fails if already migrated up to target version
We're running River migrations on each deployment with [`(*Migrator[TTx]).Migrate`](https://pkg.go.dev/github.com/riverqueue/river@v0.37.1/rivermigrate#Migrator.Migrate). As we are in the process of rolling out River Pro v0.24.0, we specify `TargetVersion: 6` in the `MigrateOpts` for the `pro` line.

This worked fine the first time it was run, but on subsequent deployments we get an error

```
Error: version 6 is not in target list of valid migrations to apply
```

It seems to me that if the target version has already been reached, calling `Migrate` should be a no-op rather than an error, which would match the behaviour when `TargetVersion` is not specified.

brandur:
Here, try to address #1260 by making up migrations with `TargetVersion`
no longer error if we find that the database has already reached that
target version. These now return idempotently instead as a user
convenience.

Down migrations with `TargetVersion` already worked with this way, so
although we add a test case to verify this, we don't have to do anything
else.

Fixes #1260.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ⚠️ **Breaking API change:** `rivermigrate.Migrator.Validate` and `rivermigrate.Migrator.ValidateTx` now take a `*rivermigrate.ValidateOpts` parameter. Pass `nil` to preserve previous behavior. We normally endeavor not to make any breaking API changes, but this one will keep the API in a much nicer state, and is on an ancillary function that most installations won't be using. [PR #1259](https://github.com/riverqueue/river/pull/1259)
 
-### Added
+### Changed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,9 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ⚠️ **Breaking API change:** `rivermigrate.Migrator.Validate` and `rivermigrate.Migrator.ValidateTx` now take a `*rivermigrate.ValidateOpts` parameter. Pass `nil` to preserve previous behavior. We normally endeavor not to make any breaking API changes, but this one will keep the API in a much nicer state, and is on an ancillary function that most installations won't be using. [PR #1259](https://github.com/riverqueue/river/pull/1259)
 
-### Added
+### Changed
```
brandur, line 12: Think "changed" is actually a bit more appropriate for both these changelog entries.

bgentry (approved):

PR #1263, opened 2026-05-31
Add `statement_timeout` on fetching available jobs

brandur:
Follows up #1255 to add a `statement_timeout` in addition to the Go
context timeout. `statement_timeout` will give us a better error
message, and also minimizes the chances of accidentally locking rows
that won't be work if we had an operation that ran long, succeeded, but
then was immediately cancelled as Go's context timeout ran out.

`statement_timeout` is Postgres only, so the code is a little gnarlier
than would be desirable in that we add a `SetLocalStatementTimeout`
function to driver `ExecutorTx`, but which is a no-op on some databases
like SQLite. We try to clarify in documentation that it needs to be used
in addition to context timeout, not instead of it, because it may no-op
depending on the database.

I also increased the timeout to 30 seconds. This matches our timeouts in
the various maintenance modules, and seems a little safer as job locks
on tables with huge numbers of dead rows could potentially take over 10
seconds, and maybe some users have this happening.

IMO, it's too random still where we put this stuff in, but we'll have to
figure that out on follow up changes. e.g. Why do we do a statement
timeout for locking jobs but not for maintenance operations? Hard to
justify.

### DIFF
--- a/riverdriver/riverdatabasesql/river_database_sql_driver.go
+++ b/riverdriver/riverdatabasesql/river_database_sql_driver.go
@@ -1050,6 +1050,10 @@ func (t *ExecutorTx) Rollback(ctx context.Context) error {
 	return t.tx.Rollback()
 }
 
+func (t *ExecutorTx) SetLocalStatementTimeout(ctx context.Context, timeout time.Duration) error {
+	return t.Exec(ctx, "SELECT set_config('statement_timeout', $1, true)", riverdriver.PostgresStatementTimeoutValue(timeout))
+}
+
 type ExecutorSubTx struct {
 	Executor
 
@@ -1101,6 +1105,10 @@ func (t *ExecutorSubTx) Rollback(ctx context.Context) error {
 	}
 
 	return nil
+}
+
+func (t *ExecutorSubTx) SetLocalStatementTimeout(ctx context.Context, timeout time.Duration) error {
+	return t.Exec(ctx, "SELECT set_config('statement_timeout', $1, true)", riverdriver.PostgresStatementTimeoutValue(timeout))
 }
 
 func interpretError(err error) error {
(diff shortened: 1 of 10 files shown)


bgentry:
@codex review

riverdriver/riverdatabasesql/river_database_sql_driver.go
```diff
@@ -1050,6 +1050,10 @@ func (t *ExecutorTx) Rollback(ctx context.Context) error {
 	return t.tx.Rollback()
 }
 
+func (t *ExecutorTx) SetLocalStatementTimeout(ctx context.Context, timeout time.Duration) error {
+	return t.Exec(ctx, "SELECT set_config('statement_timeout', $1, true)", timeout.String())
```
bgentry: I think `time.Duration.String` will return Go duration values that aren't pg-compatible?

bgentry (commented):
Having spent a bit more time thinking this over, I'm really going back and forth on whether this should be necessary. In the past we've talked about giving productionization guidance that includes bits like "always set a reasonable statement timeout on your connection pool", and as you mentioned we probably do want to have some sort of timeout for _all_ operations just to prevent infinite hangs, right?

People will forget to apply customizations like that, though, even if they're aware of the need to do so.

brandur:
Okay, just so we don't get this release stuck, opened a simpler version with a couple fixes over at #1268.

riverdriver/riverdatabasesql/river_database_sql_driver.go
brandur: I pushed a change that uses milliseconds instead, but it's a questionable — either you add code complication or you make the assumption that only common durations will get used here and rely on the test suite to catch bad ones.

Either way, won't ship this for now so it doesn't matter.

Issue #1258, opened 2026-05-26
River job stuck at running


didntpay:
We recently noticed that a lot of our river job(around 5000) is stuck at a state of running. We have a global timeout of 1 hour and thought that the rescue is suppose to resolve this? These are across different job queues. 

River job stats
```
SELECT
  COUNT(*) AS stuck_count,
  COUNT(DISTINCT kind) AS distinct_kinds,
  COUNT(DISTINCT queue) AS distinct_queues,
  MIN(created_at) AS oldest_created_at,
  MAX(created_at) AS newest_created_at,
  NOW() - MIN(created_at) AS oldest_age,
  NOW() - MAX(created_at) AS newest_age,
  AVG(NOW() - created_at) AS avg_age,
  PERCENTILE_CONT(0.5) WITHIN GROUP (ORDER BY EXTRACT(EPOCH FROM NOW() - created_at)) * INTERVAL '1 second' AS median_age,
  MIN(attempt) AS min_attempts,
  MAX(attempt) AS max_attempts,
  AVG(attempt)::numeric(10,2) AS avg_attempts,
  COUNT(*) FILTER (WHERE attempted_at < NOW() - INTERVAL '1 hour') AS not_touched_in_1h,
  COUNT(*) FILTER (WHERE attempted_at < NOW() - INTERVAL '6 hours') AS not_touched_in_6h
FROM river_jobs.river_job
WHERE state = 'running'
  AND created_at < NOW() - INTERVAL '1 day';

[
	{
		"stuck_count": 5133,
		"distinct_kinds": 42,
		"distinct_queues": 7,
		"oldest_created_at": "2026-02-24 22:08:09.618834+00",
		"newest_created_at": "2026-05-25 19:10:35.474235+00",
		"oldest_age": "91 days 00:29:38.338464",
		"newest_age": "1 day 03:27:12.483063",
		"avg_age": "40 days 29:02:53.066961",
		"median_age": "1155:35:03.048939",
		"min_attempts": 1,
		"max_attempts": 4,
		"avg_attempts": 1.08,
		"not_touched_in_1h": 5133,
		"not_touched_in_6h": 5133
	}
]


```

River version
```
github.com/riverqueue/river v0.31.0
github.com/riverqueue/river/rivertype v0.31.0
```

Let me know if any other information is helpful

brandur:
[...]

PR #1264, opened 2026-06-01
Clarify worker cancellation requirements for stuck jobs

peter941221:
This follows up the stuck-job discussion in #1258.

The current docs already mention that workers should respect context cancellation, but they don't quite connect that requirement to the failure mode users are likely to see in practice: jobs that remain in `running` because worker code is blocked without also observing `ctx.Done()`.

1. This change expands `Worker.Work`'s contract to call out blocking operations like channels, timers, and network work explicitly, and recommends a `select` that also watches `ctx.Done()`.
2. It adds the same guidance to the top-level package docs and README so that users encounter it earlier, not only in the interface comment.
3. It adds a short diagnostic hint that points users at `num_jobs_stuck` in the producer job counts log line when investigating jobs that appear stuck in `running`.
4. It adds a small note to the graceful shutdown example to make the cancellation expectation concrete in a worker example that's already about stop behavior.

The goal here isn't to change rescue semantics, only to make the existing behavior easier to understand and diagnose.

Verification:
1. Formatted the touched Go files with `gofmt`.
2. Ran `go test ./... -run TestDoesNotExist -count=1` to confirm the repo still loads and compiles after the docs/example changes.

Caveat:
1. I didn't run the database-backed example tests in this environment because local PostgreSQL authentication isn't configured.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -41,11 +41,6 @@ Workers expose a `Work` function that dictates how jobs run.
 	    return nil
 	}
 
-Workers should respect context cancellation. In particular, if work may block
-on a channel, timer, or network operation, prefer a `select` that also watches
-`ctx.Done()`. A job that ignores cancellation may continue running even after
-River has timed it out or the job rescuer has moved it out of `running`.
-
 # Registering workers
 
 Jobs are uniquely identified by their "kind" string. Workers are registered on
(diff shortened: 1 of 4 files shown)


brandur (commented):
@peter941221 I think some of this is a little too much detail for some of the areas like README and general Godoc intro — these sections are arguably already a bit too long and we should be really prescriptive about what is allowed to go in there.

I think the piece in `worker.go` is a nice improvement — do we want to revert the other stuff and we can keep that?

In terms of general documentation — I think what I'll do is put together a specific "stuck jobs" advanced topics page. We have a little already at https://riverqueue.com/docs/graceful-shutdown#stuck-programs, but this could stand to be expanded a bit.

peter941221:
@brandur sure, will do that.

peter941221:
Reverted the broader README, Godoc, and example additions and kept the worker.go clarification only. We can move the deeper stuck-jobs guidance to a dedicated advanced topics page later.

brandur (approved):
Great! Thx.

PR #1268, opened 2026-06-01
Standardize "hot" operation timeout

brandur:
This one's a simpler alternative to #1263. All we do is take the
existing timeout added in #1255 and standardize it with the timeout we
use for job completion, putting a comment on the two saying we should
revisit.

I decided to leave it as 10 seconds after all (instead of changing it to
30 seconds like I had in #1263) after seeing that job completion already
uses 10 seconds. This may still not be long enough for a jobs table in
serious trouble due to dead tuples, but if it's not, job completion is
already going to be in trouble so it's likely things have degenerated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


bgentry (approved):

brandur:
ty!

PR #1269, opened 2026-06-02
Add helper for job metadata updates

### LINKED ISSUE #1218: Official way to set keys in metadata
Currently there seems to be no official way to set keys in metadata at the end of the job run. I want to add additional metadata in `WorkEnd` but it looks this is not possible. The only official way is `river.RecordOutput` which sets the output key only.

I can use `driver.GetExecutor().JobUpdate` with `riverdriver.JobUpdateParams` but API says "API is not stable. DO NOT USE." Also, downside of using this is that it makes another database request instead of just adding to metadata in ctx which would be merged at the end when executor updates the job row.

There is already internal `jobexecutor.ContextKeyMetadataUpdates` with tests that metadata updates using it works. I suggest you expose simple API to set keys on it, like `river.SetMetadata(ctx, key, value)`. Such helper could also check that key does not have `river:` prefix to never conflict with river keys.

My use case is that we use in our codebase rich errors using [this package](https://gitlab.com/tozd/go/errors). It records stack trace and additional structured details. We want that in the case of an error to extract that additional information from the error and record it. It looks like metadata is the right place to store it (river by default just stores the message as returned from `.Error()` and not other error payload). And we can use `WorkEnd` to do that.

peter941221:
1. Adds a small public `SetMetadata(ctx, key, value)` helper for writing metadata updates onto the existing work-context merge path.
2. Keeps the surface intentionally narrow: no batch API, no delete API, no hook signature changes, and no completion semantics changes.
3. Rejects `river:`-prefixed keys so user code can't collide with River-reserved metadata keys.
4. Adds focused coverage for the helper contract itself plus worker and `HookWorkEnd` integration paths.

Closes #1218.

Verification:
1. `go test ./... -run "TestSetMetadata|Test_Client_Common/(WithWorkerSettingMetadata|WithWorkEndHookSettingMetadata)" -count=1`
2. `make test`


cmd/river/rivercli/river_cli_test.go
```diff
@@ -290,6 +286,38 @@ func TestBaseCommandSetNonParallel(t *testing.T) {
...
+		require.NoError(t, cmd.Execute())
+
+		cmd = setup(t)
+		cmd.SetArgs([]string{"migrate-down", "--database-url", testDatabaseURL, "--max-steps", "100", "--schema", schema})
+		require.NoError(t, cmd.Execute())
+
+		cmd = setup(t)
+		cmd.SetArgs([]string{"validate", "--database-url", testDatabaseURL, "--schema", schema})
+		require.Error(t, cmd.Execute())
+	})
```
brandur: This change seems unrelated I think right?

worker.go
brandur, line 75: This change was merged elsewhere already right?

client_test.go
```diff
@@ -1293,6 +1297,33 @@ func Test_Client_Common(t *testing.T) {
...
 	})
 
+	t.Run("WithWorkerSettingMetadata", func(t *testing.T) {
+		t.Parallel()
+
+		_, bundle := setup(t)
+
+		AddWorkerArgs(bundle.config.Workers, metadataWorkerArgs{}, WorkFunc(func(ctx context.Context, job *Job[metadataWorkerArgs]) error {
+			return SetMetadata(ctx, "worker_key", "worker_value")
+		}))
```
brandur, line 1307: Hey Peter, do you want to take a look at this pattern used elsewhere in this file?

``` go
		type JobArgs struct {
			testutil.JobArgsReflectKind[JobArgs]
		}

		AddWorker(client.config.Workers, WorkFunc(func(ctx context.Context, job *Job[JobArgs]) error {
			workedChan <- struct{}{}
			return nil
		}))
```

This just prevents needing to add a new `metadataWorkerArgs` to the global namespace. Instead it can stay scoped into the specific test case.

metadata.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"context"
+	"errors"
+	"strings"
+
+	"github.com/riverqueue/river/internal/jobexecutor"
+)
+
+var errMetadataNotSettable = errors.New("SetMetadata must be called within a Worker, worker middleware, or work hook")
```
brandur: It kind of seems like we're a bit inconsistent with casing in this error message. i.e. "Worker" versus "work middleware".

Maybe just use the downcased version everywhere?

brandur (commented):
Cool! Seems useful. I like the touch that we reject use of the `river:*` namespace.

peter941221:
Thanks. I dropped the unrelated CLI and worker.go changes, scoped the test-only job args into the test case, and normalized the error message casing.

peter941221:
One note on the remaining red check: the retriggered run moved the failure from build_and_test (1.26, 15) to build_and_test (1.26, 14), and both failures happened during Docker pull for postgres:15 in container setup rather than during test execution, so this looks like CI flakiness rather than a code regression.\n\nLatest run: https://github.com/riverqueue/river/actions/runs/26804256615

client_test.go
```diff
@@ -1394,6 +1425,81 @@ func Test_Client_Common(t *testing.T) {
 		require.True(t, hookWorkEndCalled.Load())
 	})
 
+	t.Run("WithWorkEndHookSettingMetadata", func(t *testing.T) {
```
brandur, line 1428: This works for `HookWorkBegin` too doesn't it? I wonder if we should add an explicit test about that just for completeness.

metadata.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+var errMetadataNotSettable = errors.New("SetMetadata must be called within a worker, worker middleware, or work hook")
+
+// SetMetadata records a metadata value to be merged into the job's metadata
+// when the current work attempt finishes.
+//
+// This function is only valid from a worker, worker middleware, or work hook
+// like rivertype.HookWorkEnd. Keys prefixed with `river:` are reserved for
+// internal use and may not be set by user code.
```
brandur: A couple things here:

* Since we call out `rivertype.HookWorkEnd`, can we also mention `rivertype.HookWorkBegin`? There's no other way to know that metadata setting would work from there, so may as well call it out in docs.
* Can you mention a little detail on the implementation? e.g. Metadata resides in context until a job is being completed, at which point it's added to the job row. Metadata will be stored regardless of whether the job executed successfully or errored.
* Mention that metadata values must be JSON marshable because they'll end up in a `jsonb` column.
* Pre-existing metadata keys will be completely replaced by a new value at the same key set by `SetMetadata`.

client_test.go
peter941221: Added a HookWorkBegin coverage case in client_test.go in f81b143.

metadata.go
peter941221: Updated the SetMetadata comment to mention HookWorkBegin, work-context staging, persistence on success and error, JSON-marshalability, and same-key replacement.

brandur:
@peter941221 Was just thinking about this a bit more today, and could you rename `GetMetadata` to `MetadataGet`?

It's a bit more of a mouthful, but since this is a top-level package function, using a `Metadata*` prefix means it'll stay sorted with alike functions if we end up adding more functions later (e.g. `MetadataClear`, `MetadataDelete`, or something like that).

It's also consistent with how things are named on `Client` like `JobGet` vs. `JobList`.

peter941221:
Renamed the helper to MetadataSet in d8e114f so the top-level metadata helpers stay grouped under Metadata*.

brandur:
@peter941221 Great, thanks! Okay, this is looking good. Can you add a CHANGELOG entry? Put it under the "added' section.

peter941221:
Added the CHANGELOG entry under Unreleased -> Added in c0e0902.

peter941221:
The red check was another windows-latest postgres bootstrap failure in action-setup-postgres before checkout or CLI execution. I pushed a no-op commit to retrigger CI on a fresh runner.

brandur:
Thanks! Yeah, we definitely have a couple GItHub Actions-related flakes going on.

brandur (approved):

PR #1271, opened 2026-06-03
Log dropped subscription events

peter941221:
1. Failure mechanism

Subscription delivery already uses non-blocking sends and the public docs already warn that slow consumers can lose events. Right now, when a subscriber channel fills up, `subscription_manager` drops the event silently. That leaves you with no signal that a subscription fell behind or which kind of event was lost.

2. Semantic change

This adds a warn log when a job or queue subscription event is dropped because the subscriber buffer is full.

The log includes `event_kind` so you can tell what was lost.

3. Why this design

I kept the existing delivery semantics. This does not add retries, change buffering, or block producers behind a slow subscriber. It only turns a known lossy path into something observable.

4. Testing

1. Added `Test_SubscriptionManager/LogsDroppedQueueEvents` in `subscription_manager_test.go`.
2. Ran `go test . -run Test_SubscriptionManager/LogsDroppedQueueEvents -count=1`.
3. Ran `go test . -run TestDoesNotExist -count=1` as a compile/load smoke check.


brandur (commented):
Yeah, this isn't a bad idea. Want to add a changelog?

peter941221:
Added a changelog entry under Unreleased -> Changed.

brandur (approved):

PR #1272, opened 2026-06-03
Align SubscribeConfig chan size contract

peter941221:
1. Failure mechanism

`SubscribeConfig.ChanSize` already treats `0` as "use the default buffer size", but the surrounding contract drifted away from that behavior. The public comment did not say that zero selects the default, it still had a wording typo, and the panic string still claimed the value had to be greater or equal to `1`.

2. Semantic change

This aligns the public comment, panic string, and tests with the existing zero-as-default behavior.

It also fixes the comment typo from `overall` to `overflow`.

3. Why this design

The runtime behavior was already right. This patch only makes the boundary conditions easier to understand and harder to misread in tests or review.

4. Testing

1. Added `Test_SubscriptionManager/PanicOnNegativeChanSize`.
2. Added `Test_SubscriptionManager/UsesDefaultChanSizeWhenZero`.
3. Ran `go test . -run "Test_SubscriptionManager/(PanicOnNegativeChanSize|UsesDefaultChanSizeWhenZero)" -count=1`.
4. Ran `go test . -run TestDoesNotExist -count=1` as a compile/load smoke check.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1328,7 +1328,7 @@ type SubscribeConfig struct {
 	// subscription. Incoming events that would overflow this buffer because a
 	// listener isn't reading from the channel in a timely manner will be dropped.
 	//
-	// Defaults to 1000 when set to 0.
+	// Defaults to 1000.
 	ChanSize int
 
 	// Kinds are the kinds of events that the subscription will receive.
--- a/client.go
+++ b/client.go
@@ -1325,10 +1325,10 @@ const subscribeChanSizeDefault = 1_000
 // Client.SubscribeConfig.
 type SubscribeConfig struct {
 	// ChanSize is the size of the buffered channel that will be created for the
-	// subscription. Incoming events that overall this number because a listener
-	// isn't reading from the channel in a timely manner will be dropped.
+	// subscription. Incoming events that would overflow this buffer because a
+	// listener isn't reading from the channel in a timely manner will be dropped.
 	//
-	// Defaults to 1000.
+	// Defaults to 1000 when set to 0.
 	ChanSize int
 
 	// Kinds are the kinds of events that the subscription will receive.
--- a/client_test.go
+++ b/client_test.go
@@ -7291,7 +7291,7 @@ func Test_Client_SubscribeConfig(t *testing.T) {
 
 		client := newTestClient(t, dbPool, config)
 
-		require.PanicsWithValue(t, "SubscribeConfig.ChanSize must be greater or equal to 1", func() {
+		require.PanicsWithValue(t, "SubscribeConfig.ChanSize must be greater or equal to 0", func() {
 			_, _ = client.SubscribeConfig(&SubscribeConfig{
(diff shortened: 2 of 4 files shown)


client.go
```diff
@@ -1325,10 +1325,10 @@ const subscribeChanSizeDefault = 1_000
 // Client.SubscribeConfig.
 type SubscribeConfig struct {
 	// ChanSize is the size of the buffered channel that will be created for the
-	// subscription. Incoming events that overall this number because a listener
-	// isn't reading from the channel in a timely manner will be dropped.
+	// subscription. Incoming events that would overflow this buffer because a
+	// listener isn't reading from the channel in a timely manner will be dropped.
 	//
-	// Defaults to 1000.
+	// Defaults to 1000 when set to 0.
```
brandur: Maybe revert this line. I think it's implied well enough that the empty value of 0 in Go would result in the default. We have this convention (i.e. a single line at the end of the docblock) in a bunch of client config places, and if we were to change this elsewhere you'd have comments like this:

``` go
	// Defaults to empty, which causes the client to look for tables using the
	// setting of Postgres `search_path`.
	Schema string
```

Changing to this:

``` go
	// Defaults to empty when left as empty string, which causes the client to look for tables using the
	// setting of Postgres `search_path`.
	Schema string
```

I just don't think it's necessary.

subscription_manager.go
```diff
@@ -211,7 +211,7 @@ func (sm *subscriptionManager) distributeQueueEvent(event *Event) {
 // an overridden size.
 func (sm *subscriptionManager) SubscribeConfig(config *SubscribeConfig) (<-chan *Event, func()) {
 	if config.ChanSize < 0 {
-		panic("SubscribeConfig.ChanSize must be greater or equal to 1")
+		panic("SubscribeConfig.ChanSize must be greater or equal to 0")
```
brandur: I'd revert this as well. I know it's technically true that it can be 0, but I feel like if you're setting the value, you should be told that you should set it to at least 1. An explicit `ChanSize: 0` should just be removed instead of left that way.

peter941221:
Reverted both wording changes and kept the zero-default behavior test coverage.

client_test.go
```diff
@@ -7291,7 +7291,7 @@ func Test_Client_SubscribeConfig(t *testing.T) {
 
 		client := newTestClient(t, dbPool, config)
 
-		require.PanicsWithValue(t, "SubscribeConfig.ChanSize must be greater or equal to 1", func() {
+		require.PanicsWithValue(t, "SubscribeConfig.ChanSize must be greater or equal to 0", func() {
```
brandur: Probably need to revert this line as well.

peter941221: Reverted the remaining test assertion as well.

brandur:
Ah doh, looks like this conflicted with your other change. Want to rebase?

brandur (approved):

PR #1274, opened 2026-06-04
Add `Pilot.JobRescueMany` for active job rescue

brandur:
This one's presented as an alternative to #1256 with its own alternative
Pro implementation and which may provide some superior properties around
potential bloat for table bloat in `river_job`.

Changes:

* Pilot picks up a new `JobRescueMany` function. In the `StandardPilot`,
  this just goes directly to the function of the same name in the
  driver's `Executor`.

* We now inject `ProducerReportInterval` in `PilotInitParams`. This
  value is useful to an underlying pilot in knowing what age of producer
  row it should respect when performing a job rescue.


bgentry (approved):
Looks good pending the other related PR :pray:

brandur:
Awesome, thanks. Will address the feedback on the other branch over there.

PR #1281, opened 2026-06-10
Validation to detect duplicate step names in resumable jobs

brandur:
Follows up #1226 to tighten things up a little with a validation. If a
user accidentally puts in duplicate step names, detect this problem and
return an error. A duplicate step name represents a step that can never
be resumed from, so it makes sense to find out about this sooner rather
than later.


bgentry:
@codex review

bgentry (approved):
LGTM, up to you if you want to wait for Codex review :v:

brandur:
Thx!

PR #1282, opened 2026-06-11
Changelog fixes for v0.40.0

brandur:
With #1115, #1224, and #1275, just a few fixups for the changelog to
make sure we get in a correction (migration no longer applies to just
SQLite) and a few more details about the migration that'll be run. Also
add a missing entry for #1275.


bgentry (approved):

brandur:
thx.

PR #1289, opened 2026-06-19
Add `Config.StopAbandonTimeout` to abandon jobs still running after stop timeouts

brandur:
Here, add a new `Config.StopAbandonTimeout` on top of the existing
`SoftStopTimeout` whose job it is to recover badly behaving job as much
as possible before coming to a full stop. Currently, if a client is
stopping and is running jobs that don't respond to context cancellation,
those jobs end up getting left in a `running` state, which means that
they won't be recoverable again until they're rescued an hour later.

`StopAbandonTimeout` engages after soft stop, and has each producer
initiate abandonment of its running jobs, which means setting them to an
error state (or discarded in case a job is out of retries). Because
they're errored, they'll get to run immediately the next time a client
starts up.

Ideally, users don't need to depend on this functionality since the
"correct" behavior would be to make sure that all jobs are able to
respond to context cancellation, so we make this new feature optional.

### DIFF
--- a/client.go
+++ b/client.go
@@ -378,11 +378,9 @@ type Config struct {
 	Schema string
 
 	// SoftStopTimeout is the maximum amount of time that the client will wait
-	// for running jobs to finish during a stop before their contexts are
-	// cancelled. After the timeout elapses, the client escalates to a hard stop
-	// by cancelling the context of all running jobs. This applies regardless of
-	// how stop is initiated — whether by calling Stop, StopAndCancel, or by
-	// cancelling the context passed to Start.
+	// for running jobs to finish during a graceful stop before entering soft
+	// stop by cancelling job contexts. This applies when stop is initiated by
+	// calling Stop or by cancelling the context passed to Start.
 	//
 	// In combination with signal.NotifyContext on the context passed to Start,
 	// this can simplify graceful stop to:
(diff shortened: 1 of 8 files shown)


producer.go
```diff
@@ -728,11 +758,89 @@ func (p *producer) addActiveJob(id int64, executor *jobexecutor.JobExecutor) {
...
+			Attempt: job.Attempt,
+			Error:   producerHardStopError,
+		})
+		if err != nil {
+			panic(fmt.Errorf("error serializing hard stop error: %w", err))
+		}
+
+		var setStateParams *riverdriver.JobSetStateIfRunningParams
+		if job.Attempt >= job.MaxAttempts {
+			setStateParams = riverdriver.JobSetStateDiscarded(job.ID, now, errData, nil)
```
brandur, line 880: This mirrors existing behavior where a soft stop will set an error and potentially send the job to discarded, but looking at this again, this existing behavior does seem potentially wrong.

client.go
```diff
@@ -221,6 +221,18 @@ type Config struct {
...
+	// after job contexts are cancelled during shutdown before forcing jobs still
+	// running to an errored state. This hard stop phase lets jobs be retried
+	// immediately on the next client start instead of waiting for rescue.
+	//
+	// The timer starts only after a soft stop has begun by cancelling job
+	// contexts, like after SoftStopTimeout elapses, StopAndCancel is called, or
+	// the Start context is cancelled without SoftStopTimeout configured.
+	//
+	// Defaults to no timeout (hard stop disabled).
+	HardStopTimeout time.Duration
```
bgentry: Missed the alphabetical sort order here

bgentry (approved):
Looks good aside form ordering nit and missing changelog :shipit:

client.go
brandur: Oops, quite right. I renamed this a couple times which is probably what happened.

bgentry:
@brandur did you want to get this one into the release?

brandur:
I would, but I guess it's prone to the same considerations you brought up in #1290. Probably easier to do a fast follow up with a couple stopping improvements after 0.40.0 is out.

PR #1291, opened 2026-06-20
Add `JobStuckHandler` callback

brandur:
Here, add a `JobStuckHandler` callback that's invoked when a producer
consider a job to be "stuck". i.e. Passed its timeout, cancellation
attempted, but job didn't respond to cancellation.

The callback includes some basic information about the job that became
stuck, along with the total number of stuck jobs. A result will
optionally open a new executor slot to replace the one taken up by the
stuck job so that a producer that continues to be run doesn't get
completed starved by stuck jobs. The idea here is that clients can
configure themselves to open new slots up to a certain point, but then
may want to restart themselves if there's enough jobs stuck that they
could become a memory liability.

### DIFF
--- a/stuck_job.go
+++ b/stuck_job.go
@@ -1,0 +1,32 @@
+package river
+
+import "context"
+
+// JobStuckHandler is invoked when a producer detects that a job exceeded its
+// timeout and did not return within the configured stuck-job timeout margin.
+type JobStuckHandler func(ctx context.Context, params JobStuckHandlerParams) JobStuckHandlerResult
+
+// JobStuckHandlerParams are parameters passed to JobStuckHandler.
+type JobStuckHandlerParams struct {
+	// ID is the ID of the stuck job.
+	ID int64
+
+	// Kind is the kind of the stuck job.
+	Kind string
+
+	// Queue is the queue where the stuck job is running.
+	Queue string
+
+	// TotalStuckJobs is the total number of jobs currently considered stuck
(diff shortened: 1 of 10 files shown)


brandur:
@codex review

stuck_job.go
```diff
@@ -0,0 +1,32 @@
...
+	TotalStuckJobs int
+}
+
+// JobStuckHandlerResult is the result returned by JobStuckHandler.
+type JobStuckHandlerResult struct {
+	// OpenWorkerSlot instructs River to treat the stuck job as no longer
+	// occupying a worker slot so another job can begin executing. This can be
+	// dangerous because the stuck job's goroutine is still running, so the queue
+	// may temporarily have more active job goroutines than MaxWorkers.
+	OpenWorkerSlot bool
```
bgentry: I wonder if `ReleaseWorkerSlot` might be slightly clearer?

bgentry (approved):
Needs a changelog entry and one suggestion on naming, otherwise LGTM!

stuck_job.go
brandur: Hmm, I think "open" is a little more correct here — "release" seems to imply that an existing slot is being released, which is not what's happening (unfortunately, we can't release the existing slot). "Open" is better IMO because it says you're "opening a new slot". That said, maybe `AddWorkerSlot` is better than either as it's even clearer. Changed to that.

PR #1297, opened 2026-07-02
reindex numbered concurrent artifacts

### LINKED ISSUE #1296: index hell: 210 GB of space for indexes that the logs say are never used...
we had a lot of used space in our database and rather than increase the size we are paying for we decided to use AI to look for ways to trim things that are not needed and it found a number of indexes that are never used ... many of them from river.  it reported that these indexes existed and were using a lot of space but were never used.  My boss actually found this and just deleted the indexes so I can't look at them anymore but thought I should report it here in case there is some optimization that could be made in river.  right now I'm being scolded for using river ...

- river_job_args_index (plus 30+ duplicate indexes _ccnew1 through _ccnew31): Size = 6.8 GB each (Totaling ~210 GB!), scans = 0

We re-deployed after they were deleted as I assumed the indexes would only come from migrations to confirm that they won't come back and they didn't however when I rgrep the code for 'ccnew' it doesn't look like these come from migrations and so I'm concerned.  If these indexes come back I'm going to have a problem.

210 GB of space on indexes that the logs indicate are never used...

bgentry:
Postgres can leave invalid `_ccnew` and `_ccold` indexes after a failed concurrent reindex. When artifacts already existed and Postgres had to append digits to keep names unique, River only checked the unnumbered forms and could keep trying the same reindex, adding more large artifacts.

Add a driver-level catalog lookup for artifacts belonging to a configured index. It matches the exact configured index name followed by `_ccnew` or `_ccold` with optional digits, returns the names for logging and cleanup, and lets stop-time cleanup drop the same numbered artifacts. SQLite keeps no-op semantics because it does not create concurrent reindex artifacts.

Extend the maintenance and driver conformance coverage for numbered forms and near-miss names so similar user-created indexes are not treated as PostgreSQL artifacts. Document the user-facing fix in the changelog.

Fixes #1296.


brandur (approved):
Cool.

PR #1305, opened 2026-07-11
preserve reserved queue metadata

bgentry:
Adds support for reserved internal queue metadata. Queue updates preserve reserved keys, while public queue reads and notifications omit them.

Queue cleanup now locks candidates before deletion and retains rows with unexpired internal state. The driver interface also exposes shared transaction advisory locks, and pilot initialization receives River's existing producer stale-retention period.

### DIFF
--- a/internal/rivercommon/river_common.go
+++ b/internal/rivercommon/river_common.go
@@ -1,8 +1,10 @@
 package rivercommon
 
 import (
+	"encoding/json"
 	"errors"
 	"regexp"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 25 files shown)


riverdriver/riverpgxv5/internal/dbsqlc/river_queue.sql
```diff
@@ -26,12 +26,21 @@ RETURNING *;
...
     FROM /* TEMPLATE: schema */river_queue
     WHERE river_queue.updated_at < @updated_at_horizon
+        AND CASE
+            WHEN NOT (river_queue.metadata ? 'river:rate_limit_rollup') THEN true
+            WHEN jsonb_typeof(river_queue.metadata -> 'river:rate_limit_rollup' -> 'expires_at_unix') = 'number'
+                AND river_queue.metadata -> 'river:rate_limit_rollup' ->> 'expires_at_unix' ~ '^[0-9]+$'
+                THEN (river_queue.metadata -> 'river:rate_limit_rollup' ->> 'expires_at_unix')::numeric
+                    <= extract(epoch FROM coalesce(sqlc.narg('now')::timestamptz, now()))
+            ELSE false
+        END
```
brandur, line 40: Just wanted to make sure you saw this part — is it intentional that we're calling out specific pro metadata keys in these queries? Seems a bit leaky.

internal/rivercommon/river_common.go
```diff
@@ -12,12 +14,48 @@ import (
...
+	AllQueuesString              = "*"
+	MaxAttemptsDefault           = 25
+	PriorityDefault              = 1
+	QueueDefault                 = "default"
+	QueueMetadataKeyUserMetadata = "river:user_metadata"
 )
 
+// QueueMetadataForUserWrite removes reserved keys from user-supplied queue
+// metadata while preserving every nonreserved value.
+func QueueMetadataForUserWrite(metadata []byte) []byte {
```
brandur, line 26: Can you clarify the "user metadata" thing for me a bit?

For metadata on jobs, aren't we kind of saying that the entire space is user metadata minus the River specific stuff that we prefix with `river:*`? I wonder if we should try to do the same thing here for queues.

PR #1311, opened 2026-07-16
Implement Turso support (basically SQLite)

brandur:
Here, start testing against Turso, which is a SQLite-compatible database
rewritten in Rust and with some nice additional features. There are a
couple minor tweaks we need to make to SQLite queries to make this work,
but nothing major, so the idea here is that like libSQL, we can get
Turso support without any major maintenance overhead and hopefully make
it a bit of a PR win.

I tried to do this originally a few months ago, but Turso had a few bugs
around JSON handling that we couldn't work our way around. These were
fixed recently, so now it's possible to do.


brandur:
@bgentry Thoughts on this one? Managed to get the diff down to be pretty small, so it shouldn't have too much future effect.

bgentry (approved):

brandur:
Thanks!

PR #1316, opened 2026-07-20
Remove automatic cross-compatibility for hooks/middleware

brandur:
This one's related to #1284. In that pull request, we made it so that
any hooks sent to `Config.Hooks` that implemented middleware or any
middleware sent to `Config.Middleware` that implement hooks were
automatically extracted and use for all types.

I've been thinking about this for the last couple of days, and in
retrospect, I think this automatic compatibility might be a little
unnecessarily aggressive. The case in favor is that when adding
additional hook-related functionality to something like `otelriver`, the
user can just leave in their existing instance in middleware and not
change anything to get the existing functionality. (Without the automatic
compatibility they'd have to move it to `Config.Plugins` instead.) The
case against is that it may result in some surprising behavior if you
only intended to add something as either a hook or middleware.

We haven't shipped a release containing #1284 yet, so we can still
change this if we want to. I didn't ship a strict revert because #1284
also brings in `Config.Plugins` and some useful internal changes to
support everything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Both hooks and middleware should now be registered under a general `Config.Plugins` instead, a change that simpifies things somewhat, and better handles cases where a hook or middleware might be _both_ a hook and a middleware as opposed to only one or the other. `Config.Hooks` and `Config.Middleware` are still available for use, but considered deprecated in favor of the more general `Config.Plugins`. [PR #1284](https://github.com/riverqueue/river/pull/1284).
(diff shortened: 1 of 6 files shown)


bgentry (approved):

brandur:
Thanks!

Okay, I couldn't fully decide whether to merge this one or not (can make arguments both directions), so I sent it through Codex and asked it to be as objective as possible:

>   ## Final recommendation
>
>  Merge the branch.
>
>  The automatic approach is defensible only if the firm long-term direction is that Hooks and Middleware are deprecated aliases for Plugins, and seamless activation of new
>  integration features is considered more important than explicit configuration.
>
>  The current branch instead keeps all three fields supported, as documented in client.go:242 and CHANGELOG.md:16. Under that strategy, role-specific semantics are more coherent,
>  safer, more expressive, and more compatible with previously released behavior.
>
>  I would coordinate the merge with an otelriver migration example showing existing users that hybrid functionality now belongs in Config.Plugins.

That works for me, so going to pull this in.

PR #1317, opened 2026-07-21
Promote `JobGetStuck` into the pilot

brandur:
While fixing up some bugs in the active rescue feature, I found that we
need to have slightly divergent variations of `JobGetStuck` for when the
rescuer is using the feature versus not.

It's possible to implement the variation entirely on the pro executor by
overriding `JobGetStuck` there, but doing so makes the code more
roundabout because you need to fetch an extra parameter off the pro
pilot. Things become more direct by instead making `JobGetStuck` a pilot
function and having two separate implementations, and this is also more
conventional elsewhere too.

### DIFF
--- a/internal/maintenance/job_rescuer.go
+++ b/internal/maintenance/job_rescuer.go
@@ -194,7 +194,8 @@ func (s *JobRescuer) runOnce(ctx context.Context) (*rescuerRunOnceResult, error)
 	res := &rescuerRunOnceResult{}
 
 	for {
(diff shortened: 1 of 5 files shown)


brandur:
Also added one extra commit here for passing a stuck horizon down into the function. This was needed to fix another bug Codex identified in the active job rescue branch. It's a pretty small change though.

bgentry (approved):

brandur:
Thanks!

PR #1318, opened 2026-07-21
Make job rescue paginated to protect against full batch of no-rescue jobs

brandur:
This one's aimed at fixing a long-standing bug accidentally detected
while working on another rescue-related feature.

The `JobRescuer` works by fetching a full batch of jobs to rescue then
going through each one to determine what it should be doing about it.
The default batch size is 10k, so this generally works perfectly fine.

Some stuck jobs are potentially not rescued if their timeout is
configured to be -1 (no timeout).

Codex identified a tail bug possibility in which if you had an entire
batch worth of jobs with -1 timeouts, they'd block any jobs after them
from being rescued. So the rescuer would rescue 10k jobs, determine none
of them needed rescue, then go back to sleep, stranding any jobs after
that.

This is such a tail possibility that it's probably fine, but just since
we're making changes to the rescue driver boundary right now anyway,
it's not a bad time to just get this one fixed up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Guard against empty job slice returned by `JobSetStateIfRunningMany` when a job has been deleted mid-run. [PR #1308](https://github.com/riverqueue/river/pull/1308).
+- Fixed `JobRescuer` pagination so a full batch of running jobs with disabled or longer worker-specific timeouts can't prevent later stuck jobs from being rescued. [PR #1318](https://github.com/riverqueue/river/pull/1318).
 
 ## [0.40.0] - 2026-07-02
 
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Guard against empty job slice returned by `JobSetStateIfRunningMany` when a job has been deleted mid-run. [PR #1308](https://github.com/riverqueue/river/pull/1308).
+- Fixed `JobRescuer` pagination so a full batch of running jobs with disabled or longer worker-specific timeouts can't prevent later stuck jobs from being rescued.
```
bgentry: PR number?

bgentry (approved):

CHANGELOG.md
brandur: Thx, fixed.

PR #1319, opened 2026-07-21
Update producer keep alive interval to 30 seconds

brandur:
Halve the producer keep alive interval from 1 minute to 30 seconds. This
is still a long enough duration that it shouldn't significantly impact
database health in any way, but since producers are reporting a little
more often, makes it a bit less likely that we accidentally rescue a
running job with the active job rescue feature.


bgentry (approved):
FYI the PR description & commit have a typo: `s/Halve he/Halve the`

Otherwise LGTM, I was looking at doing something similar for another feature anyway!

brandur:
Thx! Typo fixed.

PR #1324, opened 2026-07-22
Harden JSON unmarshaling in job rescuer + job executor

### LINKED ISSUE #1323: job_rescuer: UnmarshalJob error doesn't short-circuit makeRetryDecision, causing incorrect rescue behavior
## Bug

When `workUnit.UnmarshalJob()` returns an error inside `makeRetryDecision`, the error is only logged — execution continues to call `workUnit.Timeout()` and `workUnit.NextRetry()` on an incompletely initialized work unit.

## Affected file

`internal/maintenance/job_rescuer.go`, lines 322–326:

```go
workUnit := workUnitFactory.MakeUnit(job)
if err := workUnit.UnmarshalJob(); err != nil {
    s.Logger.ErrorContext(ctx, s.Name+": Error unmarshaling job args: %s"+err.Error(),
        slog.String("job_kind", job.Kind), slog.Int64("job_id", job.ID))
}
// execution falls through to here even on unmarshal failure:
timeout := workUnit.Timeout()
```

## Impact

- `workUnit.Timeout()` returns its zero value (`0`), which passes the `timeout < 0 || timeout > 0 && ...` guard as neither branch fires — the `jobRetryDecisionIgnore` path is skipped, so the job is never left running when it should be.
- `workUnit.NextRetry()` also returns zero time, falling back to `ClientRetryPolicy`, silently losing any worker-specific retry override.

The job gets an incorrect rescue decision (reschedule with wrong timing) instead of the discard-with-error path that `nil workUnitFactory` already uses correctly.

There's also a minor string formatting issue on line 323: `": Error unmarshaling job args: %s" + err.Error()` embeds a literal `%s` in the log string instead of using a slog attribute.

## Suggested fix

Return early on unmarshal failure, matching the pattern used for the nil-factory case above:

```go
if err := workUnit.UnmarshalJob(); err != nil {
    s.Logger.ErrorContext(ctx, s.Name+": Error unmarshaling job args",
        slog.String("job_kind", job.Kind), slog.Int64("job_id", job.ID),
        slog.String("error", err.Error()))
    return jobRetryDecisionDiscard, time.Time{}
}
```

(Whether to discard or retry with a delay is a policy call — discard mirrors the existing nil-factory behavior, but a retry with backoff could also be appropriate.)

brandur:
Here, do a little more to harden unmarshaling in the job rescuer and job
executor. Previously, if a job's JSON didn't unmarshal successfully,
we'd log an error and move on, which could result in unexpected trouble.

Realistically, this would be a very rare circumstance though luckily --
the Postgres/SQLite `jsonb` fields guarantee the formatting of the JSON
data, so the only way for an unmarshaling problem to occur is if a job's
data was stored initially, and then the JSON Go struct later change to
be incompatible (i.e. string changed to an int or something of that
nature).

Here, during an unmarshal error, use the job's standard retry schedule
and back off, discarding the job if it's at the end of its allowed
retries. This approach is best because it gives the user a chance to
notice the failure and correct a potential unmarshaling problem by
either manipulating the job row's data or fixing their Go struct.

Fixes #1323, but also addresses a similar problem in the job executor
that Codex found while working.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Guard against empty job slice returned by `JobSetStateIfRunningMany` when a job has been deleted mid-run. [PR #1308](https://github.com/riverqueue/river/pull/1308).
 - Fixed `JobRescuer` pagination so a full batch of running jobs with disabled or longer worker-specific timeouts can't prevent later stuck jobs from being rescued. [PR #1318](https://github.com/riverqueue/river/pull/1318).
+- If a job fails to unmarshal from JSON during job rescue or job execution, back off using the retry schedule and eventually discard it, similar to any other error that might occur. [PR #1324](https://github.com/riverqueue/river/pull/1324).
 
 ## [0.40.0] - 2026-07-02
 
--- a/internal/jobexecutor/job_executor.go
+++ b/internal/jobexecutor/job_executor.go
@@ -80,11 +80,12 @@ func MetadataUpdatesFromWorkContext(ctx context.Context) (map[string]any, bool) 
 }
 
 type jobExecutorResult struct {
-	Err             error
-	MetadataUpdates map[string]any
-	NextRetry       time.Time
-	PanicTrace      string
-	PanicVal        any
+	Err                error
+	JobArgsUnmarshaled bool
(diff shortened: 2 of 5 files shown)


brandur:
Tool failure on this one that caused the original title to be wrong. Correct PR title/description now.

bgentry (approved):

PR #1325, opened 2026-07-22
Move new pilot functions into a compatibility layer so we don't need immediate pro release

brandur:
In #1317 we added some new functions to the pilot to make the active job
rescue system work.

I was thinking this morning that the way this was done is a little
suboptimal for our pro customers. Expanding the pilot interface makes
the pro pilot immediately incompatible, so once we released this we'd
need to make sure to do an _immediate_ fast follow with a pro release,
otherwise there'd be no compatible version. And even with that fast
follow, if a customer tried to upgrade just River by itself, it'd break
their pro version without an upgrade.

Here, move the new functions into a special `PilotJobRescuer` interface
that's used where available by the rescuer, but allows a fall back in
case it's not. This will let us release the pro side at our leisure and
there will be multiple compatible River releases by then.

We'd want to unwind this eventually, so I left a couple comments in
place with instructions on how to do that.

### DIFF
--- a/internal/maintenance/job_rescuer.go
+++ b/internal/maintenance/job_rescuer.go
@@ -273,7 +273,7 @@ func (s *JobRescuer) runOnce(ctx context.Context) (*rescuerRunOnceResult, error)
 		}
 
 		if len(rescueManyParams.ID) > 0 {
-			_, err = s.Config.Pilot.JobRescueMany(ctx, s.exec, &rescueManyParams)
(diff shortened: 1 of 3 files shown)


brandur:
@bgentry Thoughts on this? Seems kind of nice as we won't have to rush as badly to get a Pro PR out.

bgentry (approved):
Great idea! :shipit:

brandur:
Excellent. Thanks!

PR #1327, opened 2026-07-23
Fix various small finalized-related bugs in job listing

### LINKED ISSUE #1326: JobListParams: OrderByFinalizedAt guard is inverted — rejects valid finalized-only filter, silently accepts meaningless non-finalized sort
## Bug

`JobListParams.toDBParams()` in `job_list_params.go` has an inverted guard when validating state filters for `JobListOrderByFinalizedAt`.

## Affected code (line ~244)

```go
if p.sortField == JobListOrderByFinalizedAt {
    currentNonFinalizedStates := make([]rivertype.JobState, 0, len(p.states))
    for _, state := range p.states {
        switch state {
        case rivertype.JobStateAvailable, rivertype.JobStatePending, ...:
            currentNonFinalizedStates = append(currentNonFinalizedStates, state)
        case rivertype.JobStateCancelled, rivertype.JobStateCompleted, rivertype.JobStateDiscarded:
        }
    }
    // Comment says: "user overrode the States list with only non-finalized states"
    if len(currentNonFinalizedStates) == 0 {   // ← INVERTED
        return nil, fmt.Errorf("cannot order by finalized_at with non-finalized state filters %+v",
            currentNonFinalizedStates)
    }
}
```

## What's wrong

The comment says the guard is checking for "only non-finalized states" — but the condition `len(currentNonFinalizedStates) == 0` is true when there are **zero** non-finalized states, meaning **all** states are finalized. That's actually the *correct* usage for `JobListOrderByFinalizedAt`.

The inverted conditions:

| States passed | `currentNonFinalizedStates` | `len == 0` | Result |
|---|---|---|---|
| Only finalized (e.g. `Cancelled`) | empty | true | ❌ Error returned (valid usage blocked) |
| Only non-finalized (e.g. `Available`) | `[Available]` | false | ✅ No error (meaningless sort allowed) |

When a caller passes only non-finalized states and orders by `finalized_at`, every result row has `finalized_at = NULL` — the sort is meaningless. But no error is returned. The one case that *should* work (all finalized states) is incorrectly rejected.

Additionally, the error message prints `currentNonFinalizedStates` (the empty slice `[]`) rather than the actual states being used, making the message uninformative.

## Suggested fix

```go
if len(currentNonFinalizedStates) == len(p.states) && len(p.states) > 0 {
    return nil, fmt.Errorf("cannot order by finalized_at with only non-finalized state filters %+v",
        currentNonFinalizedStates)
}
```

This fires when ALL states are non-finalized (the truly invalid case) and prints the actual problematic states in the error message.

brandur:
Fixes a few bugs related to use of finalized in job listing:

* Valid finalized-only queries were rejected because the validation
  condition was inverted. `JoblistOrderByFinalizedAt` was unusable as
  documented.

* Queries containing non-finalized states were accepted even though jobs
  would have no `finalized_at` value, making requested order meaningless.

* Mixed finalized and non-finalized filters were also accepted, leading
  to the same problem.

* Returning a non-finalized job could panic while constructing the
  pagination cursor because its `FinalizedAt` value is `nil`.

Fixes #1326, but also squashes a couple other related bugs that Codex
found while investigating.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Guard against empty job slice returned by `JobSetStateIfRunningMany` when a job has been deleted mid-run. [PR #1308](https://github.com/riverqueue/river/pull/1308).
 - Fixed `JobRescuer` pagination so a full batch of running jobs with disabled or longer worker-specific timeouts can't prevent later stuck jobs from being rescued. [PR #1318](https://github.com/riverqueue/river/pull/1318).
 - If a job fails to unmarshal from JSON during job rescue or job execution, back off using the retry schedule and eventually discard it, similar to any other error that might occur. [PR #1324](https://github.com/riverqueue/river/pull/1324).
+- Fixed `JobListOrderByFinalizedAt` validation so finalized states are accepted while non-finalized states are rejected. [PR #1327](https://github.com/riverqueue/river/issues/1327).
 
 ## [0.40.0] - 2026-07-02
 
--- a/job_list_params.go
+++ b/job_list_params.go
@@ -232,6 +232,10 @@ func (p *JobListParams) toDBParams() (*dblist.JobListParams, error) {
 	}
 
 	if p.sortField == JobListOrderByFinalizedAt {
+		if len(p.states) == 0 {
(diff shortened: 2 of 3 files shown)


bgentry (approved):
LGTM just make sure this plays nicely with riverui 🙏

brandur:
Thx! Had Codex do a review pass on River UI and it seems to think it all checks out.

PR #1332, opened 2026-07-26
Use configured time for job snoozes and retries

### LINKED ISSUE #1322: JobSnooze ignores Config.Test.Time when calculating scheduled_at
## Description

`Config.Test.Time` was introduced in #754 (following #751) to let tests provide a synthetic `rivertype.TimeGenerator`. Scheduled-job eligibility uses that configured time source, but `JobSnooze` calculates its next absolute timestamp from the process wall clock:

```go
nextAttemptScheduledAt := time.Now().Add(snoozeErr.Duration)
```

The executor then compares that wall-clock-derived timestamp with `e.Time.Now()`. When `Config.Test.Time` is intentionally ahead of or behind wall time, the two values belong to different timelines.

This is present in v0.39.0 and still appears on `master`.

## Example

Assume:

- process wall clock: `2026-07-22 12:00 UTC`
- `Config.Test.Time`: `2026-06-10 12:00 UTC`
- worker returns `river.JobSnooze(8 * time.Hour)`

The current implementation stores approximately:

```text
scheduled_at = 2026-07-22 20:00 UTC
```

The expected synthetic-time value is:

```text
scheduled_at = 2026-06-10 20:00 UTC
```

Advancing the synthetic clock by eight hours therefore does not make the job eligible; it remains scheduled until the wall-clock-derived timestamp. If the synthetic clock is ahead of wall time, the inverse can happen and the snoozed job may become immediately eligible.

This makes `Config.Test.Time` unreliable for behavioral or manual simulation tests that exercise workers which repeatedly snooze themselves.

## Suggested change

Use the executor's configured time generator when calculating the snooze target:

```diff
- nextAttemptScheduledAt := time.Now().Add(snoozeErr.Duration)
+ nextAttemptScheduledAt := e.Time.Now().Add(snoozeErr.Duration)
```

This is also consistent with the comparison immediately below, which already uses `e.Time.Now()`.

When `Config.Test.Time` is nil, River's default time generator still delegates to `time.Now()`, so ordinary runtime behavior should remain unchanged.

## Suggested regression test

Create a client/worker with `Config.Test.Time` fixed to an instant that differs materially from wall time, return `JobSnooze(duration)`, and assert that the resulting job's `ScheduledAt` equals the synthetic instant plus the snooze duration.

pablobaldez:
`Config.Test.Time` is used when River evaluates whether scheduled jobs
are eligible to run, but `JobSnooze` and the default retry policy derived
their next `scheduled_at` values from the process wall clock. When a
configured time generator differed from wall time, this mixed two
timelines in the same scheduling flow and could make snoozed or retried
jobs run too early or remain ineligible for much longer than intended.

Here, derive snooze timestamps from the executor's configured time and
initialize River's default retry policy from the same base service clock.
This also initializes the default fallback policy in producer and
`rivertest.Worker` execution paths. Custom retry policies are unchanged,
and zero-value use continues to fall back to wall time.

Regression coverage exercises both job snoozing and retrying through the
public client boundary with a configured time generator. Focused tests
and `make lint` pass. The full local `make test` run encounters existing
`TestProducer_PollOnly` timeouts that reproduce on the unchanged base
commit.

Fixes #1322.


rivertest/worker.go
```diff
@@ -192,7 +193,7 @@ func (w *Worker[T, TTx]) workJob(ctx context.Context, tb testing.TB, tx TTx, job
 		ClientJobTimeout:         w.config.JobTimeout,
 		ClientRetryPolicy:        w.config.RetryPolicy,
 		Completer:                completer,
-		DefaultClientRetryPolicy: &river.DefaultClientRetryPolicy{},
+		DefaultClientRetryPolicy: defaultClientRetryPolicy,
```
brandur: Can you move the `baseservice.Init` invocation inline here? No point in making the code more indirect by defining `defaultClientRetryPolicy` all the way up there.

retry_policy.go
```diff
@@ -27,9 +28,15 @@ type ClientRetryPolicy interface {
 
 // DefaultClientRetryPolicy is River's default retry policy.
 type DefaultClientRetryPolicy struct {
+	baseService baseservice.BaseService
 	timeNowFunc func() time.Time
 }
 
+// GetBaseService returns the retry policy's base service.
+func (p *DefaultClientRetryPolicy) GetBaseService() *baseservice.BaseService {
```
brandur: Hey @pablobaldez, ideally we don't expose this internal stuff in River's public API.

I asked Codex on how it might do this differently without exposing extra API and it suggested this:

> I’d extract the clock-aware implementation into an internal package:
> 
> ``` go
> // internal/retrypolicy/default.go
> type Default struct {
>     timeGenerator rivertype.TimeGenerator
> }
> 
> func NewDefault(timeGenerator rivertype.TimeGenerator) *Default {
>     return &Default{timeGenerator: timeGenerator}
> }
> 
> func (p *Default) NextRetry(job *rivertype.JobRow) time.Time {
>     return NextRetryAt(p.timeGenerator.Now().UTC(), job)
> }
> 
> func NextRetryAt(now time.Time, job *rivertype.JobRow) time.Time {
>     // Existing retry calculation and jitter.
> }
> ```
> 
> The public zero-value policy remains unchanged and delegates to the shared calculation:
> 
> ``` go
> type DefaultClientRetryPolicy struct {
>     timeNowFunc func() time.Time
> }
> 
> func (p *DefaultClientRetryPolicy) NextRetry(job *rivertype.JobRow) time.Time {
>     return retrypolicy.NextRetryAt(p.timeNowUTC(), job)
> }
> ```
> 
> Internally, River substitutes the clock-aware implementation when the default policy is selected:
> 
> ``` go
> if _, ok := config.RetryPolicy.(*DefaultClientRetryPolicy); ok {
>     config.RetryPolicy = retrypolicy.NewDefault(archetype.Time)
> }
> ```
> 
> Producer and rivertest.Worker fallback policies can use the internal constructor directly:
> 
> ``` go
> defaultClientRetryPolicy := retrypolicy.NewDefault(archetype.Time)
> ```
> 
> This would:
> 
> - Remove GetBaseService and the baseservice dependency from the public policy.
> - Preserve zero-value wall-clock behavior for direct external use.
> - Keep Config.Test.Time consistent in Client, producer fallback, and rivertest.
> - Avoid mutating a public policy instance if it’s shared between clients.
> - Add no externally accessible API because internal/retrypolicy cannot be imported outside River.

That seems like a reasonable thing to try. Thoughts?

pablobaldez: That makes sense. I agree that exporting `GetBaseService` only for internal wiring isn’t ideal. I’ll move the clock-aware implementation behind an internal retry policy, keep the public zero-value behavior unchanged, and inline the `rivertest` fallback as suggested. Thanks for the concrete proposal.

brandur (approved):
Looks good! Thank you.

pablobaldez:
@brandur Thanks again for merging this. Would you be open to cutting a patch release with #1331 and #1332? We're waiting on a tagged version to remove our local `JobSnooze` workaround.

brandur:
Yeah, we don't have any other changes right now that'd be going out in a release anytime soon, so can't hurt to cut a release.

PR #1248, opened 2026-05-11
"Fix" intermittent producer test: `Test_Producer_CanSafelyCompleteJobsWhileFetchingNewOnes`

brandur:
"Fix" an intermittently failing producer test observed in CI [1].

    --- FAIL: Test_Producer_CanSafelyCompleteJobsWhileFetchingNewOnes (21.87s)
        riverdbtest.go:220: Dropped 0 expired postgres schema(s) in 10.195µs
        producer_test.go:61: Generated postgres schema "river_2026_05_11t08_13_21_schema_01" with migrations [1 2 3 4 5 6] on line "main" in 252.908761ms [2 generated] [0 reused]
        logger.go:256: time=2026-05-11T08:13:27.411Z level=INFO msg="producer: Producer job counts" num_completed_jobs=1859 num_jobs_running=637 num_jobs_stuck=0 queue=default
        logger.go:256: time=2026-05-11T08:13:32.411Z level=INFO msg="producer: Producer job counts" num_completed_jobs=3713 num_jobs_running=611 num_jobs_stuck=0 queue=default
        logger.go:256: time=2026-05-11T08:13:37.411Z level=INFO msg="producer: Producer job counts" num_completed_jobs=4471 num_jobs_running=200 num_jobs_stuck=0 queue=default
        producer_test.go:160: timed out waiting for last job to run
        logger.go:256: time=2026-05-11T08:13:42.411Z level=INFO msg="producer: Producer job counts" num_completed_jobs=9317 num_jobs_running=274 num_jobs_stuck=0 queue=default
        riverdbtest.go:293: Checked in postgres schema "river_2026_05_11t08_13_21_schema_01"; 1 idle schema(s) [6 generated] [23 reused]
    FAIL
    FAIL	github.com/riverqueue/river	55.241s

I put "fix" in quotes here because this isn't a good fix because this
isn't a good test. It spins up a ton of jobs so that it's already slow
under the best of conditions, but on a slow machine there's real risk of
timeout. Even in the event of success, it takes a pretty long time and
makes CI pretty slow. We expand the allowed 20 seconds to 45 seconds,
but even if that succeeds, it means the test case may take right up to
45 seconds by itself.

Still, I understand that hot take might be controversial (lol), so I
avoided making anymore substantive changes for the time being.

[1] https://github.com/riverqueue/river/actions/runs/25658336810/job/75312202309?pr=1247

PR #1335, opened 2026-07-30
make producer race test deterministic

bgentry:
This is a draft alternative to #1248.

The existing race regression test processes 10,000 jobs and relies on the final job starting within a fixed deadline, so race-enabled CI can fail from throughput variance even when the intended concurrency path is exercised.

Replace it with a coordinated three-job scenario that blocks fetches after capacity is calculated. The test completes jobs while fetches are pending, continuously reads capacity across active-job removals and additions, and asserts each capacity transition. A queued third job preserves backlog/refill coverage, and the scenario runs through both poll-only and notifier producer setups.

This guarantees the completion-during-fetch overlap without a wall-clock throughput deadline and keeps unsafe active-job accounting visible to the race detector without adding production-only test instrumentation.

### DIFF
--- a/producer_test.go
+++ b/producer_test.go
@@ -5,13 +5,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"slices"
+	"sync/atomic"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/require"
 
 	"github.com/riverqueue/river/internal/jobcompleter"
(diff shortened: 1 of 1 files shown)


brandur (approved):

bgentry:
Oops thought I sent this comment earlier:

@brandur this is still not the easiest test to read, but I do think it retains the coverage of the old test while being fully deterministic.

PR #1337, opened 2026-07-30
Let job args return plugins like they can for hooks/middleware

brandur:
Here, implement a job args `jobArgsWithPlugins` interface that lets job
args return a set of `Plugins`. This follows up the addition of plugins
in #1284, and makes the functionality symmetrical to what's available
with hooks/middleware. I would have done this before, but I forgot that
you could do this until I was looking at docs today. As in #1284 for
client config, `Hooks` and `Middleware` continue to be supported on job
args and have not yet been deprecated.

    type EmailArgs struct {
        To string `json:"to"`
    }

    func (EmailArgs) Kind() string { return "email" }

    func (EmailArgs) Plugins() []rivertype.Plugin {
        return []rivertype.Plugin{
            &EmailPlugin{},
        }
    }

    type EmailPlugin struct {
        river.PluginDefaults
    }

    func (p *EmailPlugin) WorkBegin(ctx context.Context, job *rivertype.JobRow) error {
        return nil
    }

I also had Codex look for opportunities for implementation clean up
since all of this was feeling like a few too many LOCs. It was able to
clean up ~85 LOCs net, with the main change being to get rid of
`PluginLookupInterface` + `emptyPluginLookup`. This removes a layer of
indirection, which is good, but also `emptyPluginLookup` had become less
useful due to River installing default middleware (`ResumableMiddleware`)
on all clients whether any has been specified by a client or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `JobArgsWithPlugins` for installing plugins that apply only to a specific job type. [PR #1337](https://github.com/riverqueue/river/pull/1337).
+
 ## [0.41.1] - 2026-07-29
 
 ### Fixed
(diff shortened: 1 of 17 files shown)


bgentry (approved):

brandur:
Thx!

PR #1339, opened 2026-08-01
Add tag filtering to JobListParams

bgentry:
Adds exact tag filtering to job lists through explicit `JobListParams.TagsAny`
and `JobListParams.TagsAll` methods. The two filters can be combined, with
match-any and match-all groups joined to each other and other filters using
AND.

Tag predicates are added only when requested through driver-specific SQL
fragments. PostgreSQL uses native array overlap and containment operators,
while SQLite uses equivalent JSON predicates. `JobDeleteManyParams` is defined
from the list parameter type so their existing pointer conversion remains
structurally safe.

This is the core dependency for
[riverqueue/riverui#548](https://github.com/riverqueue/riverui/pull/548).

### DIFF
--- a/client.go
+++ b/client.go
@@ -2398,7 +2398,7 @@ func (c *Client[TTx]) jobDeleteMany(ctx context.Context, exec riverdriver.Execut
 		return nil, errors.New("delete with no filters not allowed to prevent accidental deletion of all jobs; either specify a predicate (e.g. JobDeleteManyParams.IDs, JobDeleteManyParams.Kinds, ...) or call JobDeleteManyParams.All")
 	}
 
-	listParams, err := dblist.JobMakeDriverParams(ctx, params.toDBParams(), c.driver.SQLFragmentColumnIn)
+	listParams, err := dblist.JobMakeDriverParams(ctx, params.toDBParams(), c.driver)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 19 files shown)


client.go
```diff
@@ -2403,7 +2403,13 @@ func (c *Client[TTx]) jobDeleteMany(ctx context.Context, exec riverdriver.Execut
 		return nil, err
 	}
 
-	jobs, err := exec.JobDeleteMany(ctx, (*riverdriver.JobDeleteManyParams)(listParams))
+	jobs, err := exec.JobDeleteMany(ctx, &riverdriver.JobDeleteManyParams{
```
bgentry: ah this needs to be kept aligned with list params, will fix

job_list_params.go
```diff
@@ -419,6 +422,15 @@ func (p *JobListParams) States(states ...rivertype.JobState) *JobListParams {
 	return paramsCopy
 }
 
+// Tags returns an updated filter set that will only return jobs containing at
+// least one of the given tags. Tag matching is case-insensitive.
```
bgentry: @brandur while this copies what the other PR does, it doesn't feel right now that I read this comment. Wondering if we need two separate methods for filtering by all of vs any of the tags in the list?

brandur (approved):
Nice one! Looks good to me.

job_list_params.go
brandur: Looks like you put in the two separate functions and this comment is outdated. Looks okay as is.

PR #1342, opened 2026-08-04
Stop linking testify into production binaries via testsignal

e-yavuz-1:
`testsignal` is production code — its own doc comment says it's designed to have "minimal impact on the production code that calls into it" — but it imported `riversharedtest` for a single function, `WaitTimeout`.

Go initializes every imported package, so nothing in that graph can be eliminated by the linker. The result is that testify, go-spew, goleak, and yaml end up in every binary that uses River. Measured with `go tool nm -size` on a minimal program importing `river` and `riverdriver/riverpgxv5`:

    github.com/stretchr/testify  18 symbols / 1168 bytes
    github.com/davecgh/go-spew   11 symbols / 1312 bytes
    go.uber.org/goleak            8 symbols /  896 bytes
    gopkg.in/yaml                22 symbols / 4320 bytes
    riversharedtest              13 symbols / 2248 bytes

`WaitTimeout` depends only on `os` and `time`, so move it to `rivershared/util/testutil`, which `testsignal` already imports and which has no dependencies outside the standard library. `riversharedtest.WaitTimeout` stays as a wrapper so no public API changes, and its two internal callers, `WaitOrTimeout` and `WaitOrTimeoutN`, are untouched. Its test moves alongside the implementation.

After the move, the same program links no testify, go-spew, goleak, yaml, or riversharedtest symbols at all: 951 fewer symbols and 957 KB less binary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `JobListParams.TagsAll` and `JobListParams.TagsAny` for filtering jobs that match every or any exact tag, respectively. [PR #1339](https://github.com/riverqueue/river/pull/1339).
 
+### Fixed
+
+- `testsignal` no longer imports `riversharedtest`, so testify, go-spew, and goleak are no longer linked into production binaries that use River. `WaitTimeout` moved to `rivershared/util/testutil`, with `riversharedtest.WaitTimeout` kept as a wrapper around it.
+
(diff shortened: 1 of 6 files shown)


brandur:
@e-yavuz-1 Dang great find. Do you think could sign our CLA? https://github.com/riverqueue/rivercla

Otherwise, LGTM.

brandur:
Also adding #1343 for a more structural fix by checking for this problem during linting.

e-yavuz-1:
@brandur CLA signed.

brandur (approved):

PR #1343, opened 2026-08-04
Don't allow `testsignal` to pick up non-stdlib dependencies

brandur:
This one's aimed at producing a more permanent structural fix for #1342
by keeping an eye out for the same problematic condition using the
depguard lint. As a summary of the problem corrected by #1342:

* We use the `testsignal` package in all kinds of non-test code, so
  packages including River always build against it.

* `testsignal` was importing `riversharedtest`, which imports Testify,
  Goleak, YAML (through Testify).

* This was causing all packages built against River to pick up an extra
  ~10 kB worth of dependencies in their production builds, which is bad.

Here, add some new rules for depguard:

* Don't allow test packages like Goleak or Testify to be imported by any
  non-test Go files. We make an exception for internal test support
  packages like `riverdbtest` and `riverdrivertest`.

* Don't allow `testsignal` to have any dependencies beyond stdlib and
  `testutil`.

* Don't allow `testutil` to have any non-stdlib dependencies. This rule
  is so that `testsignal` doesn't pick up unexpected dependencies
  transitively through `testutil`.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -52,6 +52,28 @@ linters:
             - desc: Don't use `riverinternaltest` package outside of test environments.
               pkg: github.com/riverqueue/river/internal/riverinternaltest
 
+        # The next two blocks have the same intent: don't allow testsignal,
+        # which is used in non-test code, to have non-stdlib dependencies.
+        # Previously, we ran into a problem where it was accidentally importing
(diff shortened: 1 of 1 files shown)


brandur:
Lint currently failing due to https://github.com/riverqueue/river/pull/1342 not being merged yet.

bgentry (approved):

PR #1347, opened 2026-08-11
Support Yugabyte as a Postgres target

### LINKED ISSUE #1346: Yugabyte compatibility issues when inserting new tasks
We're investigating a migration from PostgreSQL to Yugabyte (a PostgreSQL-compatible distributed database re-using the PostgreSQL querier) and running into compatibility issues with the xmax column used during upserts, that is currently unsupported (see https://docs.yugabyte.com/stable/yugabyte-voyager/known-issues/postgresql/#system-columns-is-not-yet-supported and https://github.com/yugabyte/yugabyte-db/issues/24843 ).

Exact symptoms are the following error at each task insertion: `ERROR: system column "xmax" is not supported yet (SQLSTATE 0A000)
`. Root cause seems to be the SQL statement at https://github.com/riverqueue/river/blob/7ac5a2cbfe2067e0a4aab7b835302287f863effa/riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql#L321

It seems a workaround is already in place for SQLite, could it be possible to expose it through an option or something similar?

brandur:
This one's aimed at #1346, in which it might be possible for us to
support Yugabyte as a database target without a hugely inordinate amount
of work.

Yugabyte is currently targeting compatibility against Postgres 15 [1].
It doesn't support `xmax` which is what #1346 is about, but somewhat
surprisingly, we only use `xmax` in one place and don't use any other
Postgres 16+ features (as Postgres 15 is still a valid target in the
CI matrix).

The `xmax` trick to determine whether an upserted row is new or existing
is a little outdated anyway because Postgres 18 added the capability to
detect an existing row with `OLD.id IS NOT NULL` [2].

Long run, we should switch to that for everything. Shorter term,
Postgres 18 is still quite new, so I propose we do something like this:

* If on Postgres 18+ (we should be getting Postgres 19 soon), use
  `OLD.id IS NOT NULL`.

* If on Yugabyte, fall back to the same trick we use in SQLite by
  upserting rows with a unique nonce and checking whether the nonce was
  the one we inserted or not.

* Otherwise, use the existing approach with `xmax`.

We do have to check which database we're on, but only once, after which
we can cache that information forever, so it shouldn't have any impact
on performance.

Fixes #1346.

[1] https://docs.yugabyte.com/stable/faq/compatibility/#what-is-the-extent-of-compatibility-with-postgresql
[2] https://www.crunchydata.com/blog/postgres-18-old-and-new-in-the-returning-clause

### DIFF
--- a/client.go
+++ b/client.go
@@ -1110,8 +1110,12 @@ func (c *Client[TTx]) Start(ctx context.Context) error {
 		// available, the client appears to have started even though it's completely
 		// non-functional. Here we try to make an initial assessment of health and
 		// return quickly in case of an apparent problem.
-		if err := c.driver.GetExecutor().Exec(fetchCtx, "SELECT 1"); err != nil {
+		executor := c.driver.GetExecutor()
+		if err := executor.Ping(fetchCtx); err != nil {
 			return fmt.Errorf("error making initial connection to database: %w", err)
+		}
+		if err := executor.InitDriver(fetchCtx); err != nil {
+			return fmt.Errorf("error initializing driver: %w", err)
 		}
 
 		// Each time we start, we need a fresh completer subscribe channel to
--- a/riverdriver/riverpgxv5/river_pgx_v5_driver.go
+++ b/riverdriver/riverpgxv5/river_pgx_v5_driver.go
@@ -16,6 +16,7 @@ import (
 	"math"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/jackc/pgx/v5"
@@ -30,6 +31,7 @@ import (
 	"github.com/riverqueue/river/rivershared/uniquestates"
 	"github.com/riverqueue/river/rivershared/util/dbutil"
 	"github.com/riverqueue/river/rivershared/util/ptrutil"
+	"github.com/riverqueue/river/rivershared/util/randutil"
 	"github.com/riverqueue/river/rivershared/util/sliceutil"
 	"github.com/riverqueue/river/rivertype"
 )
@@ -39,8 +41,9 @@ var migrationFS embed.FS
 
 // Driver is an implementation of riverdriver.Driver for Pgx v5.
 type Driver struct {
-	dbPool   *pgxpool.Pool
-	replacer sqlctemplate.Replacer
+	dbPool           *pgxpool.Pool
+	replacer         sqlctemplate.Replacer
+	uniqueInsertMode atomic.Uint32
 }
 
 // New returns a new Pgx v5 River driver for use with River.
@@ -216,6 +219,11 @@ func (e *Executor) IndexesExist(ctx context.Context, params *riverdriver.Indexes
 	return exists, nil
 }
 
+func (e *Executor) InitDriver(ctx context.Context) error {
+	_, err := e.uniqueInsertMode(ctx)
+	return err
+}
+
 func (e *Executor) JobCancel(ctx context.Context, params *riverdriver.JobCancelParams) (*rivertype.JobRow, error) {
 	cancelledAt, err := params.CancelAttemptedAt.MarshalJSON()
 	if err != nil {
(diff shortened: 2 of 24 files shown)


riverdriver/unique_insert.go
```diff
@@ -0,0 +1,108 @@
...
+	var metadataNonce string
+	if err := json.Unmarshal(metadataMap[UniqueInsertMetadataKey], &metadataNonce); err != nil {
+		return true
+	}
+	return metadataNonce != nonce
+}
+
+// UniqueInsertMetadataWithNonce returns metadata with nonce set under
+// UniqueInsertMetadataKey.
+func UniqueInsertMetadataWithNonce(metadata []byte, nonce string) ([]byte, error) {
```
brandur, line 74: This approach originally comes from SQLite, but extracted out to here to be reusable.

brandur:
@bgentry Thoughts on this?

I'm kind of thinking that we wouldn't be able to _officially_ support Yugabyte since I really don't want to be testing against it, but we could have soft support that works as long as they commit to their stated PG 15 contract with a few very minor deviations (like the `xmax` thing which admittedly is a little bit abusive of Postgres anyway).

jqueuniet:
I tested this branch in our test environment and our service managed to both enqueue asynchronous tasks and run scheduled ones without any other error.

brandur:
@jqueuniet Excellent! Thanks for checking. Next we'll see if it can stand the stress of a DB-based job queue ...

client.go
```diff
@@ -1110,7 +1129,7 @@ func (c *Client[TTx]) Start(ctx context.Context) error {
 		// available, the client appears to have started even though it's completely
 		// non-functional. Here we try to make an initial assessment of health and
 		// return quickly in case of an apparent problem.
-		if err := c.driver.GetExecutor().Exec(fetchCtx, "SELECT 1"); err != nil {
+		if err := c.driver.GetExecutor().Ping(fetchCtx); err != nil {
```
bgentry: :robot: **via Codex:** This changes the startup health check into a one-time capability initialization. Once `uniqueInsertMode` has been cached, normally by `initialPing`, `Ping` returns successfully without performing any database I/O.

A later `Start` or restart can therefore succeed while PostgreSQL is unavailable, which is the exact case this block is intended to reject, especially for poll-only and `database/sql` clients.

Could we keep a real `SELECT 1` or pool ping on every `Start` and initialize the insert mode separately, or make `Ping` always reach the database?

riverdriver/riverpgxv5/river_pgx_v5_driver.go
```diff
@@ -1036,6 +1073,32 @@ func (e *Executor) TableTruncate(ctx context.Context, params *riverdriver.TableT
 	return interpretError(err)
 }
 
+func (e *Executor) uniqueInsertMode(ctx context.Context) (riverdriver.UniqueInsertMode, error) {
+	if e.driver != nil {
+		if mode := riverdriver.UniqueInsertMode(e.driver.uniqueInsertMode.Load()); mode != riverdriver.UniqueInsertModeUnknown {
+			return mode, nil
+		}
+
+		e.driver.uniqueInsertModeInitMu.Lock()
```
bgentry: :robot: **via Codex:** Holding this non-context-aware mutex across `PGGetProductAndVersion` can make inserts wait for the full 10-second `initialPing` timeout even after their own context is canceled.

There is also a `MaxConns=1` cycle: an existing transaction holds the sole connection, `initialPing` takes this mutex and waits for the pool, and `InsertTx` uses the existing transaction but blocks on this mutex. Progress resumes only when `initialPing` times out. The new test pauses before the real `Ping` takes this lock, so it does not cover the cycle.

Could we avoid background initialization and resolve lazily, or permit redundant detection queries or context-aware waiting instead of holding a mutex across database I/O? The `database/sql` implementation has the same issue.

client.go
brandur: So both comments here were on code that I'd brought in near the last second — previously, I'd had the drivers lazily check whether a database is Yugabyte by doing it on first insert, and the idea here was to do an aggressive database check on client start. As uncovered by these issues though, doing so leads to added complexity and a number of possible races, so I think the whole scheme was more trouble than it was worth, and have now removed it.

riverdriver/riverpgxv5/river_pgx_v5_driver.go
brandur: See comment above.

brandur:
@bgentry Pushed a new version of this. Mind taking a look?

PR #1348, opened 2026-08-12
Periodic job next_run_at advances on insert failure — missing return after error log

tsushanth:
Add `return` after insert-error log in `insertBatch` to prevent `PeriodicJobUpsertMany` from advancing `next_run_at` when job insertion fails

Found via automated repo scanning, fix written and reviewed before opening.


brandur (approved):
Thanks! Do you want to add a short changelog entry for this?

PR #1350, opened 2026-08-13
Remove post-persist ambiguity and have events match emitted state

brandur:
This one's aimed at fixing [1] in which for some sequences in workers,
we'd emit a surprising event based on the state that was actually
persisted to the database. This contract was also not stable, and
changed subtly with the introduction of #1290.

From [1], this is best illustrated by a short example where we error
after invoking `JobCompleteTx`:

    func (w *Worker) Work(ctx context.Context, job *river.Job[Args]) error {
          tx, _ := w.dbPool.Begin(ctx)
          defer tx.Rollback(ctx)

          river.JobCompleteTx[*riverpgxv5.Driver](ctx, tx, job) // row -> completed
          tx.Commit(ctx)

          return errors.New("boom") // executor reports an error, but its UPDATE is an IfRunning no-op
    }

This used to emit a `job_completed`, but has changed in `master` to emit
a `job_failed`.

Here, we try to correct the emitted event and officially standardize it:

| Scenario | Persisted state | Previous event | New event |
|---|---:|---:|---:|
| `JobCompleteTx` commits, then worker returns an error | `completed` | `job_failed` | `job_completed` |
| Remote cancellation, then worker requests a retry | `cancelled` | `job_failed` | `job_cancelled` |
| Remote cancellation, then worker snoozes | `cancelled` | `job_snoozed` | `job_cancelled` |

Unambiguous persisted states always take precedence, though we retain
reason (added in #1290) to distinguish possible states of `available`,
which may be (1) an immediate retry after failure, (2) a short snooze,
or (3) an interruption caused by client shutdown.

[1] https://github.com/riverqueue/river/pull/1290#issuecomment-5258510559

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Write timestamps in SQLite to always include three digits after the second like `.000`. Previously, they may have been truncated down to just `.0` in the case of trailing zeroes. [PR #1349](https://github.com/riverqueue/river/pull/1349)
+- Job completion events now reflect the job's persisted outcome when it differs from the transition requested by the worker. For example, a job completed with `JobCompleteTx` before its worker returns an error emits `job_completed`, and a remotely cancelled job whose worker errors or snoozes emits `job_cancelled`. Applications subscribed only to `job_failed` or `job_snoozed` should note that these events may instead be delivered to `job_completed` or `job_cancelled` subscribers. [PR #1350](https://github.com/riverqueue/river/pull/1350)
 
 ## [0.43.0] - 2026-08-05
 
--- a/internal/jobcompleter/job_completer.go
+++ b/internal/jobcompleter/job_completer.go
@@ -48,6 +48,51 @@ type CompleterJobUpdated struct {
 	Reason   riverdriver.JobSetStateReason
 }
 
+func completerJobUpdatedFromStateAndReason(job *rivertype.JobRow, stats *jobstats.JobStatistics, requestedReason riverdriver.JobSetStateReason) CompleterJobUpdated {
+	var reason riverdriver.JobSetStateReason
+	switch job.State {
+	case rivertype.JobStateAvailable:
+		// Available is ambiguous: failed jobs being retried immediately, short
+		// snoozes, and jobs interrupted by client shutdown all use it. Preserve
+		// the requested reason when it describes one of those transitions.
+		switch requestedReason {
+		case riverdriver.JobSetStateReasonFailed, riverdriver.JobSetStateReasonInterrupted, riverdriver.JobSetStateReasonSnoozed:
+			reason = requestedReason
+		case riverdriver.JobSetStateReasonCancelled, riverdriver.JobSetStateReasonCompleted:
+			reason = riverdriver.JobSetStateReasonFailed
+		default:
+			panic("completion subscriber received an unknown reason, river bug")
+		}
+	case rivertype.JobStateCancelled:
+		reason = riverdriver.JobSetStateReasonCancelled
+
+	case rivertype.JobStateCompleted:
+		reason = riverdriver.JobSetStateReasonCompleted
+
+	case rivertype.JobStateDiscarded, rivertype.JobStateRetryable:
+		reason = riverdriver.JobSetStateReasonFailed
+
+	case rivertype.JobStateScheduled:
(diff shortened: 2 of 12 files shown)


riverdriver/riverpgxv5/internal/dbsqlc/river_job.sql
```diff
@@ -640,7 +640,7 @@ updated AS (
     SET
         attempt = CASE
             WHEN river_job.state = 'running'
-                 AND NOT (job_input.state IN ('retryable','scheduled') AND river_job.metadata ? 'cancel_attempted_at')
+                 AND NOT (job_input.state IN ('available','retryable','scheduled') AND river_job.metadata ? 'cancel_attempted_at')
```
brandur, line 643: These are kind of a related fix that's also adjacent. A job completer may set a job back to `available` if the scheduled time was very short in the future. Adding `available` here makes it so that a pending cancel wins if this was about to occur which is the more correct behavior (it was a bug before that this was wasn't considered).

brandur:
@bgentry This does seem like a useful fix for previously somewhat ambiguous behavior. Seem okay?

mitar:
FYI, you are making references to #1219 but I think you want to make references to #1290. #1219 is my PR which has not yet landed (but I would love for it to - and it does push similar question further for two new *Tx functions).

mitar:
I tested this branch in combination and #1219 and it fixes all the issues I encountered while rebasing it. Thanks!

BTW, from code it looks to me like `newCompleterJobUpdated` panics on `JobStateRunning`/`JobStatePending`, is this intentional?

brandur:
> FYI, you are making references to https://github.com/riverqueue/river/pull/1219 but I think you want to make references to https://github.com/riverqueue/river/pull/1290. https://github.com/riverqueue/river/pull/1219 is my PR which has not yet landed (but I would love for it to - and it does push similar question further for two new *Tx functions).

Thx. Fixed.

> BTW, from code it looks to me like `newCompleterJobUpdated` panics on `JobStateRunning`/`JobStatePending`, is this intentional?

Yeah, basically by the time we hit `newCompleterJobUpdated`, River should only ever be dealing with a job in a finalized state. The panic could be something else, but it quickly bubbles up a situation where we've got something unexpected going on. I added a comment on those lines to explain the situation better.

internal/jobcompleter/job_completer.go
```diff
@@ -48,6 +48,51 @@ type CompleterJobUpdated struct {
 	Reason   riverdriver.JobSetStateReason
 }
 
+func newCompleterJobUpdated(job *rivertype.JobRow, stats *jobstats.JobStatistics, requestedReason riverdriver.JobSetStateReason) CompleterJobUpdated {
```
bgentry: I might opt to have this method be only focused on the state/reason since afaict that's all it's actually returning, but it's minor so up to you for sure!

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Write timestamps in SQLite to always include three digits after the second like `.000`. Previously, they may have been truncated down to just `.0` in the case of trailing zeroes. [PR #1349](https://github.com/riverqueue/river/pull/1349)
+- Job completion events now reflect the job's persisted outcome when it differs from the transition requested by the worker. For example, a job completed with `JobCompleteTx` before its worker returns an error emits `job_completed`, and a remotely cancelled job whose worker errors or snoozes emits `job_cancelled`. [PR #1350](https://github.com/riverqueue/river/pull/1350)
```
bgentry: Do you think this deserves any further detail on how this might impact people whose programs are written against these states?

bgentry (approved):

internal/jobcompleter/job_completer.go
brandur: Renamed: `completerJobUpdatedFromStateAndReason`

CHANGELOG.md
brandur: I added this sentence to the existing changelog to clarify that aspect:

>  Applications subscribed only to `job_failed` or `job_snoozed` should note that these events may instead be delivered to `job_completed` or `job_cancelled` subscribers.

TBH though, I am kind of hoping that this will be quite rare for users. It's good to be complete in the changelog, but the chances of someone catching a bug by reviewing what happened made in the changelog is probably fairly low. Albeit, maybe substantially higher if AI's involved in the upgrade.

brandur:
Thanks!

PR #1352, opened 2026-08-18
Prepare version 0.44.0

brandur:
Get a release out the door for 0.44.0. There's one small SQLite
compatibility timestamp-related fix that we put in (#1349), and it'd be
nice to have this for River Pro's upcoming release so we're doing the
same thing everywhere.

[skip ci]


bgentry (approved):

brandur:
thx!

PR #1353, opened 2026-08-20
Make sure to use `timeString`/`timeStringNullable` everywhere for SQLite

brandur:
Make sure to use `timeString`/`timeStringNullable` everywhere for SQLite
rather than passing a `time.Time` or `*time.Time`. Using the latter
works, but causes the underlying driver to default to whatever
serialization method for the time it wants to use, making it a different
format compared to what we normally store, and leading to requiring
liberal input unmarshaling as times might exist in a variety of formats
(we had a recent River Pro bug where this was the case).

I also checked River Pro for problems of this sort, but we're clean over
there already.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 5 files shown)


bgentry (approved):

brandur:
thx!

PR #1356, opened 2026-08-21
Upgrade to Go 1.27 + golangci-lint 2.13.1

brandur:
Just since everything's fully available in Homebrew etc. now, here
take the opportunity upgrade Go to 1.27 and golangci-lint  to 2.13.1 so
we can get our lint CI fully passing again.


bgentry (approved):
Looks like there's an old linter to fix but otherwise LGTM

PR #1357, opened 2026-08-22
Modernize for Go 1.26

brandur:
Follows up #1356 to bring back in some modernizers that I'd disabled to
not make the diff for that original PR gigantic.

In particular:

* `errorsastype`: Requires the use of the new `errors.AsType` helper.
  This is a really nice one and turns this old code:

        var pgErr *pgconn.PgError
        if errors.As(err, &pgErr) {

    Into this much nicer generics variant:

        if pgErr, ok := errors.AsType[*pgconn.PgError](err); ok {

* `newexpr`: Converts pointer helpers like `ptrutil.Ptr(...)` to `new`.
  I remove all uses of `ptrutil.Ptr(...)`, but left its definition in
  `rivershared` for now because I think it'll allow a little more
  flexibility in what version of River a River Pro installation can use.
  For example, an older version of River Pro could potentially point to
  a newer version of River and still have compilation work (it has
  however been marked as deprecated). I doubt too many people are in
  this situation though so we can probably remove it soon.

* `stditerators`: Replaces some manual iteration patterns with newer
  standard library iterator APIs. For example this:

        for i := range typ.NumField() {
            field := typ.Field(i)

    Goes to this:

        for field := range typ.Fields() {[27;5;106~

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -93,13 +93,6 @@ linters:
       excludes:
         - G404 # use of non-crypto random; overly broad for our use case
 
-    modernize:
-      disable:
-        # Keep the Go 1.26 upgrade focused; apply its modernizations separately.
-        - errorsastype
-        - newexpr
-        - stditerators
-
     revive:
       rules:
         - name: unused-parameter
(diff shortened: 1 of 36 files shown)


bgentry (approved):

brandur:
thx!

PR #1359, opened 2026-08-22
Fix periodic state advance on insert error

### LINKED ISSUE #1304: Periodic job next_run_at can advance after job insert failure
# Periodic job next_run_at can advance after job insert failure

## Summary

`PeriodicJobEnqueuer.insertBatch` starts a transaction, calls the configured
periodic job insert function, and then upserts durable periodic job
`next_run_at` records.

If the insert function returns an error, `insertBatch` logs the error but still
continues to `PeriodicJobUpsertMany` and can commit the transaction. For insert
errors that happen before any database write, the transaction is still usable,
so the durable next-run record can advance even though no job was inserted for
the due occurrence.

on Linux. I found related periodic-job discussions and docs, but not an exact
public report for this insert-error/next-run advancement case.

## Why this seems wrong

- The durable `next_run_at` record says the due occurrence has been accounted
  for, but the corresponding job row may not exist.
- A later leader or restarted process can observe the advanced durable record
  and skip the missed occurrence.
- Insert callbacks can fail before touching the database, for example due to
  validation, middleware, or application insertion logic.
- The transaction already groups the insert and the next-run upsert, so
  aborting on insert error preserves the natural all-or-nothing boundary.

## Relevant code

In `internal/maintenance/periodic_job_enqueuer.go`, `insertBatch` logs the
insert error but then falls through to the durable next-run upsert:

```go
if len(insertParamsMany) > 0 {
    if _, err := s.Config.Insert(ctx, tx, insertParamsMany); err != nil {
        s.Logger.ErrorContext(ctx, s.Name+": Error inserting periodic jobs",
            "error", err.Error(), "num_jobs", len(insertParamsMany))
    }
}

if len(periodicJobUpsertParams.Jobs) > 0 {
    if _, err = s.Config.Pilot.PeriodicJobUpsertMany(ctx, tx, periodicJobUpsertParams); err != nil {
        s.Logger.ErrorContext(ctx, s.Name+": Error upserting periodic job next run times",
            "error", err.Error(), "num_jobs", len(insertParamsMany), "num_next_run_at_upserts", len(periodicJobUpsertParams.Jobs))
        return
    }
}
```

## Reproduction

Add a focused test to `internal/maintenance/periodic_job_enqueuer_test.go`
using an insert callback that fails before touching the transaction and a pilot
mock that records whether the durable upsert was called:

```go
t.Run("DoesNotAdvanceDurableNextRunAfterInsertError", func(t *testing.T) {
    t.Parallel()

    exec := &periodicInsertBatchExecutorMock{}
    pilotMock := NewPilotPeriodicJobMock()

    svc, err := NewPeriodicJobEnqueuer(riversharedtest.BaseServiceArchetype(t), &PeriodicJobEnqueuerConfig{
        Insert: func(ctx context.Context, tx riverdriver.ExecutorTx, insertParams []*rivertype.JobInsertParams) ([]*rivertype.JobInsertResult, error) {
            return nil, errors.New("insert failed before database write")
        },
        Pilot:  pilotMock,
        Schema: "river_test_schema",
    }, exec)
    require.NoError(t, err)
    svc.TestSignals.Init(t)

    var upsertCalled bool
    pilotMock.PeriodicJobUpsertManyMock = func(ctx context.Context, exec riverdriver.Executor, params *riverpilot.PeriodicJobUpsertManyParams) ([]*riverpilot.PeriodicJob, error) {
        upsertCalled = true
        return nil, nil
    }

    now := time.Now().UTC()
    svc.insertBatch(ctx, []*rivertype.JobInsertParams{{Kind: "periodic_insert_error", Queue: rivercommon.QueueDefault, State: rivertype.JobStateAvailable}}, &riverpilot.PeriodicJobUpsertManyParams{
        Jobs: []*riverpilot.PeriodicJobUpsertParams{{ID: "periodic_insert_error", NextRunAt: now.Add(time.Minute), UpdatedAt: now}},
        Schema: "river_test_schema",
    })

    require.False(t, upsertCalled)
    require.False(t, exec.tx.committed)
    require.True(t, exec.tx.rolledBack)
})
```

Run:

```bash
go test ./internal/maintenance -run "TestPeriodicJobEnqueuer/DoesNotAdvanceDurableNextRunAfterInsertError" -count=1 -v
```

Current result:

```text
level=ERROR msg="maintenance.PeriodicJobEnqueuer: Error inserting periodic jobs" error="insert failed before database write" num_jobs=1
Error:       Should be false
Test:        TestPeriodicJobEnqueuer/DoesNotAdvanceDurableNextRunAfterInsertError
```

The failing assertion checks that `PeriodicJobUpsertMany` was not called after
the insert callback failed.

## Expected behavior

If inserting the due periodic job fails, the batch should not commit the next
`next_run_at` for that same occurrence.

## Suggested fix

Return immediately after an insert error so the deferred rollback runs and the
next-run upsert is not committed:

```go
if _, err := s.Config.Insert(ctx, tx, insertParamsMany); err != nil {
    s.Logger.ErrorContext(ctx, s.Name+": Error inserting periodic jobs",
        "error", err.Error(), "num_jobs", len(insertParamsMany))
    return
}
```

One follow-up consideration: the scheduler currently advances the in-memory
`nextRunAt` before `insertBatch` runs. If immediate retry in the same process
is desired after an insert failure, that in-memory advancement should also be
deferred until the batch commits or rolled back when the batch aborts.

BreezeDelegate:
Stops `insertBatch` before the durable `next_run_at` upsert when periodic job insertion fails.

Adds a regression test for the insert-error path.

Fixes #1304

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Upgrade supported Go versions to 1.26 and 1.27. [PR #1356](https://github.com/riverqueue/river/pull/1356).
 
+### Fixed
+
+- Fixed periodic jobs advancing their durable next run time when job insertion fails. [PR #1359](https://github.com/riverqueue/river/pull/1359).
+
 ## [0.44.1] - 2026-08-21
 
 ### Fixed
--- a/internal/maintenance/periodic_job_enqueuer.go
+++ b/internal/maintenance/periodic_job_enqueuer.go
@@ -537,6 +537,7 @@ func (s *PeriodicJobEnqueuer) insertBatch(ctx context.Context, insertParamsMany 
 		if _, err := s.Config.Insert(ctx, tx, insertParamsMany); err != nil {
 			s.Logger.ErrorContext(ctx, s.Name+": Error inserting periodic jobs",
 				"error", err.Error(), "num_jobs", len(insertParamsMany))
+			return
 		}
 	}
 
--- a/internal/maintenance/periodic_job_enqueuer_test.go
+++ b/internal/maintenance/periodic_job_enqueuer_test.go
@@ -884,6 +884,38 @@ func TestPeriodicJobEnqueuer(t *testing.T) {
 		require.WithinDuration(t, now.Add(2*time.Hour), svc.periodicJobs[svc.periodicJobIDs["pilot_startup_durable_2h"]].nextRunAt, time.Microsecond)
 	})
 
+	t.Run("InsertErrorDoesNotAdvanceDurableNextRun", func(t *testing.T) {
+		t.Parallel()
+
+		svc, bundle := setup(t)
+
+		svc.Config.Insert = func(ctx context.Context, tx riverdriver.ExecutorTx, insertParams []*rivertype.JobInsertParams) ([]*rivertype.JobInsertResult, error) {
+			return nil, errors.New("insert failed before database write")
+		}
+
+		var periodicJobUpsertManyMockCalled bool
+		bundle.pilotMock.PeriodicJobUpsertManyMock = func(ctx context.Context, exec riverdriver.Executor, params *riverpilot.PeriodicJobUpsertManyParams) ([]*riverpilot.PeriodicJob, error) {
+			periodicJobUpsertManyMockCalled = true
+			return nil, nil
+		}
+
+		now := time.Now().UTC()
+		svc.insertBatch(ctx, []*rivertype.JobInsertParams{{
+			Kind:  "periodic_insert_error",
+			Queue: rivercommon.QueueDefault,
+			State: rivertype.JobStateAvailable,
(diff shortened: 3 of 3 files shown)


brandur:
Hah, we actually had one other person find this in #1348, but they stopped replying to the PR.

Do you want to add a changelog entry and add yourself to the CLA? https://github.com/riverqueue/rivercla

BreezeDelegate:
Yep, happy to. I added the changelog entry. For the CLA, is it okay to sign under my usual GitHub name/pseudonym (BreezeDelegate) rather than publishing my legal name?

brandur:
> Yep, happy to. I added the changelog entry. For the CLA, is it okay to sign under my usual GitHub name/pseudonym (BreezeDelegate) rather than publishing my legal name?

Okay that works. Just make sure changes are tied back to the same GitHub handle.

BreezeDelegate:
CLA submitted: riverqueue/rivercla#32.

PR #1360, opened 2026-08-23
Fix SQLite filter used in River UI queue count pane

brandur:
I noticed while testing SQLite out in the River UI that it's fairly slow
when you click on the "queues" tab. This turns out to be because there
was an accidental divergence in one of the SQLite queries in which it
was omitting a filter that was present in the Postgres version.

`JobCountByQueueAndStateResult` is used only in the queue pane of River
UI and filters to `available` and `running`. Although omitting this
filter may use the same index, there tends to be a massive quantity of
`completed` rows that have to be counted unless that state is filtered
out, which is why SQLite specifically was slow.


bgentry (approved):

brandur:
thx!

If it's alright, I'm also going to cut a pair of new releases. I want to get this fix into River UI so I can double check that the SQLite version is just as performant (or similarly) as the Postgres version.

Issue #1358, opened 2026-08-22
Remote JobCancel() can be silently lost while the notifier is reconnecting (no durable-poll fallback)


chrisgeo:
## Summary

`Client.JobCancel()` delivers its cancellation signal to a running job's executor exclusively via Postgres `LISTEN`/`NOTIFY`. If the notifier happens to be mid-reconnect (its exponential backoff loop, `internal/notifier/notifier.go`) at the moment the `NOTIFY` is sent, that specific signal is lost — not delayed — because Postgres `NOTIFY` is fire-and-forget and only reaches sessions that are actively `LISTEN`ing at commit time. The job keeps running unaware it was cancelled, and the only recovery is `JobRescuer`'s stuck-job sweep, which defaults `RescueStuckJobsAfter` to one hour when unconfigured.

This is a real, reproducible failure mode, not a theoretical one — details and evidence below.

## Mechanism

1. `notifier.go`'s run loop treats any listener error (a transient connection reset, a pooler cycling the backend, a brief network blip, etc.) as cause to disconnect and retry with exponential backoff: `0.5s, 1s, 2s, 4s, 8s, 16s, 32s, 64s` (`rivershared/util/serviceutil.ExponentialBackoff`, `MaxAttemptsBeforeResetDefault = 7`; the run loop's own `attempt` counter starts at 0, one step earlier than `ExponentialBackoff`'s doc comment assumes, so the very first retry lands at 0.5s before the schedule settles into the documented 1s→64s cycle on later retries). Each connect/listen attempt also has its own 10s timeout (`listenerTimeout`).
2. While the notifier is in that backoff window, it is not `LISTEN`ing on `river_control` (or any topic) at all.
3. If `JobCancel()` is called during that window, its `NOTIFY` commits successfully but has no active listener to reach — it is gone. `JobCancel()` itself has no visibility into notifier connectivity and returns success regardless (correctly — the durable state write, `cancel_attempted_at` in the job's metadata, did succeed).
4. The running job's executor never observes `ctx.Done()`. It keeps running until it completes normally, or until `JobRescuer` eventually rescues it — up to an hour later by default, since `RescueStuckJobsAfter` isn't tied to `cancel_attempted_at` specifically; it's the same general stuck-job window.

Confirmed by reading the source directly (not inferring from behavior) against v0.44.0 — this is not a stale/fixed-since concern; I checked master too (currently a few commits past v0.44.1, none touching notifier/producer/rescuer).

## Evidence

[...]

PR #1361, opened 2026-08-23
Reconcile active jobs against cancel_attempted_at after notifier reconnects

chrisgeo:
Fixes the mechanism described in #1358: `Client.JobCancel()` delivers its cancellation signal via Postgres `LISTEN`/`NOTIFY`, and `NOTIFY` is fire-and-forget — if the notifier is disconnected/reconnecting (its exponential backoff loop) at the moment the `NOTIFY` commits, the signal is lost for good. The running job keeps executing unaware until `JobRescuer`'s stuck-job sweep eventually catches it (defaults to a 1h window).

As discussed in #1358 (comment from @brandur): each client now polls its own currently-running jobs by primary key for `cancel_attempted_at` every time its notifier (re)establishes healthy listening — both the very first connect and every subsequent reconnect. This is cheap (a plain `id = any(...)` lookup via the existing `JobGetByIDMany`, no new query/index), rare (only fires on reconnect), and closes exactly the lost window: a signal lost during a reconnect is caught the moment the reconnect completes.

**Changes (first commit):**
- `internal/notifier`: new `Notifier.RegisterListenerReadyFunc` — lets a caller register a callback invoked every time the notifier establishes healthy listening on all its subscribed topics (including the initial connect; callers that only care about genuine reconnects can no-op on an empty precondition, which is what the producer does below).
- `producer.go`: registers a `ListenerReadyFunc` (only when a notifier is configured) that signals a new `reconnectCh`, consumed on the producer's single owning goroutine. The handler is a no-op when there are no active jobs (true at startup, so the initial-connect firing is free); otherwise it looks up its currently-active job IDs via `JobGetByIDMany` and cancels any with `cancel_attempted_at` set. A query failure retries with the existing exponential backoff utility rather than silently dropping the reconciliation; a malformed metadata row is logged and skipped without aborting the rest of the batch.

**A second gap from the same root observation (second commit, "Deliver cancellation locally when the client has no notifier"):** a client configured without a notifier at all — including an explicit `Config.PollOnly` on an otherwise listener-capable driver (pgx, SQLite) — had no delivery path whatsoever for cancelling a job running in its own process; the local same-process dispatch that already exists for this case only ever activated for drivers that can't support `LISTEN`/`NOTIFY` at all, missing the more common explicit-`PollOnly` case entirely. This commit adds a narrowly-scoped fix via its own dispatch helper rather than changing the existing shared one (which is also used by queue pause/resume/update and has different, already-correct semantics for those). Happy to split this into its own PR if preferred.

**Testing:** Added `TestProducer_JobCancelSurvivesNotifierReconnect`, a deterministic repro against a real Postgres notifier: forces one simulated connection loss after a job starts, holds the notifier at the start of its reconnect attempt, cancels the job while disconnected (so the `NOTIFY` is lost), then releases the reconnect and asserts the job observes the cancellation. Fails on `master` (job runs to completion unaware); passes with this change. Also added direct coverage for `RegisterListenerReadyFunc` in `internal/notifier`, and for the second commit: a same-process explicit-`PollOnly` cancellation test and a regression test guarding against blocking when cancelling a burst of non-running jobs.

No steady-state query cost is added to either fix — the reconciliation only runs on reconnect and is a no-op when nothing is running; the local dispatch only fires for jobs actually running in-process.


brandur:
@chrisgeo Can you add yourself to the CLA? https://github.com/riverqueue/rivercla

Also, why don't you go ahead and rebase. We merged another fix quite recently that affected the changelog.

chrisgeo:
@brandur Rebased onto current master — the changelog conflict from #1359 is resolved. CLA signed (riverqueue/rivercla#33).

client.go
```diff
@@ -2876,6 +2887,58 @@ func (c *Client[TTx]) notifyProducerWithoutListenerQueueControlEvent(queue strin
...
+// has for genuinely listener-incapable drivers. The caller (JobCancel) is
+// responsible for only invoking this for jobs it knows are actually running,
+// both because there's nothing to dispatch for a job with no in-process
+// executor and to avoid a burst of cancellations against non-running jobs
+// (e.g. many freshly-inserted jobs cancelled before the client is started)
+// filling that channel and blocking indefinitely.
+//
+// Should only ever be invoked *outside* a transaction. If invoked within a
+// transaction, the producer wouldn't yet be able to access the state that
+// triggered the notification because it's not committed yet.
```
brandur, line 2924: In the old days when everything was human written, I'd be okay with this sort of detailed comment. However, with the advent of LLMs, it's become far too easy to filibuster with super long comment blocks everywhere because the cost to generate them has gone to zero. When the whole codebase is full of them, no human will bother reading them anymore.

Do you want to see if you can your LLM to compact this down a little more? A lot of the comments elsewhere in the PR could be similarly boiled down more.

producer_test.go
```diff
@@ -53,6 +55,182 @@ func (p *beforeJobGetAvailablePilot) JobGetAvailable(
...
+	require.NoError(t, exec.NotifyMany(ctx, &riverdriver.NotifyManyParams{
+		Topic:   string(notifier.NotificationTopicInsert),
+		Payload: []string{`{"queue":"` + queueName + `-unrelated-wakeup"}`},
+		Schema:  schema,
+	}))
+
+	// Wait until the notifier has actually observed the injected error and is
+	// blocked trying to reconnect.
+	require.Eventually(t, reconnecting.Load, 5*time.Second, 5*time.Millisecond,
+		"notifier never reached its reconnect attempt after the injected connection loss")
```
brandur, line 214: We have a concept of "test signals" used throughout here that this kind of thing would be a good fit for so you don't have to introducing sleeping into tests.

OOC, what LLM are you using? Normally using Claude and Codex, I find both have been smart enough to find and use the test signal convention automatically.

brandur (commented):
Thanks @chrisgeo — sorry about the time it took me to get back to this one.

I ran my LLM against the PR and it flagged a few issues we might want to fix before bringing it in. Do you mind taking a look at these?

>   1. [P1] Reconciliation can block cancellation delivery. producer.go:857
>      JobGetByIDMany runs synchronously on the producer’s main goroutine without a timeout. While it waits, that goroutine cannot
>      process cancellations or job completions. If workers hold the pool’s connections while waiting for cancellation, this
>      creates a deadlock. Run the query asynchronously with a bounded context, then apply its results on the producer goroutine.
> 
>   2. [P1] Local cancellation can hang indefinitely on an unstarted client. client.go:2937
>      A job’s running state doesn’t establish that this client owns it. Producers exist before Start(), so an unstarted PollOnly
>      client canceling jobs running elsewhere fills its undrained control channel. I reproduced the 101st cancellation blocking
>      past its context deadline. Local delivery needs to account for producer lifecycle and avoid an unconditional blocking send.
> 
>   3. [P2] Reconnection cancels jobs that were deliberately retried. producer.go:876
>      JobRetry preserves cancel_attempted_at. After canceling a job, retrying it, and starting its new attempt, reconciliation
>      interprets the old marker as a fresh cancellation and interrupts the worker. Clear the marker when retrying or otherwise
>      scope cancellation to the current attempt.

PR #1365, opened 2026-08-27
Support mixed drivers in transactional work

### LINKED ISSUE #1364: Public way to supply the client to `JobCompleteTx` when completing in a different client’s transaction type
We run two clients against one database, and `JobCompleteTx` currently can’t be used cleanly in that topology without importing rivertest into production code.

**Setup.** Our application transactions come from a database/sql-based ORM pool, so transactional job inserts and `JobCompleteTx` must run on a `Client[*sql.Tx]` (riverprodatabasesql — we’re a Pro customer). 

But a database/sql driver can’t LISTEN, which means remote cancellation of running jobs never reaches workers. The fix was to split: keep the *sql.Tx client for inserts/completion, and run workers on a riverpropgxv5 client so producers get a notifier. This seems like it will work well.  The cancel’s pg_notify is emitted by the SQL itself, so any client can send and the pgx workers receive.

**Problem.** `JobCompleteTx[TDriver]` resolves its client with `ClientFromContext[TTx](ctx)`. Inside a worker running under the pgx client, the context carries a `*Client[pgx.Tx]`, so the `*sql.Tx`-typed lookup fails, even though we have a perfectly good *sql.Tx client to complete through. 

The only public way to fix the context is **`rivertest.WorkContext`**.

```
// production code today:
ctx = rivertest.WorkContext(ctx, sqlTxClient)
_, err := river.JobCompleteTx[*riverprodatabasesql.Driver](ctx, tx, job)
```

### Ask

Either of these would remove the rivertest import from production code:
 1. Export the context setter (e.g. `river.WithClient[TTx](ctx, client)`), or
 2. A JobCompleteTx variant that accepts the client explicitly, e.g. `client.JobCompleteTx(ctx, tx, job)` or `river.JobCompleteTxClient(ctx, client, tx, job)`.

But very flexible here if you advise something else.

Observed on v0.39.0; the relevant code is unchanged on master.

bgentry:
Worker contexts currently bind transactional worker helpers to the worker
client's transaction type. Applications that execute jobs with one River
driver but use another driver for application transactions cannot complete a
job or persist a resumable step atomically through those transactions.

Use the explicit `TDriver` parameter to adapt the supplied transaction while
continuing to take the pilot, schema, and clock from the worker client. This
keeps completion behavior associated with the worker that owns the job without
requiring the transaction and worker clients to expose the same Go transaction
type.

Apply the same separation to `JobCompleteTx`, `ResumableSetStepTx`, and
`ResumableSetStepCursorTx`.

River Pro compatibility is handled in
[riverqueue/riverpro#360](https://github.com/riverqueue/riverpro/pull/360).

Fixes #1364.

### DIFF
--- a/client_context.go
+++ b/client_context.go
@@ -3,11 +3,49 @@ package river
 import (
 	"context"
 	"errors"
+	"time"
 
 	"github.com/riverqueue/river/internal/rivercommon"
+	"github.com/riverqueue/river/rivershared/riverpilot"
 )
 
 var errClientNotInContext = errors.New("river: client not found in context, can only be used in a Worker")
+
+type clientContextData struct {
+	Pilot  riverpilot.Pilot
+	Schema string
+	Time   time.Time
+}
+
+type clientContextProvider interface {
+	clientContextData() *clientContextData
+}
+
+func clientContextDataFromContext(ctx context.Context) *clientContextData {
+	client, exists := ctx.Value(rivercommon.ContextKeyClient{}).(clientContextProvider)
+	if !exists || client == nil {
+		panic(errClientNotInContext)
+	}
+
+	data := client.clientContextData()
+	if data == nil {
+		panic(errClientNotInContext)
+	}
+
+	return data
+}
+
+func (c *Client[TTx]) clientContextData() *clientContextData {
+	if c == nil {
+		return nil
+	}
+
+	return &clientContextData{
+		Pilot:  c.Pilot(),
+		Schema: c.Schema(),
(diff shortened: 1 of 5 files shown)


bgentry:
Closing this draft after further design review. Apologies, Codex got overeager and turned an incidental pre-Pilot behavior into proposed support for mixed transactional drivers. That doesn't fit River's client/driver/Pilot ownership model, and the pgx-backed listener approach in #1366 is the better solution to #1364.

PR #1366, opened 2026-08-27
Add pgx listener support to database/sql

bgentry:
Adds `riverdatabasesql.NewWithPgxListener(dbPool, listenerPool)`. The `*sql.DB` remains the exclusive executor for River queries and transactions, while a private `riverpgxv5` driver uses the separately supplied Pgx pool only to create listeners.

This lets applications built around `database/sql`, Bun, GORM, or another ORM run one coherently typed River client with Postgres `LISTEN`/`NOTIFY` support. Cross-process actions such as cancelling a running job no longer have to wait for the poll interval, while `JobCompleteTx` continues to operate on an ordinary `*sql.Tx`. The existing `New` constructor remains explicitly poll-only, and the new constructor rejects a nil listener pool.

The caller owns both pools and must configure them for the same database and schema. A listener pool dedicated to one River client can generally use `MinConns: 0` and `MaxConns: 1`. Listener connections are hijacked and therefore no longer count against the pool maximum or close with the pool; stopping the River client closes them. PgBouncer must use session pooling or be bypassed for these connections.

This provides a design-aligned solution for #1364 without adding mixed transactional-driver support.

River Pro companion: [riverqueue/riverpro#361](https://github.com/riverqueue/riverpro/pull/361).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added `riverdatabasesql.NewWithPgxListener` for using a dedicated Pgx pool to receive Postgres notifications while continuing to execute jobs and transactions through `database/sql`. [PR #1366](https://github.com/riverqueue/river/pull/1366).
+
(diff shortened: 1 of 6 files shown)


brandur (approved):
Cool! WFM.

PR #1367, opened 2026-08-29
Prepare release v0.46.0

brandur:
[skip ci]


bgentry (approved):
Nice, this is going to warrant some doc fixes in places where we say this driver doesn’t support listen

brandur:
> Nice, this is going to warrant some doc fixes in places where we say this driver doesn’t support listen

Sweet. Yep, should be easy for Codex.

PR #1368, opened 2026-08-29
Add back more non-listener coverage for `riverdatabasesql`

brandur:
Small follow up to #1366 in which we add back non-listener test coverage
for `riverdatabasesql`. In #1366 the driver tests replaced the
non-listener driver initialization with the listener version which is
probably largely okay since most code is going to get shared, but it's
not a bad idea to also make sure the driver is fully operable without a
listener.


bgentry (approved):

brandur:
thx!

